PR #7, opened 2017-08-28
Fix Segmentation fault on Linux amd64

sepldj:
The version defines which were passed to version.c were incorrectly
constructed in the CMakeLists.txt file causing a seg fault when
invoking edge -h


lucaderi:
Thank you

PR #9, opened 2017-12-03
fix mk_tar.sh manifest list

ZoltanLajosKis:
The package building scripts are broken as mk_tar.sh expects a README file, but there is a README.md instead.


lucaderi:
Thank you

PR #23, opened 2018-07-03
Update edge.service.in

moshengrenzenme:
Sorry,I'm a Chinese， English is not good. This is Machine Translation. Please forgive me for the mistakes in the translation. If -f is ignored, it will not be able to start automatically, prompting the parent process to exit

### DIFF
--- a/packages/etc/systemd/system/edge.service.in
+++ b/packages/etc/systemd/system/edge.service.in
@@ -6,7 +6,7 @@ Wants=
 [Service]
 Type=simple
 ExecStartPre=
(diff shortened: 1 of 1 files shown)


lucaderi:
Hi, I am not sure I understand. I you add -f you will *not* fork
-f                       | Do not fork and run as a daemon; rather run in foreground.
Is this what you want contrary to today where you fork instead?

moshengrenzenme:
I use Ubuntu16.04, if I do not add-f, the boot will not run automatically, systemd prompts dead

lucaderi:
Thanks

PR #53, opened 2018-10-31
Fix packets from clients being routed even when routing is disabled

emanuele-f:
The routing check is only performed on packets received from the TAP. However, all the packets received from the remote clients are forwarded even when routing is disabled. This patch fixes this.


lucaderi:
Thank you

Issue #28, opened 2018-07-16
n2n edge connection times out


emanuele-f:
I have noticed that n2n sometimes stalls when there are a lot of packet drops due to unreliable communication channel. Restarting n2n on both the edge nodes solves the issue, so I suspect some metadata n2n packets (e.g. registration) are dropped and not retransmitted. I will provide more information when I reproduce the problem again.

dinopio:
When an edge is started and no traffic is sent (idle) it might lose the connection.
A way to keep the edge nodes up is to keep pinging the edge ip from a node on the network.
This is currently the only way I can keep the network up.

ghostry:
[a temporary solution](https://blog.ghostry.cn/doc/docs/n2n/n2n-1b1am7b0njmbf)

PR #60, opened 2019-01-28
Add AES CBC psk encryption

emanuele-f:
Adds AES encryption support for pre shared key mode. Previously AES was only enabled when a keyfile was passed with the `-K` (uppercase) option. With this pull request it's possible to use AES with `-k` (lowercase) too, by adding the `-A` option. Encription in use can be verified with using the `-v` option.

Without `-A` (default):
`28/Jan/2019 01:08:42 [transform_tf.c:132] encode_twofish 98 with SA 0.`

With `-A`:
`28/Jan/2019 01:13:09 [transform_aes.c:162] encode_aes 98 with SA 0.`

AES should become the preferred encryption method in the future as, contrary to twofish, it uses a standard system library. Here are some benchmarks on my system (`./benchmark`):

```
[transop_null] 10000 times: (6356 -> 635 nsec each) 1548636359.095106 -> 1548636359.101462.
[transop_twofish] 10000 times: (494206 -> 49420 nsec each) 1548636359.101502 -> 1548636359.595708.
[transop_aes] 10000 times: (63927 -> 6392 nsec each) 1548636359.595725 -> 1548636359.659652.
````

Note: system with AES hardware accelleration

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -76,7 +76,7 @@ supernode: sn.c $(N2N_LIB) n2n.h Makefile
 	$(CC) $(CFLAGS) sn.c $(N2N_LIB) $(LIBS_SN) -o supernode
 
 benchmark: benchmark.c $(N2N_LIB) n2n_wire.h n2n.h Makefile
-	$(CC) $(CFLAGS) benchmark.c $(N2N_LIB) $(LIBS_SN) -o benchmark
+	$(CC) $(CFLAGS) benchmark.c $(N2N_LIB) $(LIBS_EDGE) -o benchmark
 
 example_edge_embed: example_edge_embed.c $(N2N_LIB) n2n.h
 	$(CC) $(CFLAGS) example_edge_embed.c $(N2N_LIB) $(LIBS_EDGE) -o example_edge_embed
--- a/benchmark.c
+++ b/benchmark.c
@@ -70,14 +70,38 @@ uint8_t PKT_CONTENT[]={
 
 /* Prototypes */
 static ssize_t do_encode_packet( uint8_t * pktbuf, size_t bufsize, const n2n_community_t c );
+static void run_transop_benchmark(const char *op_name, n2n_trans_op_t *op_fn, uint8_t *pktbuf, n2n_community_t c);
 
 int main(int argc, char * argv[]) {
   uint8_t pktbuf[N2N_PKT_BUF_SIZE];
-  n2n_trans_op_t transop_null;
+  n2n_community_t c;
+  n2n_trans_op_t transop_null, transop_twofish, transop_aes_cbc;
+  u_char encrypt_pwd[] = "SoMEVer!S$cUREPassWORD";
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by lucaderi** [comment]
Great work

**Comment by lucaderi** [comment]
The main issue I see for making it mainstream is to enable interoperation with edge instances that have not been started with -A

**Comment by Logan007** [comment]
Wow, this is awesome!  Thank you!!

I see no problem for the mainstream, as it probably only is useful in conjunction with -k; so it could be considered as some kind of key related parameter which needs to be given manually anyway when invoking edge using -k.  It just needs to be communicated together with the psk which needs to be shared somehow anyway on some occasion.

Does it also work using N2N_KEY environment?  I need to check it out immediately. [EDIT: As of now, not yet... seems that -A is coupled to -k]

**Comment by emanuele-f** [comment]
Hi, it works for me with the N2N_KEY env. N2N_KEY is equivalent to `-k`

**Comment by Logan007** [comment]
Oh, my bad!  You are so right, of course.  I should have watched better the path to the executable file...   Maybe I am getting too old for this.  I am sorry and embarrassed.

Anyway, "-A" is something I dreamt of!  Wonderful work and thank you again!

A real quick benchmark here (some older i7-2860QM including cpu hw aes flag) delivers the following output:
```
[transop_null] 10000 times: (2822 -> 282 nsec each) 1548964976.065853 -> 1548964976.068675.
[transop_twofish] 10000 times: (376056 -> 37605 nsec each) 1548964976.068716 -> 1548964976.444772.
[transop_aes] 10000 times: (52552 -> 5255 nsec each) 1548964976.444788 -> 1548964976.497340.
```

I guess, real life throughput is much more limited by bandwitdth.

emanuele-f:
7x speed increase is not bad :) Next step to improve n2n is to fix the connection issues like #28 . Regarding encryption, we also have to evaluate if adding some salt or hash on the raw passphrase before feeding it into aes/twofish is better from a security point.

Logan007:
A hash on the key before feeding it into the AES or TF would be a good thing to do, as hashed values usually spread the bits broader and look more pseudo-random as an ASCII coded password ever could. It makes more unpredictable use of the available key space.

A salt would provide additional security in situations where the password hash (or, in the case mentioned above: a hash of the hash) should be stored (together with the salt) for verification purposes, i.e. authentication. So you probably do not need salt for now.

SHA256 or similar should do for AES with 256 bit key size.

By the way, wouldn't a 512 bit key (and thus SHA512) be an even more secure choice with a view to a block size of 256 bit? There are 256! permutations for blocks of 256 bits, by far not each of them could be named by a 256 bit key (2 ^ 256 << 256!) ; 2 ^ 512 gets at least a bit closer...

emanuele-f:
@Logan008 if possible please open a new feature request for the use of an hash function and explain with some examples why using SHA512 would be a better choice than SHA256, thank you

PR #66, opened 2019-02-11
Submit minor changes to enable Windows CMake builds w/o requiring additional local mods.

csheely:
For your consideration: I've made a few minor changes so that Windows builds complete without errors.

* Added "build" directory to .gitignore, to ignore out-of-source CMake build artifacts
* Updated CMakeLists.txt to tag builds with version 2.5.0, like `configure` does for other platforms
* Added OpenSSL include/library references for CMake builds
* Added conditional compilation for setsockopt() calls referencing SO_REUSEPORT (not included in Windows)
* Added winconfig.h to take the place of config.h (generated by `configure` on other platforms).
* Replaced `signal(SIGHUP, ...)` with `signal(SIGINT, ...)` on Windows platforms (SIGHUP doesn't exist).

### DIFF
--- a/sn.c
+++ b/sn.c
@@ -20,6 +20,10 @@
 
 #include "n2n.h"
 
+#ifdef WIN32
+#include <signal.h>
+#endif
+
 #define N2N_SN_LPORT_DEFAULT 7654
 #define N2N_SN_PKTBUF_SIZE   2048
 
@@ -919,7 +923,11 @@ int main(int argc, char * const argv[]) {
 
   traceEvent(TRACE_NORMAL, "supernode started");
 
+#ifndef WIN32
   signal(SIGHUP, dump_registrations);
+#else
+  signal(SIGINT, dump_registrations);
+#endif
   
   return run_loop(&sss_node);
 }
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ benchmark
 edge
 example_edge_embed
 supernode
+build
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@ project(n2n)
 cmake_minimum_required(VERSION 2.6)
 
 # N2n information
-set(N2N_VERSION 2.3.0)
+set(N2N_VERSION 2.5.0)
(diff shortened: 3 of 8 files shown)


ghostry:
sn.c  409  u_int32_t num_communities = 0; 
mingw32-make prompt uint32
After the compilation is complete, supernote can be used
Edge needs to copy libcrypto.dll to the directory

sn.c
```diff
@@ -919,7 +923,11 @@ int main(int argc, char * const argv[]) {
 
   traceEvent(TRACE_NORMAL, "supernode started");
 
+#ifndef WIN32
   signal(SIGHUP, dump_registrations);
+#else
+  signal(SIGINT, dump_registrations);
```
emanuele-f, line 929: Please avoid handling the SIGINT signal, it should be used to terminate the program as normal

emanuele-f:
I can compile successfully with this fix, however I have a problem reaching the windows edge from an external edge node (the other way works fine). Moreover, after pinging the windows edge from outside, I cannot terminate the edge program (even closing the terminal window does not work). Anyway, I think these issues can be discussed in another thread. Also, adding a link to http://build.openvpn.net/downloads/releases would be useful in order to install the tap drivers. @csheely please check out my comment on the sigint handling

emanuele-f (commented):
I can compile successfully with this fix, however I have a problem reaching the windows edge from an external edge node (the other way works fine). Moreover, after pinging the windows edge from outside, I cannot terminate the edge program (even closing the terminal window does not work). Anyway, I think these issues can be discussed in another thread. Also, adding a link to http://build.openvpn.net/downloads/releases would be useful in order to install the tap drivers. @csheely please check out my comment on the sigint handling

emanuele-f:
I'll remove the SIGINT signal handler above in a separate commit

csheely:
Thanks, and apologies for not getting back to this in time.  Completely understand about removing SIGINT reference.  Since SIGHUP wasn't defined on windows I was just looking for a replacement.

emanuele-f:
Thank you for the fixes, I've fixed that and added instructions in commit 5742ef978f264cb9692c0f8d830515c3200b20be

PR #67, opened 2019-02-20
Fixes packet drops while communicating with multiple nodes in AES PSK

emanuele-f:
Per-node AES structures must be kept as CBC cannot work with a single structure across multiple nodes.
The peer mac address is used as a key to determine the structure to use.

### DIFF
--- a/benchmark.c
+++ b/benchmark.c
@@ -102,6 +102,7 @@ int main(int argc, char * argv[]) {
 static void run_transop_benchmark(const char *op_name, n2n_trans_op_t *op_fn, uint8_t *pktbuf, n2n_community_t c) {
   n2n_common_t cmn;
   n2n_PACKET_t pkt;
+  n2n_mac_t mac_buf;
 
   struct timeval t1;
   struct timeval t2;
@@ -115,6 +116,7 @@ static void run_transop_benchmark(const char *op_name, n2n_trans_op_t *op_fn, ui
 
   n=10000;
 
+  memset(mac_buf, 0, sizeof(mac_buf));
   gettimeofday( &t1, NULL );
   for(i=0; i<n; ++i)
     {
@@ -122,7 +124,7 @@ static void run_transop_benchmark(const char *op_name, n2n_trans_op_t *op_fn, ui
 
       nw += op_fn->fwd( op_fn,
 			      pktbuf+nw, N2N_PKT_BUF_SIZE-nw,
-			      PKT_CONTENT, sizeof(PKT_CONTENT) );
+			      PKT_CONTENT, sizeof(PKT_CONTENT), mac_buf);
 
       idx=0;
       rem=nw;
--- a/edge_utils.c
+++ b/edge_utils.c
@@ -791,7 +791,7 @@ static int handle_PACKET(n2n_edge_t * eee,
 	eh = (ether_hdr_t*)eth_payload;
 	eth_size = eee->transop[rx_transop_idx].rev(&(eee->transop[rx_transop_idx]),
 						    eth_payload, N2N_PKT_BUF_SIZE,
-						    payload, psize);
+						    payload, psize, pkt->srcMac);
 	++(eee->transop[rx_transop_idx].rx_cnt); /* stats */
 
 	if(!(eee->allow_routing)) {
@@ -1192,7 +1192,7 @@ void send_packet2net(n2n_edge_t * eee,
 
   idx += eee->transop[tx_transop_idx].fwd(&(eee->transop[tx_transop_idx]),
 					  pktbuf+idx, N2N_PKT_BUF_SIZE-idx,
-					  tap_pkt, len);
+					  tap_pkt, len, pkt.dstMac);
   ++(eee->transop[tx_transop_idx].tx_cnt); /* stats */
 
   send_packet(eee, destMac, pktbuf, idx); /* to peer or supernode */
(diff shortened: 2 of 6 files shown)


Logan007:
I am so glad you found that... I was already wondering why packet loss visibly went up with the latest git. I will test it soon and give feedback.

Logan007:
Okay, it took me a while to figure out how to download the code including the fix - my coding times were way before git :)  I compiled, installed and used the n2n-3723...cc11 (zip download from the tree/sha-code web site) and still found significantly more packet losses on a flood ping (sudo ping -f) while using AES (via -A option) as opposed to  Twofish (same command line just without -A option).

Maybe I screwed up the download?

emanuele-f:
I didn't know the -f flag, cool. How are you performing the tests exactly? For testing, you can use the command "git clone https://github.com/ntop/n2n", then download "https://pastebin.com/raw/LkGnXnfV" and name it n2n.patch, then inside the n2n directory `git apply n2n.patch`. To doublecheck, you should have a function named aes_psk_get_peer_sa (added with this patch)

dinopio:
fix works fine!

Logan007:
Thank you for the patch file, I applied it to completly fresh and full git clones on all the computers I use.  After building, I installed :). Unfortunately, with the same result. While null transform and TF do not show significant packet loss, AES does (1 to 3 percent).

I have two local edges, a public supernode and another far, far away double NATed edge. It's all on Linux. Packet loss with AES even occurs locally. For testing, I watch «sudo ping -f»'s output for a while.

I am sorry that I do not have more specific hints for now.

Does AES work for you without any packet losses that increase over time? Maybe it's just my setup.

emanuele-f:
With ping -f you are basically performing a 1 to 1 communication between two edges, so this patch should not be relevant in your case as the problem it solves is multiple nodes connecting each other. Please correct me if I'm missing something. Here are my statistics for a ping flood of a remote host on a 7 minutes interval:

```
PING 192.168.4.1 (192.168.4.1) 56(84) bytes of data.
......
--- 192.168.4.1 ping statistics ---
20102 packets transmitted, 20096 received, 0.0298478% packet loss, time 1186ms
rtt min/avg/max/mdev = 35.997/37.895/116.505/2.200 ms, pipe 8, ipg/ewma 17.652/37.538 ms
```

Logan007:
I agree that ping is 1 to 1, still all the nodes (three) communicate in the background with each other, there always some is traffic flowing: broadcasts, keep alives and automated data transfer stuff. Does that match the multiple node commuication scenario the patch addresses?

In my setup described above, a floodping on TF or NULLTRANSFORM shows some initial packet loss, but then does not increase significantly. On AES, there also is this initial packet loss and then a slowly constantly further increasing packet loss.

Again, maybe it's just my setup... TCP can cope the packet loss.

emanuele-f:
Please perform some tests 1 to 1 only before digging into the 1 to n situation. Try both twofish and aes in a community with only 2 hosts with a ping -f for 5 minutes in both cases and post here the results like my ping output above.

Logan007:
Thanks for getting back to this. I did it for 5 minutes, 1 to 1 disabling the additional edge.

TF shows some initial packet loss again (the points you can see) right in the beginning, remaining stable over the rest of the five minutes:
```
sudo timeout -s SIGINT 300 ping -f 10.19.49.14
PING 10.19.49.14 (10.19.49.14) 56(84) bytes of data.
...........                        
--- 10.19.49.14 ping statistics ---
18430 packets transmitted, 18419 received, 0.0596853% packet loss, 
time 1286ms rtt min/avg/max/mdev = 172.729/179.523/557.474/23.809 ms, 
pipe 35, ipg/ewma 16.277/174.319 ms
```

On AES (same command line but with -A), it also showed roughly the same initial packet loss but then increased constantly:
```
sudo timeout -s SIGINT 300 ping -f 10.19.49.14
PING 10.19.49.14 (10.19.49.14) 56(84) bytes of data.
..............................................................................................................................................................................................................................................................................................................................................     
--- 10.19.49.14 ping statistics ---
18390 packets transmitted, 18086 received, 1.65307% packet loss, 
time 1284ms rtt min/avg/max/mdev = 169.149/174.149/498.313/19.052 ms, 
pipe 30, ipg/ewma 16.313/170.976 ms
```
As you might see from the ping times, it was a connection half around the world using a supernode to bridge double-NAT in between.

But even if I chose two local nodes, I get relatively (!) similar results and behaviour (first loss only **vs.** first loss plus additional constantly increasing loss), not as many absolute losses, though. On TF:
```
sudo timeout -s SIGINT 300 ping -f 10.19.49.14
PING 10.19.49.14 (10.19.49.14) 56(84) bytes of data.
.              
--- 10.19.49.14 ping statistics ---
46688 packets transmitted, 46687 received, 0.00214188% packet loss, 
time 1291ms rtt min/avg/max/mdev = 2.975/6.776/251.459/7.324 ms, 
pipe 15, ipg/ewma 6.425/6.498 ms
```
On AES:
```
sudo timeout -s SIGINT 300 ping -f 10.19.49.14
PING 10.19.49.14 (10.19.49.14) 56(84) bytes of data.
.....................................................................              
--- 10.19.49.14 ping statistics ---
46119 packets transmitted, 46050 received, 0.149613% packet loss, 
time 1295ms rtt min/avg/max/mdev = 3.001/6.920/266.580/8.540 ms, 
pipe 16, ipg/ewma 6.504/6.393 ms
```

emanuele-f:
Since the issue is not directly linked with this PR I've merged the changes. Please open a separate issue with the information above. Please also check out the CPU usage while performing these tests

Issue #73, opened 2019-03-01
Add protection against replay attacks


emanuele-f:
As explained in #70, the current implementation does not provide a protection against replay attacks.

Logan007:
I would add this feature by coding something special in the initialization vector (IV) for cipher block chaining (CBC). That way, there is no need for an extra data field. Basically, there are three options:

1) A **counter**. Just counting it up starting at 0 or some other value maybe derived from the key. The receiver must be tolerant to packet loss or even a longer supernode turndown. Would that require a regular re-synchronisation, maybe with every supernode register? The current counter value could be internally stored in the sa structure.

2) A more sophisticated way of counting using a predictable **pseudo random number generator** (PRNG), initialized by some common seed, probably derived from the key. Also, the PRNG's paramters could depend on the key. This also would require the receiver to be somewhat tolerant, the checking needs some window in which some PRNGed numbers are precalculated in advance. The sa structure could hold the current status of the PRNG.

3) A **time stamp**, e.g. the 32 bit unix time stamp with precision of seconds - windows should have something comparable. Given, that most systems worldwide that are connected to a network usually access a ntp server - just think of the huge ntp.org time server pools, clocks should be somewhat in sync. The checking just needs to add a +/- time frame to compensate for clocks keeping bad time and some transmission delay, something between 4 to 8 seconds probably. If a time stamp of higher precision is available, it could help to identify the correct order of the packets to prevent fast replay attacks - assuming that packets do not overtake each other on the line.

Alternatives 1. and 2. are sequential, meaning that they repeat in the same order starting from some point, i.e. the first packet after (re-)initialization gets the same counter or PRNGed number.

Alternatives 1. and 3. clearly run counter to the randomness of the IV - it shall be as unpredictable as possible, especially in our case that has a highly predictable first cipher block. This could be solved by  mixing the bits of the counter or the time stamp somewhat randomly but reversibly in the IV. How could that be achieved any better than using the encryption algorithm already at hand? Thus, a possible scheme would be:

```
Generate Random IV of 128 bit (minus time stamp size) block size
Fill it up with anti-replay-part, maybe the time stamp, at the end to fill up full 128 bit block size
Encrypt this block using K1
Use the result as IV for cbc mode encryption of the data
Encrypt this block using K2 (as already suggested in IV implementation of #76)
Store this doubly encrypted block in front of the encrypted package (as already suggested in IV implementation of #76)
```

[...]

Issue #68, opened 2019-02-20
Hash the key before use & Max out the key size


Logan007:
A hash on the key before feeding it into the AES or TF would be a good thing to do, as hashed values usually spread the bits broader and look more pseudo-random than an ASCII coded password ever could. This is also known as "key stretching". It makes more unpredictable use of the available key space.

As maximum AES and TF key size is 256 bit, _the first half of SHA512 would deliver the key. The second half of the hash value could be used for HMAC or similar (seperate issue [to be created later](https://github.com/ntop/n2n/issues/70))_.

And then, the maximum key length of 256 bit should always be used. First of all, because it is easier to _always_ use the _complete first half of the_ hash value as key.

But also, as the bigger key size makes better use of the available key space as follows: There are (2 ^ 128)! different possible swirls for plain text blocks being transformed into cipher, with 128 being the block size and ! the factorial function. Each key (2 ^ 256) addresses one of these permutations, but not all, as (2 ^ 256) is much smaller than (2 ^ 128)! . So, if we already are not able to use the complete key space, why should the software limit itself even further by using even smaller keys? Though, performance might be an excuse. Not so sure about the difference it might make.

As for AES, a suitable place to apply the hash function might be the "add_aes_key"-function.

No need for a salt as long as no password hashes (or hashes of password hashes) are stored for verification purposes, i.e. authentication _in terms of "approval"_.

[_EDITed_]

Logan007:
Please find [attached](https://github.com/ntop/n2n/files/2902803/sha512edKey.txt) a real quick and dirty patch as _proof of concept_.

emanuele-f:
Thank you for the poc. If you want to contribute please read https://gist.github.com/MarcDiethelm/7303312 . When performing changes, please try to only modify the minimal code to accomplish the result.

Logan007:
Still learning git I probably only can contribute patches for now.

Sorry for being too lengthy in the proof of concept. Please find attached a shorter [patch](https://github.com/ntop/n2n/files/2922316/sha512edKey.txt) including way less changes. Still somewhat dirty though...

emanuele-f:
Implemented in bb07f0426e6b83e8851307590620245a6950f80b

PR #76, opened 2019-03-03
Revamp AES Part 1; including random & encrypted IV, hashed key and max keysize

Logan007:
This pull request addresses some security issues in the AES module _transform_aes_ as follows:

**Key Stretching**
- the key provided by the user gets hashed
- this hash is used to derive the actual encryption and decryption key
- also, a secondary key is derived from the hash
- for being able to derive two independent keys in the same step, a hash function with wider output is used (SHA512)

**CBC data encryption**
- make use of an full block sized initialization vector (IV) instead of a nonce
- the IV is randomly set for every packet
- as the first data block is quite predictable (preamble, NIC's MACs), the IV gets encrypted as single block using the secondary key
- replay protection could be added to the IV and may be considered in Part 2

**Message authentication**
- a field for a message authentication code is filled with a ECBC-MAC, i.e. the last cipher block gets encrypted using the secondary key
- to ensure safety of ECBC-MAC, i.e. keep it prefix-free, data length is prepended to the data to be encrypted
- thus, the number of padded bytes at the last packet's end is not neccessary anymore and gets removed
- verification of the message authentication code was not first priority and may be included in Part 2

**Any other business**
- version number increased to 2 as packet design changed
- fixes a packet loss issue - at least for local edges
- author's first pull request ever - please be kind :)

### DIFF
--- a/benchmark.c
+++ b/benchmark.c
@@ -35,7 +35,7 @@ static int gettimeofday(struct timeval *tp, void *tzp)
 	time_t clock;
 	struct tm tm;
 	SYSTEMTIME wtm;
-	GetSystemTime(&wtm);
+	GetLocalTime(&wtm);
 	tm.tm_year = wtm.wYear - 1900;
 	tm.tm_mon = wtm.wMonth - 1;
 	tm.tm_mday = wtm.wDay;
@@ -89,7 +89,7 @@ int main(int argc, char * argv[]) {
   transop_twofish_setup_psk(&transop_twofish, 0, encrypt_pwd, sizeof(encrypt_pwd)-1);
   memset(&transop_aes_cbc, 0, sizeof(transop_aes_cbc));
   transop_aes_init(&transop_aes_cbc);
-  transop_aes_setup_psk(&transop_aes_cbc, 0, encrypt_pwd, sizeof(encrypt_pwd)-1);
+  transop_aes_setup_psk(&transop_aes_cbc, 0, encrypt_pwd, sizeof(encrypt_pwd)-1,1);
 
   /* Run the tests */
   run_transop_benchmark("transop_null", &transop_null, pktbuf, c);
--- a/benchmark.c
+++ b/benchmark.c
@@ -35,7 +35,7 @@ static int gettimeofday(struct timeval *tp, void *tzp)
 	time_t clock;
 	struct tm tm;
 	SYSTEMTIME wtm;
-	GetLocalTime(&wtm);
+	GetSystemTime(&wtm);
 	tm.tm_year = wtm.wYear - 1900;
 	tm.tm_mon = wtm.wMonth - 1;
 	tm.tm_mday = wtm.wDay;
--- a/edge_utils.c
+++ b/edge_utils.c
@@ -19,10 +19,6 @@
 #include "n2n.h"
 #include "lzoconf.h"
 
-#ifdef WIN32
-#include <process.h>
-#endif
-
 #ifdef __ANDROID_NDK__
 #include "android/edge_android.h"
 #include <tun2tap/tun2tap.h>
@@ -36,7 +32,7 @@
 #define REGISTER_SUPER_INTERVAL_DFL     60 /* sec */
 #endif /* #if defined(DEBUG) */
 
-#define REGISTER_SUPER_INTERVAL_MIN     5    /* sec */
+#define REGISTER_SUPER_INTERVAL_MIN     20   /* sec */
 #define REGISTER_SUPER_INTERVAL_MAX     3600 /* sec */
 
 #define IFACE_UPDATE_INTERVAL           (30) /* sec. How long it usually takes to get an IP lease. */
@@ -659,8 +655,11 @@ int edge_init_twofish_psk(n2n_edge_t * eee, uint8_t *encrypt_pwd,
 
 int edge_init_aes_psk(n2n_edge_t * eee, uint8_t *encrypt_pwd,
 		      uint32_t encrypt_pwd_len) {
+
+traceEvent (TRACE_ERROR, "!!! !!! !!! %d", eee->preferred_aes);
+
   return transop_aes_setup_psk(&(eee->transop[N2N_TRANSOP_AESCBC_IDX]),
-			       0, encrypt_pwd, encrypt_pwd_len);
+			       0, encrypt_pwd, encrypt_pwd_len, eee->preferred_aes);
 }
 
 /* ************************************** */
@@ -1694,11 +1693,11 @@ const char *random_device_mac(void)
       continue;
     }
 #ifdef WIN32
-#define random rand
+#define random() rand()
 #endif
     mac[i] = key[random() % sizeof(key)];
 #ifdef WIN32
-#undef random
+#undef random()
 #endif
   }
   mac[sizeof(mac) - 1] = '\0';
@@ -1747,10 +1746,8 @@ int quick_edge_init(char *device_name, char *community_name,
     
     /* allow multiple sockets to use the same PORT number */
     setsockopt(eee.udp_multicast_sock, SOL_SOCKET, SO_REUSEADDR, &enable_reuse, sizeof(enable_reuse));
-#ifdef SO_REUSEPORT /* no SO_REUSEPORT in Windows / old linux versions */
     setsockopt(eee.udp_multicast_sock, SOL_SOCKET, SO_REUSEPORT, &enable_reuse, sizeof(enable_reuse));
-#endif
-
+    
     mreq.imr_multiaddr.s_addr = inet_addr(N2N_MULTICAST_GROUP);
     mreq.imr_interface.s_addr = htonl(INADDR_ANY);
     if (setsockopt(eee.udp_multicast_sock, IPPROTO_IP, IP_ADD_MEMBERSHIP, &mreq, sizeof(mreq)) < 0) {
--- a/transform_aes.c
+++ b/transform_aes.c
@@ -32,7 +32,10 @@
 
 #define N2N_AES_TRANSFORM_VERSION       2  /* version of the transform encoding */
 #define N2N_AES_IVEC_SIZE               16 /* Enough space for biggest AES ivec */
-#define N2N_AES_MSGAUTH_SIZE		16 /* size of the message authentication code length, from 0 ... 16 = AES_BLOCK_SIZE */
+#define N2N_AES_MSGAUTH_SIZE		4 /* size of the message authentication code length, from 0 ... 16 = AES_BLOCK_SIZE */
+
+#define N2N_MAX_PACKET_DELAY		16000000 /* the maximum allowable time stamp difference time in microseconds = 16 sec */
+#define N2N_MAX_PACKET_DEVIATION_TIME	160000 /* the maximum allowable time stamp difference [usec] a packet may be earlier than another = 160 ms*/
 
 typedef unsigned char n2n_aes_ivec_t[N2N_AES_IVEC_SIZE];
 
@@ -41,11 +44,10 @@ struct sa_aes
     n2n_cipherspec_t    spec;           /* cipher spec parameters */
     n2n_sa_t            sa_id;          /* security association index */
     AES_KEY             enc_key;        /* tx key */
-    n2n_aes_ivec_t      enc_ivec;       /* tx CBC state */
     AES_KEY             dec_key;        /* tx key */
-    n2n_aes_ivec_t      dec_ivec;       /* tx CBC state */
     AES_KEY		enc_key_2;	/* secondary key for IV encryption and msg auth signing */
     AES_KEY		dec_key_2;	/* secondary key for IV encryption and msg auth signing */
+    uint64_t		prev_time_stamp;/* time stamp [usec] of previous packet sent from the associated node */
 };
 
 typedef struct sa_aes sa_aes_t;
@@ -67,8 +69,7 @@ struct transop_aes
     /* PSK mode only */
     int                 psk_mode;
     u_int8_t            mac_sa[N2N_AES_NUM_SA][N2N_MAC_SIZE]; /* this is used as a key in the sa array */
-    uint8_t             *encrypt_pwd;
-    uint32_t            encrypt_pwd_len;
+    uint8_t             *encrypt_pwd;    uint32_t            encrypt_pwd_len;
     size_t              sa_to_replace;
 };
 
@@ -213,6 +214,7 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
     uint8_t assembly[N2N_PKT_BUF_SIZE];
     uint16_t * data_len;
     uint8_t full_msg_auth[AES_BLOCK_SIZE];
+    uint8_t temp_iv[AES_BLOCK_SIZE];
 
     if ( (in_len + TRANSOP_AES_DATA_LEN) <= N2N_PKT_BUF_SIZE )
     {
@@ -250,22 +252,52 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
 	    memcpy( assembly + TRANSOP_AES_DATA_LEN , inbuf, in_len );
             len2 = ( ( ( len - 1) / AES_BLOCK_SIZE) + 1 ) * AES_BLOCK_SIZE; /* Round up to next whole AES block size */
 
+            /* generate random padding for the last plain text block and also */
             /* generate a random initialization vector for CBC directly to its field in the buffer */
             /* only proceed if random values are cryptographically safe (return value == 1) */
-            if (RAND_bytes(outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, N2N_AES_IVEC_SIZE) == 1)   /* get a cryptographically safely randomized initialization vector */
+            if ( (RAND_bytes(assembly + len + 1, len2 - len) == 1) &&
+                 (RAND_bytes(outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, N2N_AES_IVEC_SIZE) == 1) )
             {
-                /* sa->enc_ivec gets changed during cbc encryption procedure; to keep the original value safe (it is needed later), copy it from its field in the buffer */
-                memcpy((sa->enc_ivec), outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, N2N_AES_IVEC_SIZE );
+                /* encode system time to IV for replay attack protection */
+                /* get current system time (for international use: not local time) to encode it in IV */
+                struct timeval tod;
+                uint64_t micro_seconds;
+                gettimeofday (&tod, NULL);
+                /* We will calculate and put the microseconds since 1970 leftbound into the IV.
+                   As microseconds fraction never exceeds 1,000,000 , a max of 20 bits is
+                   used to encode the value tv_usec. 32 bits are used for tv_sec.
+                   Thus, micro_seconds' 12 most significant bits are not used.
+                   As we do not want 12 definetely zeroed bits IV, some bit masking is required
+                   to only copy the 52 (least) significant bits leaving the other 12 random as before
+                   - converting in host byte order to make sure bitwise masking is applied correctly;
+                   shifting might be independet from byte order, though. */
+                micro_seconds = be64toh(*(uint64_t*)(outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE));
+                micro_seconds = htobe64( ( (tod.tv_sec * 1000000 + tod.tv_usec) << 12)
+                                         | (0x0000000000000FFF & micro_seconds));
+                memcpy (outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, &micro_seconds, sizeof (micro_seconds));
+                /* nybblewise (4 bit portions), the whole 128 bit IV should look as follows:
+                   MMMMMMMMMMMMMrrrrrrrrrrrrrrrrrrr
+                   M = 52 bit value of microseconds since 1970
+                   r = random data */
+                /* to prevent an all too predictable content of this modified IV,
+                   before use it gets encrypted using K1 directly in the buffer */
+                AES_encrypt (outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, /* src: IV position in outbuffer */
+                             outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE , /* dst: same place as src */
+                             &(sa->enc_key)); /* using key K1 */
+                /* end of preparatory action against replay attacks */
+
+                /* make a copy of IV to the field provided to AES_cbc_encrypt, it gets corrupted during cbc mode encryption */
+                memcpy(temp_iv, outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, N2N_AES_IVEC_SIZE );
 
                 /* encrypt the payload (including packet length)  */
                 AES_cbc_encrypt( assembly, /* source */
                                  outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE + N2N_AES_IVEC_SIZE, /* dest */
                                  len2, /* enc size */
                                  &(sa->enc_key), /* using key K1 */
-                                 &(sa->enc_ivec), /* the copy of IV */
+                                 temp_iv, /* the copy of IV */
 				 1); /* = encryption */
 
-                /* now that iv was used, it gets encrypted for transmission. */
+                /* now that IV was used, it gets encrypted for transmission. */
                 /* this is necessary because first block of plain text is quite predicable (length, preamble, NIC MACs) */
                 /* encrypt the initialization vector directly on its field in the buffer using secondary key K2 */
                 AES_encrypt (outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, /* src: IV field in the buffer */
@@ -294,7 +326,7 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
             }
             else
             {
-                traceEvent ( TRACE_ERROR, "encode_aes no random data available for initialization vector." );
+                traceEvent ( TRACE_ERROR, "encode_aes no random data available for padding and initialization vector." );
             }
         }
         else
@@ -358,13 +390,13 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
     transop_aes_t * priv = (transop_aes_t *)arg->priv;
     uint8_t assembly[N2N_PKT_BUF_SIZE];
     uint8_t assembly2[N2N_PKT_BUF_SIZE];
-    n2n_aes_ivec_t iv;
+    n2n_aes_ivec_t temp_iv;
+    n2n_aes_ivec_t actual_iv;
     uint8_t full_msg_auth[AES_BLOCK_SIZE];
 
-
     if ( ( (in_len - (TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE + N2N_AES_IVEC_SIZE)) <= N2N_PKT_BUF_SIZE ) /* Cipher text fits in assembly */ 
-         && (in_len >= (TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE + N2N_AES_IVEC_SIZE + AES_BLOCK_SIZE) ) /* Has at least all neccessary fields and at least one cipher block */
-        )
+        && (in_len >= (TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE + N2N_AES_IVEC_SIZE + AES_BLOCK_SIZE) ) /* Has at least all neccessary fields and at least one cipher block */
+       )
     {
         n2n_sa_t sa_rx;
         ssize_t sa_idx=-1;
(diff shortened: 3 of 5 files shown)


emanuele-f:
Thank you! The pull request is very clean, your contribution is very welcome. I'm adding some inline comments for the review. If possible, also provide some links to online resources which explain the concepts used, for example the "prefix free" part of ECBC-MAC. Here is a reminder for me of the encryption/decryption process (please correct me if something is wrong).

```
*Encryption*:
iv = randomIv()
aes_cbc_encrypt(iv, raw_packet, primary_key) -> encryped_payload
aes_encrypt(iv, secondary_key) -> encrypted_iv
aes_encrypt(last_block, secondary_key) -> message_mac

*Decryption*:
aes_decrypt(encrypted_iv, secondary_key) -> iv
aes_cbc_decrypt(iv, encryped_payload, primary_key) -> raw_packet

Authentication (todo):
```
~~aes_decrypt(last_encrypted_block, secondary_key) -> transmitted_mac~~
~~aes_encrypt(last_encrypted_block, secondary_key) -> expected_mac~~
```
transmitted_mac from "header" in front of IV
aes_cbc_encrypt(decrypted_packet, key) to another buffer -> aes_encrypt (last_packet_of that, secondary key) -> expected_mac

if transmitted_mac != expected_mac
  drop_packet()
```

Please also note that this adds a 34 byte overhead to every packet (reducing this may be desiderable). A comparison of the results of the `benchmark` program available in n2n would also be useful to understand the performance impact of the changes.

transform_aes.c
```diff
@@ -22,15 +22,18 @@
...
 #endif
 
 #define N2N_AES_NUM_SA                  32 /* space for SAa */
 
-#define N2N_AES_TRANSFORM_VERSION       1  /* version of the transform encoding */
-#define N2N_AES_IVEC_SIZE               32 /* Enough space for biggest AES ivec */
+#define N2N_AES_TRANSFORM_VERSION       2  /* version of the transform encoding */
+#define N2N_AES_IVEC_SIZE               16 /* Enough space for biggest AES ivec */
+#define N2N_AES_MSGAUTH_SIZE		16
+#define HASH_SIZE			64
```
emanuele-f: 64 is because of SHA512, please add a comment

transform_aes.c
```diff
@@ -162,28 +167,23 @@ static ssize_t aes_choose_rx_sa( transop_aes_t * priv, const u_int8_t * peer_mac
...
+ *  - a 128 bit inner message authentication code (last encrypted data block encrypted again using K2),
+ *  - the separately encrypted IV (using K2),
+ *  - the 16-bit packet length of the following data (encrypted together with...
+ *  - ...the following payload data (using K1).
+ *  with K1 = key1 and K2 = key2.
  *
- *  [V|SSSS|nnnnDDDDDDDDDDDDDDDDDDDDD]
- *         |<------ encrypted ------>|
+ *  [V|SSSS|MMMMMMMMMMMMMMMM|IIIIIIIIIIIIIIII|LLDDDDDDDDDDDDDDDDDDDDDDDDDDDDDD]
+ *         |< encrypted K2 >|< encrypted K2 >|<------- encrypted K1 --------->|
```
emanuele-f: Replace "encrypted K2" with "encrypted with K2" or similar to avoid confusion

transform_aes.c
```diff
@@ -162,28 +167,23 @@ static ssize_t aes_choose_rx_sa( transop_aes_t * priv, const u_int8_t * peer_mac
...
-    }
 }
 
 /** The aes packet format consists of:
  *
  *  - a 8-bit aes encoding version in clear text
  *  - a 32-bit SA number in clear text
- *  - ciphertext encrypted from a 32-bit nonce followed by the payload.
+ *  - a 128 bit inner message authentication code (last encrypted data block encrypted again using K2),
+ *  - the separately encrypted IV (using K2),
```
emanuele-f: add size (128 bit)

transform_aes.c
```diff
@@ -219,28 +220,37 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
+            len = in_len + TRANSOP_AES_DATA_LEN;
+
+            /* The assembly buffer is a source for encrypting data. The data length is
              * written in first followed by the packet payload. The whole
              * contents of assembly are encrypted. */
-            pnonce = (uint32_t *)assembly;
-            *pnonce = rand();
-            memcpy( assembly + TRANSOP_AES_NONCE_SIZE, inbuf, in_len );
+            data_len = (uint16_t *)assembly;
+            *data_len = in_len;
```
emanuele-f: In order to make connection between big endian and little endian devices work properly, this must be encoded in network byte order. Check out https://en.wikipedia.org/wiki/Endianness for details. Please replace `*data_len = in_len;` with `*data_len = htons(in_len);`. By the way, why sending payload length required here?

transform_aes.c
```diff
@@ -219,28 +220,37 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
 
-            /* Need at least one encrypted byte at the end for the padding. */
-            len2 = ( (len / AES_BLOCK_SIZE) + 1) * AES_BLOCK_SIZE; /* Round up to next whole AES adding at least one byte. */
-            assembly[ len2-1 ]=(len2-len);
-            traceEvent( TRACE_DEBUG, "padding = %u", assembly[ len2-1 ] );
+	    memcpy( assembly + TRANSOP_AES_DATA_LEN , inbuf, in_len );
+            len2 = ( ( ( len - 1) / AES_BLOCK_SIZE) + 1 ) * AES_BLOCK_SIZE; /* Round up to next whole AES block size */
+
+            RAND_bytes(&iv, 16);  /* get a cryptographically safely randomized initialization vector */
+            /* REVISIT: return value of RAND_bytes function should be 1 to continue cryptographically safe, otherwise exit */
```
emanuele-f: Please add failure check. In this case, the communication should not proceed

transform_aes.c
```diff
@@ -219,28 +220,37 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
+            memcpy((sa->enc_ivec), iv, 16 ); /* sa->enc_ivec gets changed during CBC encryption procedure, so keep a copy in iv */
```
emanuele-f: Is it necessary to provide the IV on every packet? In my understanding, the IV is only required during the decryption of the very first block, is this correct?

transform_aes.c
```diff
@@ -325,36 +338,34 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
...
                                      assembly, /* destination */
                                      len, 
                                      &(sa->dec_key),
-                                     sa->dec_ivec, 0 /* decrypt */ );
-
-                    /* last byte is how much was padding: max value should be
-                     * AES_BLOCKSIZE-1 */
-                    padding = assembly[ len-1 ] & 0xff; 
+                                     &(sa->dec_ivec), 0 /* decrypt */ );
+                    data_len = *(uint16_t*)assembly;
```
emanuele-f: For the endianess, replace with: `data_len = ntohs(*(uint16_t*)assembly;)`

transform_aes.c
```diff
@@ -325,36 +338,34 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
...
 
-                    if ( len >= (padding + TRANSOP_AES_NONCE_SIZE))
+                    // REVISIT: verifiy message authentication code by CBC'ing the decrypted block again and encrypting the last block using enc_key_2, and finally compare it to MSGAUTH in inbuf
```
emanuele-f: Since the MAC has already been added to the packet, please add the remaining checks to validate it

transform_aes.c
```diff
@@ -219,28 +220,37 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
+            len2 = ( ( ( len - 1) / AES_BLOCK_SIZE) + 1 ) * AES_BLOCK_SIZE; /* Round up to next whole AES block size */
+
+            RAND_bytes(&iv, 16);  /* get a cryptographically safely randomized initialization vector */
+            /* REVISIT: return value of RAND_bytes function should be 1 to continue cryptographically safe, otherwise exit */
+            memcpy((sa->enc_ivec), iv, 16 ); /* sa->enc_ivec gets changed during CBC encryption procedure, so keep a copy in iv */
 
-            memset( &(sa->enc_ivec), 0, sizeof(N2N_AES_IVEC_SIZE) );
             AES_cbc_encrypt( assembly, /* source */
-                             outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE, /* dest */
+                             outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE + N2N_AES_IVEC_SIZE, /* dest */
```
emanuele-f: Add a comment before each encryption/descryption function, explaining the operation performed (e.g. decrypting the iv, decrypting the packet payload)

transform_aes.c
```diff
@@ -219,28 +220,37 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
                              len2, /* enc size */
-                             &(sa->enc_key), sa->enc_ivec, 1 /* encrypt */ );
+                             &(sa->enc_key), &(sa->enc_ivec), 1 /* encrypt */ );
+
+            AES_encrypt (iv, /* encrpyt the original initialization vector */
+                         outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE, /* to the buffer */
+                         &(sa->enc_key_2) ); /* using the secondary key */
+            AES_encrypt (outbuf + len2 - AES_BLOCK_SIZE, /* encrypt the last cipher block again */
+                         outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE, /* to use it as message authentication code */
```
emanuele-f: Please provide links to online resources/explain why performing the operation on the last block is enough for the authentication of the whole packet

emanuele-f (changes requested):

Logan007:
Thank you for the feedback! I give additional comments as you suggested in reply to your inline comments above.

Authentication verification should happen differently than outlined above, because CBC decryption happens in "parallel", i.e. the last block does not change in decryption process if the first cipher block has altered, because the decryption of the last block only requires the ciper of the (n-1)th block. So, the scheme would look like:
```
*Authentication* (todo):
transmitted_mac from "header" in front of IV
aes_cbc_encrypt(decrypted_packet, key) to another buffer -> aes_encrypt (last_packet_of that, secondary key) -> expected_mac

if transmitted_mac != expected_mac
  drop_packet()
```
By the way, a few bytes per packet could be saved here if the message authentication code only contained the maybe first 32 bit of the two-times-encrypted last bock. That is a choice to make with a view to the desired level of security.

Still learning git, I wonder how I can make the additional changes to the code as suggested - your comments are very valuable. New Pull Request? Please advice.

transform_aes.c
Logan007: Okay, I will add a comment. Alternatively, I could just use SHA512_DIGEST_LENGTH throughout the code without definition here - found out about it later.

transform_aes.c
Logan007: I will think of something.

transform_aes.c
Logan007: OK.

transform_aes.c
Logan007: Thank you, I surely will change that.

The data length is required here in the first place because we need to keep the encrypted data "prefix free", meaning that no other ever encrypted message could be put in front of any other package encrypted data forming a valid packet again. To put packet's length upfront is an effective countermeasure to prevent this and therefore eliminate the possibility of a  variable-length attack described [here](https://en.wikipedia.org/wiki/CBC-MAC#Security_with_fixed_and_variable-length_messages).

transform_aes.c
Logan007: Sure.

transform_aes.c
Logan007: The IV is different for each packet. So, it must be provided once for every transmitted packet - at the beginning of the CBC decryption process.

It is generated and stored in "iv". Afterwards, it gets copied to enc_ivec. During the CBC decrpytion process, it seems that the content of enc_ivec gets changed with every single block inside the CBC crypto function. I guess this from some simple debugging showing that enc_ivec has a different content after the CBC function call. That's why we should keep it somewhere, here iv to use it later to put in the packet.

Ummm, maybe I could have directly generated it into the packet buffer and afterwards copy it into enc_ivec? How could byte order be correct?

Let me add that a random IV is not an issue. Au contraire! The encryption is necessary as the first data block is extremly predictable (preamble and MACs that are already known because transmitted in plain text as part of framing packet) - so we should not transmit IV in plain text and therefore scramble it. A possible attack as described [here](https://en.wikipedia.org/wiki/CBC-MAC#Allowing_the_initialization_vector_to_vary_in_value) is also encountered by the encryption of the IV.

transform_aes.c
Logan007: Thank you!

transform_aes.c
Logan007: Okay, will try!

transform_aes.c
Logan007: Sure.

transform_aes.c
Logan007: As CBC always uses the cipher of the block before to encrypt the next one, every cipher always depends on the block before and all other blocks before. Thus, the last block is a fingerprint of the whole encrypted data. Anyway, we can't just use it but need to encrypt it again using another key to prevent an attack described [here](https://en.wikipedia.org/wiki/CBC-MAC#Using_the_same_key_for_encryption_and_authentication).

emanuele-f:
Thank you, I've updated the comment above for the authentication part. In order to update this pull request you only need to commit the new changes to your branch. So, after making the changes, do a `git add ...your_changed_files...`, `git commit` and then push the changes to your online branch with `git push`. This will automatically update the pull request, since your online branch is now *bound* to this pull request

Logan007:
I have just finished implementing the changes you suggested and applied some obvious corrective changes.

You are right, concerning packet size, we now see larger packets: I would just put "minus 1" for dropping the padding length and "minus 4" for replacing the nonce. However, to keep it a little bit more flexible, I implemented the message authentication in a way that allows to change its length in bytes at compile time - just change the definition (0 ... 16, all seem to work well). As the length of the message authentication code is not explicitly transmitted, a change in message authentication code size causes incompatibility to clients with a different size. Maybe something to ponder for Part 2...

For now, there is just the need to agree on a reasonable value for it.

transform_aes.c
Logan007: For Part 2 I already took a note to remove enc_iv an dec_iv from the sa structure and use local variables only. As IVs are varying for every packet, there is no need to keep track of them in the more static sa structure.

transform_aes.c
Logan007: I just realized that the message authentication does not comprise the version and security association number. I will fix that and create a new commit (3rd).

Logan007:
The aes_version and sa number now are included in the message authentication by just xor'ing them before the message authentication code gets encrypted using K2. So, the accordingly updated message authentication verification scheme should look like this:

```
**transmitted_mac** from "header" in front of IV

aes_cbc_encrypt(decrypted_packet, key) to another buffer
-> xor'ing in the aes_version and sa field from the received packet to the last block
-> aes_encrypt (last_block_of that, secondary key) 
-> **expected_mac**

if transmitted_mac != expected_mac
  drop_packet()
```

Leftovers for Part 2 of AES revamp: Encounter Replay Attacks by modifying IV scheme (could remain compatible to this aes version) and clean up of sa structure.

I did some benchmarking (i7-5775C). The modifications indeed cost a bit of performance as follows:

Before modifcation (6.5 x slower than null):
```
./benchmark 
..........
[transop_null] 10000 times: (5255 -> 525 nsec each) 1551993236.392130 -> 1551993236.397385.
..........
[transop_twofish] 10000 times: (269364 -> 26936 nsec each) 1551993236.397415 -> 1551993236.666779.
..........
[transop_aes] 10000 times: (34192 -> 3419 nsec each) 1551993236.666785 -> 1551993236.700977.
```

After modification (15 x slower than null)
```./benchmark 
..........
[transop_null] 10000 times: (4033 -> 403 nsec each) 1551993168.241907 -> 1551993168.245940.
..........
[transop_twofish] 10000 times: (267498 -> 26749 nsec each) 1551993168.245970 -> 1551993168.513468.
..........
[transop_aes] 10000 times: (62269 -> 6226 nsec each) 1551993168.513472 -> 1551993168.575741.


```

So the gain of AES's hardware acceleration is melting a bit, but it is still way ahead of TF. Possible shortcuts could be taken by later maybe adding a command line parameters for fixed, lower keysize and disabled message authentication. Though, I would not recommend that at all from a security point of view.

Logan007:
Sorry, if this pull request now gets a little bit lengthy... I have just added replay protection by encoding a  time stamp to the initialization vector as outlined in #73.

There are two values to parametrize the replay protection (definitions at the top of transform_aes.c):

- One is the time frame in which packets with a certain time stamp are accepted by the receiver by comparing the timestamp applied by the sender to the receiver's system time. This is set to +/- 16 seconds, allowing for 32 seconds of time difference between the two clocks plus transmission time. In the age of ubiquitious NTP servers, this might be reasonable.
- Nodes keep track of the previous packet's time stamp and make sure, that only packets with higher time stamp are accepted - assuming that packets do not overtake each other on the line too often. But as exactly this seems to happen from time to time, there is a grace time set to 160 ms which seems to work well in most cases, at least in the author's setup.

Also, I changed the message authentication code length to 4 bytes. Would that be a reasonable value?

The replay prevention does not come for free: There is a slight performance slump as follows (15.8 x slower than null, see the post above):

```
./benchmark
..........
[transop_null] 10000 times: (3732 -> 373 nsec each) 1552294709.652039 -> 1552294709.655771.
..........
[transop_twofish] 10000 times: (268661 -> 26866 nsec each) 1552294709.655796 -> 1552294709.924457.
..........
[transop_aes] 10000 times: (74913 -> 7491 nsec each) 1552294709.924461 -> 1552294709.999374.
```

By the way, talking performance, I found that key size has a somewhat non-neglectable impact on performance. Changing the now hard-wired keysize from 256 to 128 bit makes it look as follows (1400 ns faster):

```
./benchmark
..........
[transop_null] 10000 times: (3230 -> 323 nsec each) 1552293781.135773 -> 1552293781.139003.
..........
[transop_twofish] 10000 times: (270719 -> 27071 nsec each) 1552293781.139030 -> 1552293781.409749.
..........
[transop_aes] 10000 times: (60573 -> 6057 nsec each) 1552293781.409753 -> 1552293781.470326.
```

AES version incremented to 3. **This shall conclude the changes I wanted to apply for now.** Please let me know what you think and please test. Do the times for replay attack protection work for you? What does it take to make this pull request acceptable? I would love to hear back from you to make n2n even better.

A next step (new pull request, later) could include implementation of support for aes version 2 by iffing out the message authentication, the replay attack protection and also maybe switching to a lower keysize - allowing for reasonable use on less performing systems, e.g. routers.

transform_aes.c
Logan007: Used SHA512_DIGEST_LENGTH.

Logan007:
Author's evaluation after five days: works flawless in a setup with one public supernode and four edges around the world.

Logan007:
I think there might be the need to make all the above more compatible with the existing dev branch versions. What do you think?

My thought is to add some automatism detecting the used version (1,2, or 3) per node by evaluating the first received packet and not allowing to go with less than that to prevent downgrade attacks. That way, the current dev version still would work with that. 

Also, I would like to hijack the "-A" option implemented by @emanuele-f . It could be extended to "-A 1", "-A 2", and "-A 3" to indicate the maximum AES packet version to use to send out from this edge node, still allowing other versions to be received. What would be a good default in case it is not given or a key file is used? 1 or 3? I might need a few hints to fully understand the mechanism how that information could get passed through to the transform_aes module.

Versions could be 

1. as in current dev (AES w/ CBC, 32 bit nonce, IV = 0) + [new] hashed password* + full key size* + length field in front of encrypted data* + randomized padding at data's end (not for packet length, if applicable)
2. as before + [new] full and random IV
3. as before + [new] replay attack protection

The three features marked with "*" could also be put to version 2 leaving version 1 nearly completely untouched and preserving full backward compatibility, if desired or necessary - which I would not recommend with a view to security as discussed in #68.

Worth further discussion?

emanuele-f:
Hi, this may be a good idea to specify different modes as this implementation requires more computation and leads to an increase in packet sizes. In my opinion the default mode should be a compromise between security and speed/size whereas a "paranoid" mode could be more security oriented. I would not add too much modes as it could confuse the users

Logan007:
Okay, I guess we just need to think of them as protocol versions. Will try to implement the above sketched, merging version (2+3) to just (2). Might take a while...

Last night, I had an even more enhanced idea of how to further ameliorate security... will open a new issue with that.

Logan007:
It took a while to restore backward compatibility towards the current dev branch aes encryption scheme which I shall call version 1.  All other changes to the AES scheme (random IV, message authentication, replay protection) now are bundled in encryption scheme version 2.

V2 is optional.  It can be activated for outgoing packets by using the "-A2" option instead of "-A" or "-A1" (new).  I was not able to get a blank between the A and the version number as it is an optional parameter argument for the sake of compatibility to "-A" only.  Incoming packets are always parsed, no matter what version.  That way, mixed networks are possible.

There were two things which I originally wanted to implement, but did not do so: 
1. The key length is still depending on the password size also in V2 to use full key spectrum.  That way, all 128 bit keys and 192 bit keys as well as 256 bit keys get a chance which enhances the usable key space.  By the way, for better entropy, V2 requires 4 password characters per one key byte.  So, to make use of a 256 bit key a password needs at least a 128 characters.  V1 remains unchanged in this respect.
2. A host based flag to keep track of the encryption scheme used was not implemented as there might be implications in key schedule mode which I could not foresee yet.  If it proves to be safe, it could be implemented anytime later without implications on the so far implemented schemes V1 or V2.

The code for encode and decode was completely linearized - I am sorry, Mr. [Dijkstra](https://homepages.cwi.nl/~storm/teaching/reader/Dijkstra68.pdf).  I did not see any reasonable way to indent error dependent parts of code across version related switch statements.  Some may find the code more readable now.

There is one drawback:  benchmark for AES is not working anymore - and I cannot really explain why. 
 Usually, I would not publish half-baked code, but in this case I hope for some hints or help.

Any thoughts?

edge_utils.c
```diff
@@ -659,8 +655,11 @@ int edge_init_twofish_psk(n2n_edge_t * eee, uint8_t *encrypt_pwd,
 
 int edge_init_aes_psk(n2n_edge_t * eee, uint8_t *encrypt_pwd,
 		      uint32_t encrypt_pwd_len) {
+
+traceEvent (TRACE_ERROR, "!!! !!! !!! %d", eee->preferred_aes);
```
Logan007, line 659: This will be removed during next commit. :)

edge_utils.c
```diff
@@ -36,7 +32,7 @@
 #define REGISTER_SUPER_INTERVAL_DFL     60 /* sec */
 #endif /* #if defined(DEBUG) */
 
-#define REGISTER_SUPER_INTERVAL_MIN     5    /* sec */
+#define REGISTER_SUPER_INTERVAL_MIN     20   /* sec */
```
emanuele-f, line 35: Please don't change in this pull request.

benchmark.c
```diff
@@ -89,7 +89,7 @@ int main(int argc, char * argv[]) {
   transop_twofish_setup_psk(&transop_twofish, 0, encrypt_pwd, sizeof(encrypt_pwd)-1);
   memset(&transop_aes_cbc, 0, sizeof(transop_aes_cbc));
   transop_aes_init(&transop_aes_cbc);
-  transop_aes_setup_psk(&transop_aes_cbc, 0, encrypt_pwd, sizeof(encrypt_pwd)-1);
+  transop_aes_setup_psk(&transop_aes_cbc, 0, encrypt_pwd, sizeof(encrypt_pwd)-1,1);
```
emanuele-f, line 92: It's better to use a define AES_SIMPLE_VERSION or AES_VERSION_1

edge_utils.c
```diff
@@ -1702,11 +1701,11 @@ const char *random_device_mac(void)
       continue;
     }
 #ifdef WIN32
-#define random rand
+#define random() rand()
```
emanuele-f, line 1704: please explain the change

edge_utils.c
```diff
@@ -1755,10 +1754,8 @@ int quick_edge_init(char *device_name, char *community_name,
     
     /* allow multiple sockets to use the same PORT number */
     setsockopt(eee.udp_multicast_sock, SOL_SOCKET, SO_REUSEADDR, &enable_reuse, sizeof(enable_reuse));
-#ifdef SO_REUSEPORT /* no SO_REUSEPORT in Windows / old linux versions */
     setsockopt(eee.udp_multicast_sock, SOL_SOCKET, SO_REUSEPORT, &enable_reuse, sizeof(enable_reuse));
-#endif
-
+
```
emanuele-f, line 1758: you have removed code here, SO_REUSEPORT, undo

transform_aes.c
```diff
@@ -162,7 +189,7 @@ static ssize_t aes_choose_rx_sa( transop_aes_t * priv, const u_int8_t * peer_mac
  */
 static size_t aes_best_keysize(size_t numBytes)
 {
-    if (numBytes >= AES256_KEY_BYTES )
+    if (numBytes >= AES256_KEY_BYTES)
```
emanuele-f, line 192: Since hashing is now performed on the key, if you think that AES256 would fit both the computational use case of a desktop pc as well as a single core SBC board, then you can always return AES256_KEY_BYTES here

transform_aes.c
```diff
@@ -196,62 +236,174 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
+       at buffer's end as well as the IV, also choose key */
+    AES_KEY key;
+    switch (priv->version)
+    {
+	case 1 :
+ 	{
+	    assembly[ len2-1 ]=(len2-len);
             traceEvent( TRACE_DEBUG, "padding = %u", assembly[ len2-1 ] );
+	    /* set IV to 0 for v1 */
+	    memset (iv, 0, N2N_AES_IVEC_SIZE);
```
emanuele-f, line 320: Version 1 should have IV properly initialized, not 0. This must be fixed in a lightweight way.

transform_aes.c
```diff
@@ -196,62 +236,174 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
+	    gettimeofday (&tod, NULL);
+	    /* We will calculate and put the microseconds since 1970 leftbound into the IV.
+	       As microseconds fraction never exceeds 1,000,000 , a max of 20 bits is
+	       used to encode the value tv_usec. 32 bits are used for tv_sec.
+	       Thus, micro_seconds' 12 most significant bits are not used.
+	       As we do not want 12 definetely zeroed bits IV, some bit masking is required
+	       to only copy the 52 (least) significant bits leaving the other 12 random as before
+	       - converting in host byte order to make sure bitwise masking is applied correctly;
+	       shifting might be independet from byte order, though. */
+	    micro_seconds = be64toh(*(uint64_t*)(outbuf + TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + N2N_AES_MSGAUTH_SIZE));
```
emanuele-f, line 349: Please also consider this issue with 32bit seconds: https://en.wikipedia.org/wiki/Year_2038_problem . Since we have extra bits it's a waste not to use them and instead make they random. Moreover, do we really need microseconds accuracy here? What if the clocks of the two machines are not synchronized?

emanuele-f:
Thank you for the work so far. I have added some comments to the code. I'm still not so much convinced by the changes as they add a lot of complexity (both coding and conceptual) although of course solve real problems. I'd have preferred a prior discussion on the design of all of this (maybe via a more direct communication channel) as this is now in a state "accept all and trust me that it works or throw it away". We need other users feedback on this, and also a feedback of my colleague @lucaderi .

Logan007:
This is not meant to be be "take it or leave it".   I will remain open to apply any changes suggested and for discussion or explanations of course.  Though, I too was wondering why nobody else participated in the discussion so far. 

Thank you for your comments. Tomorrow, I will have a closer look and explain more, if required.

I just want to contribute to the project where I could be of use.  You can trust me :)

_PS: Please consider this more like a "just take what you need and leave the rest"-option._

emanuele-f:
Probably the crypto aspects are obscure for a lot of people. Anyway as I said, your contribution is welcome, we only have to tune clarity, simplicity and performance requirements with the security aspect.

benchmark.c
Logan007: Good idea, I will add that - here and also throughout transform_aes.

edge_utils.c
Logan007: Ummm, I honestly can't. Something weird happened to edge_utils. I will have a deeper look.

edge_utils.c
Logan007: I will check, see above.

transform_aes.c
Logan007: That was my first thought, too. But to remain completely compatible to current builds (what _now_ is treated as V1), I suggest to keep it, otherwise compatibility of current dev version "-A" and this pull request's "-A1" will break.

As we saw in the benchmarks above, AES with 128 bit key is still 23% faster. So, my idea was, to leave it to the user to decide: by choosing the password length one determines the key length an thus computational load.

Based on explanation from #68, a 256 bit key is space is better than a 128 bit key space only. During the last days, when I found that the three zeroed keys (128, 192, 256 bits respectively) seem to deliver different encryptions, I had the idea that using the full available keyspace (all 128, 192 **and** 256 bitted keys) would be more than just restricting to the full key size of 256 bit.

That is why I left this routine also for V2. An attacker would not know what size of key we have used, he would need to brute force all 128, 192 and 256 keys instead of all 256 bit keys only - which of course already are a lot.

transform_aes.c
Logan007: I left it initialized to 0  for the sake of compatibility. If design decision is, that a new program version does not necessarily need to be compatible, the user has to replace the software on all hardware platforms used. This might be an issue, as some might use other forks that seem specifically designed for their needs or other platforms, e.g. Hin2n for unrooted android. But that is a decision to make.

Basically, an IV should not repeat under the use of the same key. If this is not possible, it should remain constant. That's why it is chosen to be 0. Another constant pattern would be as good as that, too.

The 32 bit nonce (I reintroduced for V1 for compatibility) serves "somehow as a kind of IV" because it prepends the first data bock that gets encrypted in cipher block _chaining_ mode - but it only is 32 bit wide. A full "random" IV as in V2 requires full block size of 128 bit. So the question here is, how to specify "lightweight"?

If it is decided that compatibility is not an issue, for V1, we could remove the nonce and switch to a random IV instead - it must be sent with the packet anyway; encrypted with a "not the data encrytion key (K2)" as first data block is highly predictable and we cannot use the data key. To stay lightweight, we could use half block size (64 bit) and use some hash of the secret password to fill up another 64 bit to full block size. That way, we only need to transmit the random part, IV would be quite random, and the constant part of IV would be password dependent and not hardwired in the code. That might be a compromise taking all concerns into account to at least some extent. That still does not allow for replay attack protection as in V2.

transform_aes.c
Logan007: I actually have thought about it. What will happen is that the time counter turns over to zero. A reboot/restart after that will fix it.

IVs should not just not repeat under the same key, they also should be as unpredictable as possible. That is achieved by 2 means: Using as much randomness as space allows (thus only using the neccessary amount of the highly predictable time bits) and the following encryption step using K1 before feeding the IV to the cbc encryption.

The time stamp part serves two purposes in replay attack protection:

1. It makes sure that the packet was sent not too long ago. Taking into account that packets may take some time and also, that clocks might be not completely in sync, there is a grace time of +/- 16 seconds (= 32 seconds) for comparison of _clocks of different computers_, specified at the top of transform_aes. It is user's deliberately made decision to use V2 by setting the parameter. The user should be aware that higher security requirements such as replay attack protection also have some boundary conditions such as somewhat synchronous clocks in this case, I will add a section to the man page. In times of ubiquitous ntp servers, I think this is a feasible approach. Let me know what you think.

2. By comparing the packet time stamps to each other, i. e. comparing time stamps from the _same clock_, we make sure that packets are in sequence. The microseconds are a kind of counter. To be specific, we make sure that they are "somewhat" in sequence as we allow for some grace time of 160 ms because packets sometimes may overtake each other on the line or even the packet stack. This time is defined at the top of transform_aes, too. In my worldwide setup, this value yielded good results. The purpose is to prevent fast replay attacks inside the 32 seconds window mentioned above. That's what we need the high accuracy for. If tests show that the 160 _milli_seconds order is a sufficent reasonable grace time dimension, we could lower the accuracy of the time stamp further by the factor of 128...1024  microseconds by shifting it 7...10 bits further to the right and allowing more space for random bits at the same time.

Logan007:
@emanuele-f, I think a design decision is required. There are several options on the table that need to be discussed in terms of compatability, security, performance and flexibility. At the end, these options could be handled in parallel as differenct encryption scheme versions. My guess is that a "one size fits it all" will probably not work out to cover user's needs in the long term.

Again, there is plenty of time and no need to rush these changes in the dev or any another branch as they should be considered security _enhancements_. Being an active n2n user, I will be around for while. And as this is a field where I feel that I am somewhat able to contribute, I am glad for the opportunity to give back to the open source community whose software I usually use.

If you prefer a more direct way of discussion, we could meet via teamspeak in the public lounge of www.awesomebros.net during the weekend. What time would be good for you?

edge_utils.c
Logan007: I can't remember having changed that.  Maybe, that happened when I tried to have git resolve conflicts between latest dev and this pull request? I will put it manually to the state it should be.

Logan007:
So, any input on compatibility and whether n2n should be kept compatible to current aes implementation?

emanuele-f:
The -A parameter of n2n has only been introduced recently, and there is no need to keep compatibility for now, in particular as no stable version of n2n has been released with the -A support

Logan007:
And what about compatibility to aes from keyfiles? Those keys also use the value '1' in the version field?

emanuele-f:
I think keyfiles are only used by very few people, if any. The initialization fix I made should have already broken the interoperability between old and new version.

Logan007:
I guess the dev branch has developed too far away in the meantime. Just take any idea or whatever you might need from the code. In whatever way you plan to implement any of these features (IV, replay protection, MAC), I would recommend to keep it flexible and extensible in terms of security, e.g. keysize, IV length, MAC tag length.

Shall we close this?

PR #102, opened 2019-05-01
Add feature: selectable tap adapter in windows

HorseLuke:
I have to use multiple TAP devices in Windows, so decide to deal with feature request in issue #1.

But I am not familiar with C, may be with some vuls. Code review is really needed.

Limitation:
Length of tap name in Windows should lesser than 16 (as const N2N_IFNAMSIZ defined)


emanuele-f:
Implemented in 69e1d50ca3e9eac400e3a6ff96b39f33d91f7464

Issue #106, opened 2019-05-17
Pending_peers is never purged


realjiangms:
In int run_edge_loop(n2n_edge_t * eee, int *keep_running) we purge peer lists as following:

l1:  numPurged =  purge_expired_registrations(&(eee->known_peers));
l2:   numPurged += purge_expired_registrations(&(eee->pending_peers));

However "static time_t last_purge"  in size_t purge_expired_registrations(struct peer_info ** peer_list) is global shared and will be updated in l1. So eee->pending_peers is never purge.

Enable -v on edge could easily notice that

realjiangms:
I raised an PR #109 as suggestion of solution

emanuele-f:
Pull request merged

PR #109, opened 2019-05-20
Fix purge expired registrations kickoff 2.4 stable

realjiangms:
It is based on stable 2.4 branch. 

Fix the pure code issue mentioned in #106
Also fix the "send_register" to add dst mac in packet.

Additional notes for timeout arguments in purge_expired_registrations:

1. Actually eee->pending_peers should has different timeout as eee->known_peers. For example, if REGISTER message is lost during UDP transpot (somehow first REGISTER on one of edge in my test rig has high possibility of lost), entry in pending_peers should be purged more frequently so that we can retry REGISTER soon.

2. If we specify mac address in command line, it can be simply broken if we ctrl-C and restart an edge instance. That's because the local port as well as external port is changed after local edge is restarted but the mapping of mac -> ip:port is not changed on peer edge. Therefore peer edge actually can no longer send message to local edge. Peer edge doesn't update the mapping when a message from local edge arrives because it is known and the message is from the supernode. Local edge can't perform register process because there's no message from peer edge arrives (even no message forwarded from supernode, because local edge is 'known' in peer edge and peer edge always try to send message by staled  'known' port), so local edge always send message through supernode. 
-> We stuck in this situation until eee->known_peers on peer edge purge the local edge with staled 'known' port. But if we set timeout on eee->known_peers with short several seconds, the connection could be recovered in several seconds (instead of 20 minutes by default).

In this PR I leave timeout arguments in purge_expired_registrations as original default but we can consider to make them as configurable.

### DIFF
--- a/edge_utils.c
+++ b/edge_utils.c
@@ -1419,6 +1419,8 @@ int run_edge_loop(n2n_edge_t * eee, int *keep_running) {
   size_t numPurged;
   time_t lastIfaceCheck=0;
   time_t lastTransop=0;
+  time_t last_purge_known = 0;
+  time_t last_purge_pending = 0;
 #ifdef __ANDROID_NDK__
     time_t lastArpPeriod=0;
 #endif
@@ -1502,8 +1504,8 @@ int run_edge_loop(n2n_edge_t * eee, int *keep_running) {
 
     update_supernode_reg(eee, nowTime);
 
-    numPurged =  purge_expired_registrations(&(eee->known_peers));
-    numPurged += purge_expired_registrations(&(eee->pending_peers));
+    numPurged =  purge_expired_registrations(&(eee->known_peers), &last_purge_known);
+    numPurged += purge_expired_registrations(&(eee->pending_peers), &last_purge_pending);
 
     if(numPurged > 0) {
       traceEvent(TRACE_NORMAL, "Peer removed: pending=%u, operational=%u",
--- a/n2n.c
+++ b/n2n.c
@@ -307,18 +307,17 @@ void peer_list_add(struct peer_info * * list,
 }
 
 
-size_t purge_expired_registrations(struct peer_info ** peer_list) {
-  static time_t last_purge = 0;
+size_t purge_expired_registrations(struct peer_info ** peer_list, time_t* p_last_purge) {
   time_t now = time(NULL);
   size_t num_reg = 0;
 
-  if((now - last_purge) < PURGE_REGISTRATION_FREQUENCY) return 0;
+  if((now - (*p_last_purge)) < PURGE_REGISTRATION_FREQUENCY) return 0;
 
   traceEvent(TRACE_INFO, "Purging old registrations");
 
   num_reg = purge_peer_list(peer_list, now-REGISTRATION_TIMEOUT);
 
-  last_purge = now;
+  (*p_last_purge) = now;
   traceEvent(TRACE_INFO, "Remove %ld registrations", num_reg);
 
   return num_reg;
--- a/n2n.h
+++ b/n2n.h
@@ -297,7 +297,7 @@ size_t peer_list_size( const struct peer_info * list );
 size_t purge_peer_list( struct peer_info ** peer_list, 
                         time_t purge_before );
 size_t clear_peer_list( struct peer_info ** peer_list );
-size_t purge_expired_registrations( struct peer_info ** peer_list );
+size_t purge_expired_registrations( struct peer_info ** peer_list, time_t* p_last_purge );
 
 /* version.c */
 extern char *n2n_sw_version, *n2n_sw_osName, *n2n_sw_buildDate;
--- a/sn.c
+++ b/sn.c
@@ -859,6 +859,7 @@ static int run_loop( n2n_sn_t * sss )
 {
   uint8_t pktbuf[N2N_SN_PKTBUF_SIZE];
   int keep_running=1;
+  time_t last_purge_edges = 0;
 
   sss->start_time = time(NULL);
 
@@ -934,7 +935,7 @@ static int run_loop( n2n_sn_t * sss )
 	  traceEvent( TRACE_DEBUG, "timeout" );
         }
 
-      purge_expired_registrations( &(sss->edges) );
+      purge_expired_registrations( &(sss->edges), &last_purge_edges );
 
     } /* while */
 


emanuele-f:
Thank you. Please separate the removal of static variables fix from the other enhancements, as the purge fix will go into stable whereas the other enhancements will go in dev as they may have collateral effects.

Great finding about the `reg.dstMac` not being set, it is probably the missing part which prevents proper p2p communication with current n2n! Please provide a pull request for this and I'll check asap.

Regarding the timeout issue, please check out 49fd9fc5daa1ba049adb1eae976b0516c3848d71, this should already reduce to about 10 seconds the dead time before the connection is resumed after local port change.

realjiangms:
I'll separate them and use this PR for 1st commit only. 

Regarding the 2nd commit on reg.dstMac, I found it is not enough. Even we call check_peer on REGISTER, there might already been an entry in pending_list with incorrect peer-ip-and-port, and the correct sock information from REGISTER will be ignored. We need to also adjust the policy in try_send_register to update the sock information

emanuele-f:
Please note that for the purge fix there is no need to add the frequency and registration_timeout parameters.
For the registration issue, do this still applies to the dev branch?

realjiangms:
I checked  [49fd9fc](https://github.com/ntop/n2n/commit/49fd9fc5daa1ba049adb1eae976b0516c3848d71). Yes it resolves the 2nd issue I mentioned. However the 1st REGISTER issue still exists. Since you introduced scan->timeout in that commit, I would suggest to also used it to retry/expire pending_peers.

Actually, I would argue for purge_expired_registrations. It is already-exists mechanism to remove stale entries from known_peers and pending_peers, it just has too long interval and have some problem in implementation. We should use it for expiration of peer_info lists. If we have additional timeout in peer_info structure itself and use that to expire entries, I think we can just completely remove it, at least in edge side.

By the way, I figured out why first REGISTER always fail on one side. Because both edges are IP-Restricted Cone NAT or Port-Restricted Cone NAT, only the later REGISTER could arrive. So current implementation only get one-way p2p transportation if both edges are IP-Restricted Cone NAT or Port-Restricted Cone NAT. Actually once one direction p2p tunnel is built, another direction should be easily built. We potentially have many chance to do it, when we receive Rx REGISTER, and when we receive each p2p message from peer. But we:
1. Retry REGISTER after 20minutes
2. Never really call check_peer on Rx REGISTER

emanuele-f:
> I checked [49fd9fc](https://github.com/ntop/n2n/commit/49fd9fc5daa1ba049adb1eae976b0516c3848d71). Yes it resolves the 2nd issue I mentioned. However the 1st REGISTER issue still exists.

Ok, so that should be fixed by adding the `reg.dstMac`.

> Since you introduced scan->timeout in that commit, I would suggest to also used it to retry/expire pending_peers.
> 
> Actually, I would argue for purge_expired_registrations. It is already-exists mechanism to remove stale entries from known_peers and pending_peers, it just has too long interval and have some problem in implementation. We should use it for expiration of peer_info lists. If we have additional timeout in peer_info structure itself and use that to expire entries, I think we can just completely remove it, at least in edge side.

Even after the `scan->timeout` fix I made, we should purge peers which become idle, so I think purge_expired_registrations is still needed.

> 
> By the way, I figured out why first REGISTER always fail on one side. Because both edges are IP-Restricted Cone NAT or Port-Restricted Cone NAT, only the later REGISTER could arrive. So current implementation only get one-way p2p transportation if both edges are IP-Restricted Cone NAT or Port-Restricted Cone NAT. Actually once one direction p2p tunnel is built, another direction should be easily built. We potentially have many chance to do it, when we receive Rx REGISTER, and when we receive each p2p message from peer. But we:
> 
>     1. Retry REGISTER after 20minutes
> 
>     2. Never really call check_peer on Rx REGISTER

This should be fixed with your pull request right?

realjiangms:
>Even after the scan->timeout fix I made, we should purge peers which become idle, so I think purge_expired_registrations is still needed.

Your scan->timeout did purge 'current' entry. But I would agree that purge_expired_registrations checks all entries.

> 1. Retry REGISTER after 20minutes
>
> 2. Never really call check_peer on Rx REGISTER
>This should be fixed with your pull request right?
>

Actually No. Because your fix with timeout doesn't exists in stable-2.4 branch.

--------------------
In addition, I just found another problem. I used to try to build p2p connection between my home server (with PortRestrictedConeNat) and mobile (with mobile network, it seems Addr Rest Cone Nat). However it never works. I usually use -c mynetwork -k encryptme as readme. But I just tried it again without 
-k encryptme and it works! Not sure what's the difference. I captured network by wireshark on my labtop and don't find different in UDP data size. I'm trying to figure out why. (please don't ask me whether or not packets arrives my home router. I can't capture on it as it is provided by ISP and they don't give me admin privilege)

realjiangms:
Updated PR as your comments for only purge_expired_registrations

emanuele-f:
Thank you

emanuele-f:
In b6d12c009c2416b2454d8674cc9ce7baa43f7048 I've lowered the registration timeout to 1 minute. Please open new issues/pull requests for the other stuff

emanuele-f:
@realjiangms I've made another fix for P2P connections in https://github.com/ntop/n2n/commit/3aec02d3e64be01c6e3ec5caeed7e61df31cec46 . Please check out the new dev, all the p2p issues I had are now fixed

realjiangms:
Will you consider port [b6d12c0](https://github.com/ntop/n2n/commit/b6d12c009c2416b2454d8674cc9ce7baa43f7048) back to 2.4-stable? It would be great if I can make further fix based on peer_info.timeout

realjiangms:
> @realjiangms I've made another fix for P2P connections in [3aec02d](https://github.com/ntop/n2n/commit/3aec02d3e64be01c6e3ec5caeed7e61df31cec46) . Please check out the new dev, all the p2p issues I had are now fixed

After quick glance I think it works. All I need to do is to add the reg.dstMac as I mentioned before in dev branch. This makes the tunnel available as soon as the REGISTER message. The latest commit just remove the stale entry in pending_peers, but registration would not happen until next p2p message

realjiangms:
Just a note. It seems you've already fixed dev branch in 52d33ed8 for dstmac

realjiangms:
Regarding this issue:
> In addition, I just found another problem. I used to try to build p2p connection between my home server (with PortRestrictedConeNat) and mobile (with mobile network, it seems Addr Rest Cone Nat). However it never works. I usually use -c mynetwork -k encryptme as readme. But I just tried it again without
> -k encryptme and it works! Not sure what's the difference. I captured network by wireshark on my labtop and don't find different in UDP data size. I'm trying to figure out why. (please don't ask me whether or not packets arrives my home router. I can't capture on it as it is provided by ISP and they don't give me admin privilege)

I figured out that it is not related to -k encryptme. It just has very little chance to get p2p succeed.
That's because we do not follow standard STUN behavior. REGISTER is expected not arrive peer, super node should forward the REGISTER and peer node would try ACK. That's a reliable way to PUNCH UDP HOLE. I found in sn.c we already have related code to handle this, but somehow it is not used, maybe because a minor logic error.
I raised PR #110 to fix it.

PR #110, opened 2019-05-23
Fix supernode handling REGISTER message from edge and let edge ask su…

realjiangms:
…pernode to forward REGISTER message in case of non-p2p-reflection.

This is reliable way to PUNCH UDP HOLE (following STUN spec)

Issue was originally mentioned in #109


emanuele-f:
Thank you!

Issue #27, opened 2018-07-16
Edge nodes public port for direct communication


emanuele-f:
In the current implementation, in order for two edge nodes to communicate directly without going through a supernode, an edge node (or both?) is required to have all the ports open. In a real world environment, this is impractical to implement because of:

1. Firewalls: we don't want do open all the ports
2. NAT

When a user is in control of his firewall and NAT, it's easy for him to just open/port map a single port for incoming communications. It would be great if n2n provided a way for a node to advertise such a *public port* to the other nodes, so that they can try to contact it directly through this port.

emanuele-f:
The same result can be achieved as follows with the current implementation:
- Setup a port forwarding rule for a UDP port (e.g. 2300)
- Use the `-p 2300` parameter in the edge node

Now the edge node will receive P2P packets

PR #115, opened 2019-05-27
More reliable punching

realjiangms:
1. Some nat router blocks the port if incoming traffic arrives before outcoming traffic being sent. Give edge ability to set proper TTL so that the registration packet is dropped before it arrives peer.
2. Support Symmetric NAT by predicting 15 more ports when sending registration packet
3. Purge pending mac also on P2P normal packet. This is actually more usual condition.

This fixes P2P failure on my environment:
1. Mobile network sometimes act as Symmetric NAT. 2. and 3. makes it P2P connectable
2. Router in my parents' home shows behavior of 1. and setting TTL =  number of nat layers + 1 make it P2P connectable.

Let me know if we need to split it into 2 PRs.
Contentious notes for discussion: 
1. Not sure whether setsock TTL works fine in win32 and I lack of environment to test win32. So I just discard it in win32.
2. We might need to introduce another command line argument to specify the number for predicting port. In this PR (as PoC) I hard code 16 and limited to use this tech only when TTL is specified (I'm afraid that lots of unknown incoming traffic on peer might trigger more router firewall policy). 
3. How to detect how much nat layers between local host and public network automatically?

### DIFF
--- a/doc/NEW_FEATURES.txt
+++ b/doc/NEW_FEATURES.txt
@@ -4,3 +4,18 @@ Between 2.0.x and 2.1.x
 * Better ming Windows build support.
 * Added -E flag to allow multicast ethernet traffic.
 
+* Added -L flag to allow set TTL for registration packet.
+This is an advanced flag to make sure that the registration packet is dropped immediately when it goes out of local nat
+so that it will not trigger some firewall behavior on target peer. Actually, the registration packet is only expected to
+make local nat UDP hole and is not expected to reach the target peer, see https://tools.ietf.org/html/rfc5389.
+To achieve this, the flag should be set as nat level + 1. For example, if we have 2 layer nat in local, we should set
+-L 3.
+Usually we know exactly how much nat layers in local.
+If we are not sure how much nat layers in local, we can use traceroute on Linux to check. Following example shows a local
+single layer nat because on second jump it shows a public ip address. In this case it should set -L 2.
+$ /usr/sbin/traceroute -w1 8.8.8.8
+traceroute to 8.8.8.8 (8.8.8.8), 30 hops max, 60 byte packets
+ 1  192.168.3.1 (192.168.3.1)  0.464 ms  0.587 ms  0.719 ms
+ 2  112.65.17.217 (112.65.17.217)  5.269 ms  7.031 ms  8.666 ms
+
+But this method is not always work due to various local network device policy.
--- a/edge.c
+++ b/edge.c
@@ -143,7 +143,7 @@ static void help() {
 #ifndef __APPLE__
 	 "[-D] "
 #endif
-	 "[-r] [-E] [-v] [-i <reg_interval>] [-t <mgmt port>] [-A] [-h]\n\n");
+	 "[-r] [-E] [-v] [-i <reg_interval>] [-L <reg_ttl>] [-t <mgmt port>] [-A] [-h]\n\n");
 
 #if defined(N2N_CAN_NAME_IFACE)
   printf("-d <tun device>          | tun device name\n");
@@ -155,6 +155,7 @@ static void help() {
   printf("-s <netmask>             | Edge interface netmask in dotted decimal notation (255.255.255.0).\n");
   printf("-l <supernode host:port> | Supernode IP:port\n");
   printf("-i <reg_interval>        | Registration interval, for NAT hole punching (default 20 seconds)\n");
+  printf("-L <reg_ttl>             | TTL for registration packet when UDP NAT hole punching through supernode (default 0 for not set )\n");
   printf("-p <local port>          | Fixed local UDP port.\n");
 #ifndef WIN32
   printf("-u <UID>                 | User ID (numeric) to use when privileges are dropped.\n");
@@ -303,6 +304,10 @@ static int setOption(int optkey, char *optargument, n2n_priv_config_t *ec, n2n_e
     conf->register_interval = atoi(optargument);
     break;
 
+  case 'L': /* supernode registration interval */
+    conf->register_ttl = atoi(optarg);
+    break;
+
 #if defined(N2N_CAN_NAME_IFACE)
   case 'd': /* TUNTAP name */
     {
@@ -393,7 +398,7 @@ static int loadFromCLI(int argc, char *argv[], n2n_edge_conf_t *conf, n2n_priv_c
   u_char c;
 
   while((c = getopt_long(argc, argv,
-			 "k:a:bc:Eu:g:m:M:s:d:l:p:fvhrt:i:SD"
+			 "k:a:bc:Eu:g:m:M:s:d:l:p:fvhrt:i:SDL:"
 #ifdef N2N_HAVE_AES
 			 "A"
 #endif
--- a/edge_utils.c
+++ b/edge_utils.c
@@ -384,9 +384,43 @@ static void register_with_new_peer(n2n_edge_t * eee,
 	       HASH_COUNT(eee->pending_peers));
 
     /* trace Sending REGISTER */
-    send_register(eee, &(scan->sock), mac);
     if(from_supernode) {
+	  /* UDP NAT hole punching through supernode. Send to peer first(punch local UDP hole)
+       * and then ask supernode to forward. Supernode then ask peer to ack. Some nat device
+       * drop and block ports with incoming UDP packet if out-come traffic does not exist.
+       * So we can alternatively set TTL so that the packet sent to peer never really reaches
+       * The register_ttl is basically nat level + 1. Set it to 1 means host like DMZ.
+       */
+      if (eee->conf.register_ttl == 1) {
+        /* We are DMZ host or port is directly accessible. Just let peer to send back the ack */
+#ifndef WIN32
+      } else if(eee->conf.register_ttl > 1) {
+        /* Setting register_ttl usually implies that the edge knows the internal net topology
+         * clearly, we can apply aggressive port prediction to support incoming Symmetric NAT
+         */
+        int curTTL = 0;
+        socklen_t lenTTL = sizeof(int);
+        n2n_sock_t sock = scan->sock;
+        int alter = 16; /* TODO: set by command line or more reliable prediction method */
+
+        getsockopt(eee->udp_sock, IPPROTO_IP, IP_TTL, (void *)(char *)&curTTL, &lenTTL);
+        setsockopt(eee->udp_sock, IPPROTO_IP, IP_TTL,
+                  (void *)(char *)&eee->conf.register_ttl,
+                  sizeof(eee->conf.register_ttl));
+        for (; alter > 0; alter--, sock.port++)
+        {
+          send_register(eee, &sock, mac);
+        }
+        setsockopt(eee->udp_sock, IPPROTO_IP, IP_TTL, (void *)(char *)&curTTL, sizeof(curTTL));
+#endif
+      } else { /* eee->conf.register_ttl <= 0 */
+        /* Normal STUN */
+        send_register(eee, &(scan->sock), mac);
+      }
       send_register(eee, &(eee->supernode), mac);
+    } else {
+      /* P2P register, send directly */
+      send_register(eee, &(scan->sock), mac);
     }
 
     register_with_local_peers(eee);
@@ -1460,6 +1494,16 @@ static void readFromIPSocket(n2n_edge_t * eee, int in_sock) {
 
 	  if(is_valid_peer_sock(&pkt.sock))
 	    orig_sender = &(pkt.sock);
+
+	  if(!from_supernode) {
+	    /* This is a P2P packet from the peer. We purge a pending
+	     * registration towards the possibly nat-ted peer address as we now have
+	     * a valid channel. We still use check_peer_registration_needed in
+	     * handle_PACKET to double check this.
+	     */
+	    traceEvent(TRACE_DEBUG, "Got P2P packet");
+	    find_and_remove_peer(&eee->pending_peers, pkt.srcMac);
+	  }
 
 	  traceEvent(TRACE_INFO, "Rx PACKET from %s (sender=%s) [%u B]",
 		     sock_to_cstr(sockbuf1, &sender),
--- a/n2n.h
+++ b/n2n.h
@@ -217,6 +217,7 @@ typedef struct n2n_edge_conf {
   uint8_t             tos;                    /** TOS for sent packets */
   char                *encrypt_key;
   int                 register_interval;      /**< Interval for supernode registration, also used for UDP NAT hole punching. */
+  int                 register_ttl;           /**< TTL for registration packet when UDP NAT hole punching through supernode. */
   int                 local_port;
   int                 mgmt_port;
 } n2n_edge_conf_t;


emanuele-f:
>     1. Some nat router blocks the port if incoming traffic arrives before outcoming traffic being sent. Give edge ability to set proper TTL so that the registration packet is dropped before it arrives peer.

Why using a TTL to drop the packet instead of avoid sending it at all? What do you mean with "block" here? Only block the incoming packet or also subsequent packets from the host behind nat?

>     2. Support Symmetric NAT by predicting 15 more ports when sending registration packet

Can you explain why this works?

>     3. Purge pending mac also on P2P normal packet. This is actually more usual condition.

Why is this needed?

How did you figured out the number of NAT layers in your network?

realjiangms:
> > ```
> > 1. Some nat router blocks the port if incoming traffic arrives before outcoming traffic being sent. Give edge ability to set proper TTL so that the registration packet is dropped before it arrives peer.
> > ```
> 
> Why using a TTL to drop the packet instead of avoid sending it at all? What do you mean with "block" here? Only block the incoming packet or also subsequent packets from the host behind nat?

We have to send the registration packet. The STUN trick is first send out UDP packet from local so that there's a record in nat. Then supernode notify peer to reply the UDP packet (but actually, it usually doesn't really receive the UDP packet from local because its nat will drop the incoming packet from where it never sends packet to previously). The replied UDP packet could reach local successfully because local nat think it's valid replied UDP packet. The P2P tunnel exists.
"Block" means all subsequent packets on specific port would be dropped. Maybe the firewall on the nat exactly want to block P2P communication (as part of flow control policy)
So my trick is to use TTL on first UDP packet from local, so it would only cause record on local nat device, but never really reach peer nat device (to avoid block policy).

> 
> > ```
> > 2. Support Symmetric NAT by predicting 15 more ports when sending registration packet
> > ```
> 
> Can you explain why this works?

Symmetric NAT allocate different port for same src local port to different dst. However most of them just increase or decrease port number for new allocation. For example, we use local port 7654, nat might allocate 63360 for supernode, but 63361 (or 63362, 63363) for peer edge node, depends on which one is free. A more smart solution is to have 2 supernode so that we can check the port allocation policy of the Symmetric NAT. I would leave it for further implementation as it might need change on n2n protocol. Currently as my experience most  most common Symmetric NAT is mobile network which always increase the port number for new connection.

> 
> > ```
> > 3. Purge pending mac also on P2P normal packet. This is actually more usual condition.
> > ```
> 
> Why is this needed?

We should not miss any possible to create P2P tunnel. Currently we purge pending mac on P2P registration. Actually it is not common as registration packet is not expected to reach peer unless peer is DMZ host exposed on public internet directly, or local and peer happened to send registration at almost same time (the faster one will receive the packet, and hope the nat on slower one doesn't have block policy as I mentioned above).
So I introduce pending mac purge on normal P2P packet. Because once there's P2P packet, a tunnel already exists and we should try to confirm it by P2P registration. Checking on each P2P packet might hit performance if there's a long chain on pending list for current implementation. But once we finish the hash data structure on peer_list, that would not be a problem. And anyway, we already have check_peer_registration_needed() on each P2P packet

> 
> How did you figured out the number of NAT layers in your network?

I leave it as command line argument. But recently I figure out a rough way:
/usr/sbin/traceroute -w1 <supernode>
Then in 1s it will give all route trace on local to supernode. The first TTL without local ip(10.x.x.x, 172.x.x.x, 192.168.x.x.x, 169.254.x.x.x, 100.x.x.x) should be set as argument for n2n edge

realjiangms:
> > > ```
> > > 1. Some nat router blocks the port if incoming traffic arrives before outcoming traffic being sent. Give edge ability to set proper TTL so that the registration packet is dropped before it arrives peer.
> > > ```
> > 
> > 
> > Why using a TTL to drop the packet instead of avoid sending it at all? What do you mean with "block" here? Only block the incoming packet or also subsequent packets from the host behind nat?
> > We have to send the registration packet. The STUN trick is first send out UDP packet from local so that there's a record in nat. Then supernode notify peer to reply the UDP packet (but actually, it usually doesn't really receive the UDP packet from local because its nat will drop the incoming packet from where it never sends packet to previously). The replied UDP packet could reach local successfully because local nat think it's valid replied UDP packet. The P2P tunnel exists.
> > "Block" means all subsequent packets on specific port would be dropped. Maybe the firewall on the nat exactly want to block P2P communication (as part of flow control policy)
> > So my trick is to use TTL on first UDP packet from local, so it would only cause record on local nat device, but never really reach peer nat device (to avoid block policy).
> > > ```
> > > 2. Support Symmetric NAT by predicting 15 more ports when sending registration packet
> > > ```
> > 
> > 
> > Can you explain why this works?
> > Symmetric NAT allocate different port for same src local port to different dst. However most of them just increase or decrease port number for new allocation. For example, we use local port 7654, nat might allocate 63360 for supernode, but 63361 (or 63362, 63363) for peer edge node, depends on which one is free. A more smart solution is to have 2 supernode so that we can check the port allocation policy of the Symmetric NAT. I would leave it for further implementation as it might need change on n2n protocol. Currently as my experience most  most common Symmetric NAT is mobile network which always increase the port number for new connection.
> > > ```
> > > 3. Purge pending mac also on P2P normal packet. This is actually more usual condition.
> > > ```
> > 
> > 
> > Why is this needed?
> > We should not miss any possible to create P2P tunnel. Currently we purge pending mac on P2P registration. Actually it is not common as registration packet is not expected to reach peer unless peer is DMZ host exposed on public internet directly, or local and peer happened to send registration at almost same time (the faster one will receive the packet, and hope the nat on slower one doesn't have block policy as I mentioned above).
> > So I introduce pending mac purge on normal P2P packet. Because once there's P2P packet, a tunnel already exists and we should try to confirm it by P2P registration
> > How did you figured out the number of NAT layers in your network?
> > I leave it as command line argument. But recently I figure out a rough way:
> > /usr/sbin/traceroute -w1 
> > Then in 1s it will give all route trace on local to supernode. The first TTL with non-local ip(10.x.x.x, 172.x.x.x, 192.168.x.x.x, 169.254.x.x.x, 100.x.x.x) should be set as argument for n2n edge

emanuele-f:
> "Block" means all subsequent packets on specific port would be dropped. Maybe the firewall on the nat exactly want to block P2P communication (as part of flow control policy)
> So my trick is to use TTL on first UDP packet from local, so it would only cause record on local nat device, but never really reach peer nat device (to avoid block policy).

Let's express the situation like this:

1. A is a peer, B is a peer behind NAT/firewall F, S is supernode
2. B registers with S, so F now has a `F:x -> S:z => B:l` mapping
3. A sends a P2P packet to B: `A:y -> F:x`. F blocks the packet as it is an asymmetric NAT and A:y does not correspond to S:z. This is normal with asymmetric nat.

Are you saying that other than the "block" in step 3, you also have subsequent `F:x -> S:z` packets dropped? Anyway, also your solution with TTL assumes that the problematic NAT is the last NAT on the B side. What if there are `F1 -> F2 -> F3 -> B` and F2 is the problematic NAT? In this case you can only set TTL=1 . Anyway, I think this is a very complex situation and the average user (or even me) does not know this stuff and the correct TTL to use.

> Symmetric NAT allocate different port for same src local port to different dst. However most of them just increase or decrease port number for new allocation. For example, we use local port 7654, nat might allocate 63360 for supernode, but 63361 (or 63362, 63363) for peer edge node, depends on which one is free.

I think this is too much a speculation and also produces a lot of traffic. Do you know some software which uses this speculation in production? Can't a more standard hole punching mechanism work in this case?

> A more smart solution is to have 2 supernode so that we can check the port allocation policy of the Symmetric NAT. I would leave it for further implementation as it might need change on n2n protocol.

This is even more complex!

>
> >    How did you figured out the number of NAT layers in your network?
>
> I leave it as command line argument. But recently I figure out a rough way:
> /usr/sbin/traceroute -w1 

Ok.

Edit: have you considered the option to add a UDP port forwarding rule in non-mobile environments as described in https://github.com/ntop/n2n/issues/27#issuecomment-489463311 ?

realjiangms:
> 
> Let's express the situation like this:
> 
> 1. A is a peer, B is a peer behind NAT/firewall F, S is supernode
> 2. B registers with S, so F now has a `F:x -> S:z => B:l` mapping
> 3. A sends a P2P packet to B: `A:y -> F:x`. F blocks the packet as it is an asymmetric NAT and A:y does not correspond to S:z. This is normal with asymmetric nat.
> 
> Are you saying that other than the "block" in step 3, you also have subsequent `F:x -> S:z` packets dropped? Anyway, also your solution with TTL assumes that the problematic NAT is the last NAT on the B side. What if there are `F1 -> F2 -> F3 -> B` and F2 is the problematic NAT? In this case you can only set TTL=1 . Anyway, I think this is a very complex situation and the average user (or even me) does not know this stuff and the correct TTL to use.

Let me answer other questions and describe this case in next comment.
> 
> > Symmetric NAT allocate different port for same src local port to different dst. However most of them just increase or decrease port number for new allocation. For example, we use local port 7654, nat might allocate 63360 for supernode, but 63361 (or 63362, 63363) for peer edge node, depends on which one is free.
> 
> I think this is too much a speculation and also produces a lot of traffic. Do you know some software which uses this speculation in production? Can't a more standard hole punching mechanism work in this case?

Actually this is a simplified solution. A more standard method is described in https://www.goto.info.waseda.ac.jp/~wei/file/wei-apan-v10.pdf. This is far more complex (below 'more smart solution'.
Regarding the traffic, the registration phase is not the main offload, so I think it is not a big concern. Actually I also see somewhere a more aggressive solution: try all of the 65535 ports!
> 
> > A more smart solution is to have 2 supernode so that we can check the port allocation policy of the Symmetric NAT. I would leave it for further implementation as it might need change on n2n protocol.
> 
> This is even more complex!
Yes
> 
> > > How did you figured out the number of NAT layers in your network?
> > 
> > 
> > I leave it as command line argument. But recently I figure out a rough way:
> > /usr/sbin/traceroute -w1
> 
> Ok.
> 
This can help to give a correct TTL option
> Edit: have you considered the option to add a UDP port forwarding rule in non-mobile environments as described in [#27 (comment)](https://github.com/ntop/n2n/issues/27#issuecomment-489463311) ?

UDP port forwarding or setting DMZ host is a good and reliable solution for P2P communication. However the reason we need P2P VPN is that usually this is not an option:
1. User might not know how to configure routers
2. User might not have full access-control to his router (it is usually a EPON modem provided by ISP and they never give you admin access as they might want to limit the possibility that you can expose service on internet)

realjiangms:
> > "Block" means all subsequent packets on specific port would be dropped. Maybe the firewall on the nat exactly want to block P2P communication (as part of flow control policy)
> > So my trick is to use TTL on first UDP packet from local, so it would only cause record on local nat device, but never really reach peer nat device (to avoid block policy).
> 
> Let's express the situation like this:
> 
> 1. A is a peer, B is a peer behind NAT/firewall F, S is supernode
> 2. B registers with S, so F now has a `F:x -> S:z => B:l` mapping
> 3. A sends a P2P packet to B: `A:y -> F:x`. F blocks the packet as it is an asymmetric NAT and A:y does not correspond to S:z. This is normal with asymmetric nat.
> 
> Are you saying that other than the "block" in step 3, you also have subsequent `F:x -> S:z` packets dropped? Anyway, also your solution with TTL assumes that the problematic NAT is the last NAT on the B side. What if there are `F1 -> F2 -> F3 -> B` and F2 is the problematic NAT? In this case you can only set TTL=1 . Anyway, I think this is a very complex situation and the average user (or even me) does not know this stuff and the correct TTL to use.
> 


I'm not assuming problematic NAT is latest NAT. TTL is assumed to be properly set on peer. The problematic NAT could be any one as setting TTL on peer would make the packet not reach any of the NAT on the problematic peer side.
Let me reuse your situation. We are talking about asymmetric NAT in this case, which means port mapping is same for all remote peers (and that's why usually STUN process works)
Here's how it works without problem:
1. A is a peer, B is a peer behind NAT/firewall F, S is supernode. Let's assume F **is not** problematic 
Let's assume A is also behind NAT Fa. And A also use port l for local.
2. B registers with S, so F now has a F:x -> S:z => B:l mapping
3. A sends a P2P register to B: A:l -> Fa:y -> F:x. Now Fa has Fa:y -> F:x => A:l
4. P2P register is dropped.
Note that it is asymmetric, so F:x -> Fa:y should also map to B:l. However since there's no previous B:l -> F:x -> Fa:y traffic exists, the packet is dropped (Actually, for a FULL CONE NAT it would not be dropped. But let's assume we have Port/Address Restricted Nat). 
But this is just as expected in standard STUN process. The trick is at A side. Now A has Fa:y -> F:x => A:l, and there's already A:l -> Fa:y -> F:x traffic exists. 
5. B send P2P register ACK to A (because supernode notified it to do so, although it doesn't receive the P2P register from A). It would send B:l -> F:x -> Fa:y. Since there's already A:l -> Fa:y -> F:x(Fa:y -> F:x => A:l) on A side, the packet would reach peer A as F:x -> Fa:y -> A:l. Then we have P2P tunnel done.

Now assume F **is** problematic.
1. A is a peer, B is a peer behind NAT/firewall F, S is supernode. Let's assume F **is** problematic 
Let's assume A is also behind NAT Fa. And A also use port l for local.
2. B registers with S, so F now has a F:x -> S:z => B:l mapping
3. A sends a P2P register to B: A:l -> Fa:y -> F:x. Now Fa still has Fa:y -> F:x => A:l.
4. P2P register is dropped. But since F *is* problematic, it will dropped all further income/outcome traffic on F:x.
5. B send P2P register ACK to A. It would send B:l -> F:x. Unfortunately, there will be no F:x -> Fa:y because F drops. Even more, if B try to send P2P register to A, F would also block. Each retry from A or B might refresh the block period and we never get P2P tunnel.

So my solution is add TTL on A side. Then in step #3 there will only be A:l -> Fa:y, no Fa:y -> F:x, then F will not be provoked. If there's 3 nat in front of A, the TTL could be 4 so that the first router on public internet would drop the P2P register packet. P2P register packet is expected to reach peer, so it is not necessary to have it reach peer's NAT.

emanuele-f:
@realjiangms thank you for the explanation. Can you check if f36814b152f3a8d6a47e94f3735df507e52147a4 solves the need to add the find_and_remove_peer ?

realjiangms:
I checked f36814b. I think we still need find_and_remove_peer on normal p2p packet. I used to be interested in the minor change in edge_utils.c which updates the scan->sock with latest peer (in case of normal p2p packet, it will be updated to correct peer sock). However it doesn't help as no one refers to scan->sock later in pending_peer list. To be easy to refer to the code I mentioned, I added a comment so that we can reference it: https://github.com/ntop/n2n/commit/f36814b152f3a8d6a47e94f3735df507e52147a4#r33868811

Since meyered fix are merged, I might consider to reconstruct my PR.

emanuele-f:
Please do the necessary checks when you have time, I've some concerns about these two lines:

- https://github.com/ntop/n2n/pull/115/files#diff-6dd127424612a99b5f9eb1d1afe7c59aR1426
- https://github.com/ntop/n2n/pull/115/files#diff-6dd127424612a99b5f9eb1d1afe7c59aR436

as exchanging more register packets may increase network traffic and also increase the chance of entering a registration loop where two edge nodes continue to try p2p while they could already p2p with the existing connections.

It's ok for me to integrate your additional -T parameter as I understand that it will give you more possibility to establish p2p communications, however for me this new punching method (as documented in the pdf above) is a bit insane as it would require 2 supernodes and 1000 packets to be echanged to just try p2p. You can increase the 16 value if it will increase the chances to do p2p. We also have to document such behavior in a new file under /doc, otherwise no one will know how to use it!

realjiangms:
I'm reconstructing the PR. Yes I'll create new material under /doc for more description. I might need your review / comments on the wording. 
I'm busy with my daytime work in the past 2 weeks so I don't have much big progress these days.
I'll try to make some time next several days and get it done ASAP

realjiangms:
@emanuele-f I've rebased with latest dev branch and tested in all the endpoint I have.
Please have a check

emanuele-f:
Thank you. I've moved the documentation to the man page.

Issue #101, opened 2019-04-30
Thoughts on Key Entropy


Logan007:
Let's assume that an input key has a certain "variety". For example, if we restricted ourselves to the use of 64 ASCII characters (at 6 bit each), an input key of 4 characters length would have a variety of 24.

As of now, the input key gets stretched/shrinked to 512 bits in the key hash (SHA512). Hashing the key does not change overall variety. Variety just gets distributed over the hash now. But there is a good [reason](https://github.com/ntop/n2n/issues/72#issuecomment-485420620) to hash it anyway.

So, following the example given above, every bit in the hash (from which the key is taken) would carry a  variety of 24 / 512 = 0.047. Our aim would be that every bit of the hash that gets used in the key carries a variety of at least 1 (to make sure the full key space gets used).

Thus, the first conclusion is that we would want to keep the hash as short as possible but as long as necessary to make best use of the input key's variety. But how long should the hash be?

The major security defining parameter for AES encryption is its key. In case of 128 bit keys, we need 128 bit. Furthermore, we use a different, additional 128 bit key for IV encryption which essentially contributes to security. This adds up to 256 bit. Luckily, there is a SHA256 that would deliver exactly a 256 bit value that could be split into the two keys needed. The extra bits needed for an implicit IV padding before encryption or use (see #72) could be generated by hashing the hash again; this is feasible in this case of a short input key. But we need to be aware that those extra bits do not come with extra variety.

Assuming that the input key carries a varitey of only 6 bit per byte, every input key with a length equal or below 256 / 6 = 43 should be hashed using SHA256.

For longer input keys, we would want to switch to 192 bit keys for longer input to make use of the full key space. SHA384 delivers 192 + 128 + 64 bit for the payload-key, the IV encryption key, and a possible IV padding, respectively. Following the presumed expectation that longer input keys yield to higher security, the additional 64 bit for IV padding are derived directly from the key and thus carrying their own variety. With 6 bit per byte, every input key of length between 44 and (384 / 6 =) 64 should be treated this way.

For even longer input keys, we finally would need to switch to SHA512 using the well know scheme of 256 bit payload key, 128 bit IV key, and 64 bit IV padding. We even have extra 64 bit! Following the above mentioned scheme, this would be the way to chose for input keys with length equal or above 65 characters. If a user explicitly wants to go for the _256 bit payload encryption_, an "_optimal" input key length_ would be around (512 / 6 =) _83 characters_.

[...]

PR #118, opened 2019-05-31
input key size dependant AES key -hashes and -sizes

Logan007:
This pull request applies the scheme of #101 to generate internal key material and key sizes from the input key.

As a by-product, `TRANSOP_AES_IV_SEED_SIZE` now can range between 0 and 16 allowing the compiler savvy user to determine the trade-off between highest security (=16 byte random IV seed, longer packet) and (=0 byte random IV seed to be transmitted, shorter packet). Default is the well discussed and balanced value of 8.

These changes break compatibility to current dev version as key material now is generated in a different way than before, all edges would need to be updated.

### DIFF
--- a/transform_aes.c
+++ b/transform_aes.c
@@ -25,7 +25,7 @@
 #include "openssl/sha.h"
 
 #define N2N_AES_TRANSFORM_VERSION       1  /* version of the transform encoding */
-#define N2N_AES_IVEC_SIZE               32 /* Enough space for biggest AES ivec */
+#define N2N_AES_IVEC_SIZE               (AES_BLOCK_SIZE)
 
 #define AES256_KEY_BYTES (256/8)
 #define AES192_KEY_BYTES (192/8)
@@ -34,11 +34,15 @@
 /* AES plaintext preamble */
 #define TRANSOP_AES_VER_SIZE     1       /* Support minor variants in encoding in one module. */
 #define TRANSOP_AES_SA_SIZE      4
-#define TRANSOP_AES_IV_SEED_SIZE 8
+#define TRANSOP_AES_IV_SEED_SIZE 8	/* size of transmitted random part of IV in bytes, between 0 .. 16 */
+#define TRANSOP_AES_IV_PADDING_SIZE (N2N_AES_IVEC_SIZE - TRANSOP_AES_IV_SEED_SIZE)
+#define TRANSOP_AES_IV_KEY_BYTES (AES128_KEY_BYTES) /* use AES128 for IV encryption */
 #define TRANSOP_AES_PREAMBLE_SIZE (TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + TRANSOP_AES_IV_SEED_SIZE)
 
 /* AES ciphertext preamble */
 #define TRANSOP_AES_NONCE_SIZE   4
+
+const uint8_t RAND_BYTES = log2(RAND_MAX) / 8; /* number of random bytes returned from rand() function */
 
 typedef unsigned char n2n_aes_ivec_t[N2N_AES_IVEC_SIZE];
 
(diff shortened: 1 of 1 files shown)


transform_aes.c
```diff
@@ -34,27 +34,25 @@
...
-#define TRANSOP_AES_IV_SEED_SIZE 8
+#define TRANSOP_AES_IV_SEED_SIZE 8	/* size of transmitted random part of IV in bytes, between 0 .. 16 */
+#define TRANSOP_AES_IV_PADDING_SIZE (N2N_AES_IVEC_SIZE - TRANSOP_AES_IV_SEED_SIZE)
+#define TRANSOP_AES_IV_KEY_BYTES (AES128_KEY_BYTES) /* use AES128 for IV encryption */
 #define TRANSOP_AES_PREAMBLE_SIZE (TRANSOP_AES_VER_SIZE + TRANSOP_AES_SA_SIZE + TRANSOP_AES_IV_SEED_SIZE)
 
 /* AES ciphertext preamble */
 #define TRANSOP_AES_NONCE_SIZE   4
 
+const uint8_t RAND_BYTES = log2(RAND_MAX) / 8; /* number of random bytes returned from rand() function */
```
emanuele-f: log2 is probably not standard to be used at the pre-compilation stage

transform_aes.c
```diff
@@ -259,37 +245,80 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
...
-    /* We still use aes_best_keysize (even not necessary since we hash the key
-     * into the 256bits enc_dec_key) to let the users choose the degree of encryption.
-     * Long keys will pick AES192 or AES256 with more robust but expensive encryption.
+    /* create a working buffer of maximal occuring hashes size and generate
+     * the hashes for the aes key material, key_mat_buf_lengh indicates the
+     * actual "filling level" of the buffer
      */
-    aes_keysize_bytes = aes_best_keysize(key_size);
-    aes_keysize_bits = 8 * aes_keysize_bytes;
+    key_mat_buf = calloc(1, SHA512_DIGEST_LENGTH + SHA256_DIGEST_LENGTH);
```
emanuele-f: A normal stack allocation with `uint8_t key_mat_buf[SHA512_DIGEST_LENGTH + SHA256_DIGEST_LENGTH]` should be enough

transform_aes.c
```diff
@@ -141,11 +118,16 @@ static int transop_encode_aes( n2n_trans_op_t * arg,
...
+             * and sometimes < 32bit (e.g. Windows).
              */
-            iv_seed = ((((uint64_t)rand() & 0xFFFFFFFF)) << 32) | rand();
-            encode_buf(outbuf, &idx, &iv_seed, sizeof(iv_seed));
+	    uint32_t random_int;
+	    for (int i=0; i < TRANSOP_AES_IV_SEED_SIZE; i++) {
+		if ((i % RAND_BYTES) == 0) random_int = rand();
+		iv_seed[i] = (random_int >> ((i % RAND_BYTES) * 8)) & 0xFF;
+            }
+            encode_buf(outbuf, &idx, iv_seed, TRANSOP_AES_IV_SEED_SIZE);
```
emanuele-f: This is not part of this pull request topic, we will discuss this separately, please restore the old code for now

transform_aes.c
```diff
@@ -208,12 +192,14 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
             decode_uint32( &sa_rx, inbuf, &rem, &idx );
 
             /* Get the IV seed */
-            decode_buf((uint8_t *)&iv_seed, sizeof(iv_seed), inbuf, &rem, &idx);
+	    decode_buf(iv_seed, TRANSOP_AES_IV_SEED_SIZE, inbuf, &rem, &idx);
 
-            traceEvent( TRACE_DEBUG, "decode_aes %lu with seed %016lx", in_len, iv_seed );
+	    char * iv_seed_output[TRANSOP_AES_IV_SEED_SIZE*2];
+	    for (int i=0; i < TRANSOP_AES_IV_SEED_SIZE; i++) sprintf(iv_seed_output, "%s%02x", iv_seed_output,iv_seed[i]);
```
emanuele-f: Too many calls to sprintf here. Anyway, please remove and keep the old iv seed code for now

emanuele-f (changes requested):

Logan007:
Thank you for your feedback. Will go through it, maybe somewhen during the weekend and "update" the pull request.

transform_aes.c
Logan007: You are right, that would be much easier... and doesn't create the need to take care of de-allocation.

transform_aes.c
Logan007: D'accord. We should discuss random numbers and their generation separately.

transform_aes.c
Logan007: Will do... I just hope I can manage git to accept my local changes to this pull request. I guess my remote HEAD changed meanwhile.... Let's see.

transform_aes.c
Logan007: I am not sure about that. Anyway, this point shall be subsumed under the random number discussion.

Logan007:
I have restored `iv_seed` back to `uint64_t` to get back to the former invocation of rand(). Thus, `TRANSOP_AES_IV_SEED_SIZE` can not be handled in a flexible way, it has to stick to `8`. 

Maybe that would be worth a new issue discussion...

Logan007:
The "Resolve Conflicts" button just produces a never-ending-wheeling copy cat... any hint on how to resolve the indicated conflict (which should already be resolved in code)?

Logan007:
Mmmmh , still conflicting... "Resolve Conflicts" not working... I am lost.

_Will try to make a new pull request._

Issue #120, opened 2019-05-31
AES: fields in packet for 2.6 (SA?)


Logan007:
While removing the nonce for #119, I stumbled on the SA field in the packet: Do we still need it?

If it comes to a stable 2.6 in the near future that introduces AES, do we need to think about how to make it future proof?

Do we accept that we probably already might break AES mode compatibility with release 2.8 again? If **yes**, we now could also get rid of SA field as potential upcoming features (DH key generation, MAC, ...) we probably will require some packet changes anyway. If **no**, we should think of the packet design more thoroughly.

emanuele-f:
We don't need it and compatibility of aes is not a problem. It can be removed to save space.

Issue #116, opened 2019-05-30
Remove nonce


Logan007:
While working on the changes required for #101, I found that the nonce still seems to be in the code. With a view to the current IV implementation, it probably became obsolete allowing for a 4 bytes savings in the packet.

emanuele-f:
I agree that's useless right now and can be removed in aes (in twofish it should remain to keep compatibility, see https://github.com/ntop/n2n/milestone/1).

Logan007:
~~Would you mind if I solved it together with #101?~~ I better put it into another pull request as soon as the other one is merged.

Logan007:
~~#119~~

Issue #83, opened 2019-03-24
AES psk setup: possible security issue in dev branch


Logan007:
While working on the implementation of aes v2 and especially compatibility towards v1, I started thinking over line 418 in transform_aes.c:

```
memcpy(keybuf, padded_keybuf, pstat);
```
This might cause a huge security issue, allowing everybody to decrypt the packages using the zero-key in one of the three keysizes (128,192,256 - seem to deliver different results) or just any key added via setup_aes_key.

I suggest to apply the attached [patch](https://github.com/ntop/n2n/files/3000787/fixSetupAesKey.txt) to dev branch builds.

emanuele-f:
Fixed in c71f629d1b9f34cf0b80556791e563b167bfa0ba. This was a severe security bug, the key (used both for aes and twofish) was zeroed, making it useless. This was introduced in 37233553a451df6caa3895c6197c1f8cf779cc11 . Thank you for reporting!

Issue #72, opened 2019-03-01
Properly initialize the AES IV


emanuele-f:
As explained in #70, the IV used in AES encryption is currently zeroed. This affects the security and must be fixed.

Logan007:
There is a solution as proposed in the patch provided for [#70](https://github.com/ntop/n2n/issues/70) and also in #76.

emanuele-f:
To fix this issue in the current n2n AES implementation, I would replace the 32bit nonce number of AES (which is set but never checked anyway) with a random 64bit clear-text number. The remaining 64 bits of the IV can be set to 0 for now. I think this simple fix can improve security as it will prevent two exact plaintexts to produce the same ciphertexts with high probability (clash will only occur with 1/2^64 probability). This should be effective as the IV is not required to be kept secret (this is what I read about it online). Do you think this could be a good compromise for the current AES version?

Logan007:
Basically, you are right that usually the IV could be transmitted unencryptedly. The underlying assumption is that an attacker does not know anything about the plaintext then.

However, in this case, there is a problem. The data transmitted is an ethernet frame. An ethernet frame always starts with a constant preamble and a start frame delimeter, so the first 8 bytes are always the same. In every packet. These eight bytes are followed by destination and source MAC at 6 bytes each. The MACs are well known as they also are transmitted outside the encrypted portion, in the transformation-independent envelope of n2n packets. Thus, the plaintext of the first data block is extremly predictable if not "well known" to any attacker.

Let us assume that we unencryptedly transmit the IV. In cipher block chaining, the IV gets XORed to the first block of ethernet plaintext before it gets encrypted. Openly transmitting the IV, an attacker could find out the first plaintext data that went into encryption by just XORing the IV to the assumed ethernet plaintext.

One may think that this no big deal as this data is known anyway. But knowing a plaintext that went into encryption and the cipher that comes out of it, allows for _known plaintext attacks_ which helps an attacker to find out the key. And if we randomly deliver more variants of the plaintext (with every new IV) together with its cipher, known plaintext attack gets easier with every packet sent.

That would not happen, if IV were encrypted for transmission.

[...]

PR #128, opened 2019-06-10
optimization to flexible AES keysize depending on input key size

Logan007:
This pull request brings the thoughts of #101 to the code. Basically, it is to make full use of the key space and thus enhancing security.

As the internal key material now gets generated from the input key in a different way than before, this breaks compatibility; all edge nodes in the network would be required to update. This only applies if relying on AES.

### DIFF
--- a/transform_aes.c
+++ b/transform_aes.c
@@ -60,7 +60,7 @@ static int transop_deinit_aes(n2n_trans_op_t *arg) {
     return 0;
 }
 
-static void set_aes_cbc_iv(transop_aes_t *priv, n2n_aes_ivec_t ivec, uint8_t *iv_seed) {
+static void set_aes_cbc_iv(transop_aes_t *priv, n2n_aes_ivec_t ivec, uint64_t iv_seed) {
     uint8_t iv_full[N2N_AES_IVEC_SIZE];
 
     /* Extend the seed to full block size with padding value */
@@ -237,6 +237,9 @@ static int setup_aes_key(transop_aes_t *priv, const uint8_t *key, ssize_t key_si
     size_t aes_key_size_bytes;
     size_t aes_key_size_bits;
 
+    uint8_t key_mat_buf[SHA512_DIGEST_LENGTH + SHA256_DIGEST_LENGTH];
+    size_t key_mat_buf_length;
+
     /* Clear out any old possibly longer key matter. */
     memset( &(priv->enc_key), 0, sizeof(priv->enc_key) );
     memset( &(priv->dec_key), 0, sizeof(priv->dec_key) );
@@ -257,8 +260,6 @@ static int setup_aes_key(transop_aes_t *priv, const uint8_t *key, ssize_t key_si
      * the hashes for the aes key material, key_mat_buf_lengh indicates the
      * actual "filling level" of the buffer
      */
-    uint8_t key_mat_buf[SHA512_DIGEST_LENGTH + SHA256_DIGEST_LENGTH];
-    size_t key_mat_buf_length;
 
     if (key_size >= 65)
     {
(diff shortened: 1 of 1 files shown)


lucktu:
Careful operation. Suggest to put aside temporarily.

Logan007:
@lucktu: I am not sure whether I understand that correctly as a call for reluctance. 

Do you refer to breaking compatibility - of which we already have quite a good history (hashing the key #68, changing IV scheme #72, key treatment #83) for AES being a dev-only feature so far?

Could you please explain a little bit more in detail?

lucktu:
Yes, worry about incompatibility with previous v2, unless used for v3/v4 ... ...

Logan007:
It is not up to me to decide. But with a view to recent compatibility breaking changes in dev branch, I wouldn't see any reason not to apply these changes anytime soon.

transform_aes.c
```diff
@@ -64,33 +60,12 @@ static int transop_deinit_aes(n2n_trans_op_t *arg) {
...
-    }
-    else
-    {
-        return AES128_KEY_BYTES;
-    }
-}
-
-static void set_aes_cbc_iv(transop_aes_t *priv, n2n_aes_ivec_t ivec, uint64_t iv_seed) {
-    uint8_t iv_full[AES_BLOCK_SIZE];
+static void set_aes_cbc_iv(transop_aes_t *priv, n2n_aes_ivec_t ivec, uint8_t *iv_seed) {
```
emanuele-f: iv_seed should be uint64_t, I think this is a leftover from the previous changes

transform_aes.c
```diff
@@ -235,60 +210,99 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
...
-    aes_keysize_bytes = aes_best_keysize(key_size);
-    aes_keysize_bits = 8 * aes_keysize_bytes;
 
-    /* Hash the main key to generate subkeys */
-    SHA512(key, key_size, (u_char*)&keybuf);
+    /* create a working buffer of maximal occuring hashes size and generate
+     * the hashes for the aes key material, key_mat_buf_lengh indicates the
+     * actual "filling level" of the buffer
+     */
+    uint8_t key_mat_buf[SHA512_DIGEST_LENGTH + SHA256_DIGEST_LENGTH];
```
emanuele-f: Move declarations at the start of the function otherwise some compilers will argue

emanuele-f (changes requested):

Logan007:
The changes are no problem. I will take care of those next week - I am on leave :palm_tree:

Logan007:
Sorry for closing... I am just not pro on git, still fat-fingering.

transform_aes.c
Logan007: Applied, your assumption is correct. However, it limits n2n to 64 bit random ivec... I still am in favour of a flex-sized random part of iv (by just changing TRANSOP_AES_IV_SEED_SIZE in the code according to needs). I might open another issue specifically referring to that.

transform_aes.c
Logan007: Okay, applied.

Logan007:
I would be glad to also implement #116 and #120 in `transform_aes.c`. But I guess, that would require another pull request.  git-question: Is there a more elegant way than to wait until this pull request here (#128) will have been merged?

emanuele-f:
Thank you. Merged in bc260c23128f9a91cd46e444511ae9358b25bc71 . I had to undo some whitespaces changes, please try to make a cleaner PR next time. Regarding the other PR question, when working on distinct non-overlapping features, the proper way to do this in git is to create a new branch and create a new PR based on that. The pending PR will still be bound to its branch whereas you can develop the new changes on the new independent branch, which should be based on the n2n upstream branch.

Logan007:
P.S.  Thank you for your remarks and explainations! I sure will try to work cleaner, of course.  In this case, however, I think some white space changes came in when trying to adjust to meanwhile occured changes (while still working on the former original file) using github's feature to resolve confilcts. I will definetly have to learn how to better handle git...

PR #139, opened 2019-06-16
non-Windows trace events no longer print the full path of the code file.

fengdaolong:
1. non-Windows trace events no longer print the full path of the code file.
2. fix the problem that the program is forced to exit by pressing the "Ctrl + C" key on the Windows system.

### DIFF
--- a/edge.c
+++ b/edge.c
@@ -551,7 +551,6 @@ static void daemonize() {
 
 static int keep_on_running;
 
-#ifdef __linux__
 
 static void term_handler(int sig) {
   static int called = 0;
@@ -566,7 +565,6 @@ static void term_handler(int sig) {
 
   keep_on_running = 0;
 }
-#endif
 
 /* *************************************************** */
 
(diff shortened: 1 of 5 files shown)


fengdaolong:
For the second fix above, using MingW-W64 GCC-8.1.0 under Windows 10 can be compiled normally.

emanuele-f:
Does it compile fine with visual studio compiler too?

fengdaolong:
I don't have a visual studio environment, so I haven't tried it. 
If visual studio is not compatible with linux's “signal.h” , it may not compile.
Unless you change the program, you will exit the program under Windows system and the program will not perform normal cleanup actions.

emanuele-f:
Cleanups are mostly useful to run `valgrind` properly and detect memory leaks, the other things can be shut down brutally since no further cleanup is necessary. Anyway, I'll check this on vs compiler

fengdaolong:
In the windows system, the hidden problem is memory leaks. The obvious problem is that P2P Packet stats are not displayed when exiting.

emanuele-f:
This is now fixed in a70641897da90ce81f404b9882de09df506c1d11

Issue #165, opened 2019-07-08
Make edge automatically read edge.conf if no option is given


jakeecn:
The edge.conf file cannot be read, edge for windows？

jakeecn:
When running edge.exe, it is determined that if there is edge.conf in the program directory, it will be read automatically, and the version information will be displayed if there is no edge.conf in the program directory.

emanuele-f:
Please try "edge.exe edge.conf" after fix in 9cc2aee7fc232b4b393be8f0c91806da4fd61000

jakeecn:
Thank you! 
1、Can you enable the running edge.exe to automatically read the edge.conf by default？
2、Whether registration as a windows service can be implemented？

lucktu:
Yes, it can read edge.conf.
However, the version number will not change? 2.5.0

jakeecn:
D:\software\Remote\n2n>edge.exe
Welcome to n2n v.2.5.1 for Windows
Built on Jul  8 2019 23:18:29
Copyright 2007-19 - ntop.org and contributors

edge <config file> (see edge.conf)
or
edge -d <tun device> -a [static:|dhcp:]<tun IP address> -c <community> [-k <encr
ypt key>]
    [-s <netmask>] [-m <MAC address>] -l <supernode host:port>
    [-p <local port>] [-M <mtu>] [-r] [-E] [-v] [-i <reg_interval>] [-t <mgmt po
rt>] [-b] [-A] [-h]

-d <tun device>          | tun device name
-a <mode:address>        | Set interface address. For DHCP use '-r -a dhcp:0.0.0
.0'
-c <community>           | n2n community name the edge belongs to.
-k <encrypt key>         | Encryption key (ASCII) - also N2N_KEY=<encrypt key>.
-s <netmask>             | Edge interface netmask in dotted decimal notation (25
5.255.255.0).
-l <supernode host:port> | Supernode IP:port
-i <reg_interval>        | Registration interval, for NAT hole punching (default
 20 seconds)
-b                       | Periodically resolve supernode IP
                         | (when supernodes are running on dynamic IPs)
-p <local port>          | Fixed local UDP port.
-m <MAC address>         | Fix MAC address for the TAP interface (otherwise it m
ay be random)
                         | eg. -m 01:02:03:04:05:06
-M <mtu>                 | Specify n2n MTU of edge interface (default 1400).
-r                       | Enable packet forwarding through n2n community.
[...]

PR #166, opened 2019-07-09
enable the running edge or supernode to automatically read the conf by default

fengdaolong:
The default configuration file (if it exists) can be automatically read in the current directory when the program is launched.
[https://github.com/ntop/n2n/issues/165#issuecomment-509443405](url)


emanuele-f:
Implemented in 0b965c37c0e2b98f2b14592bdfc6e7f68bf8ac4b . For the tracking events issue please open a new issue/pr with output before and after the fix

Issue #117, opened 2019-05-30
Random Numbers


Logan007:
As of now, random numbers get generated for use in nonce or IV seed using rand() .

For one of the future releases, we should discuss whether it is worth considering another, cryptographically more secure random number generator.

emanuele-f:
The nonce will be removed so let's just ignore it. Regarding the IV seed, do you think there is an issue using rand here? The openssl random generator is very expensive as we saw in the benchmarks

Logan007:
The way rand() generates pseudo random numbers is not considered cryptographically strong. It's max period is "only" 2 ^ 32 (or probably: 2 ^ 31 as unsigned). I wouldn't be surprised if there somewhere were a 32 bit x 4G = 16G file containing all rand() values in their sequence of occurence. Even though rand() gets seeded with a timer value (eee->start_time), this just changes the starting point of the sequence.

That makes the random numbers, i.e. our IV seed, somewhat predictable (their transmission is not encrypted). There is an attack described in the second part of [this](https://en.wikipedia.org/wiki/Initialization_vector#Properties) which highlights the need for true randomness. Although, our AES-ECB 128 bit encryption step before use complicates things for an attacker: the described XOR will not work, but still a start. A lot of assumptions and thus maybe just a theoretical problem.

For now, it just feels wrong to use such a predictable sequence where random numbers are expected. This is extremely unusual for cryptographic applications.

Alternatives could be openssl (indeed costly with a view to performance), some AES based PRNG method (performance to be evaluated), hash based PRNG (probably more costly than the AES based), or a Mersenne Twister (much longer period than rand() and thus less predictable, but without modifications also not considered cryptographically secure).

If we decided to look into alternatives, I would investigate some AES based solution as it is hardware accelerated on some platforms.

emanuele-f:
A more robust generator would be better, but the performance impact of the openssl random_bytes is too much for the use in IV seed

Logan007:
I would suggest to use AES (1 block, CTR mode, 128 bit key) with some key derived from the PSK, seeded **once** from XORing somewhat entropy-considered data, i.e. the seconds-fraction of current time, processor time, as well as full (unix) time stamp, and finally RDRAND (if available) or openssl's random_bytes.

Logan007:
[...]

PR #170, opened 2019-07-29
new random number generator (xorshift128+)

Logan007:
To better meet higher security standards, the use of C's linear congruent `rand()` function is replaced by the `xorshift128+` generator as random number generator. The `xorshift128+` generator has a period of 2^128 _(as opposed to 2^31)_, is way less linear, and passes major randomness testing _(BigCrush)_. Furthermore, its performance is on par with `rand()`.

Also, the seeding changed to more timer- as well as cpu cylce-depending values.

Even though this already is fully functional _(and thus provided as pull request)_, it is just the first step towards better random numbers – #117 is halfway done. Further steps to be taken include an additional generator as well as even more sophisticated seeding.

### DIFF
--- a/random_numbers.c
+++ b/random_numbers.c
@@ -1,20 +1,9 @@
-
 /* The following code offers an alterate pseudo random number generator
    namely XORSHIFT128+ to use instead of C's rand(). Its performance is 
    on par with C's rand().
  */
 
-
-
 #include <stdint.h>
-#ifdef __GNUC__
-#include <sys/time.h>
-#endif
-#include <time.h>
-#ifdef __linux__
-    #include <sys/syscall.h>
-    #define GRND_NONBLOCK	1
-#endif
 #include "n2n.h"
 #include "random_numbers.h"
 
@@ -22,7 +11,7 @@ struct rn_generator_state_t {
     uint64_t a, b;
  };
 
-/* The state must be seeded so that it is not all zero, choose some
+/* The state must be seeded in a way that it is not all zero, choose some
    arbitrary defaults (in this case: taken from splitmix64)
  */
 static struct rn_generator_state_t rn_current_state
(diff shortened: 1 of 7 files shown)


random_numbers.c
```diff
@@ -0,0 +1,154 @@
...
+    tm.tm_sec = wtm.wSecond;
+    tm.tm_isdst = -1;
+    clock = mktime(&tm);
+    tp->tv_sec = clock;
+    tp->tv_usec = wtm.wMilliseconds * 1000;
+    return (0);
+}
+#endif
+
+static int32_t set_seed (uint64_t seed) {
```
emanuele-f: This function mixes previous state, time, and user input together (depending on the user input) and this is not good. Why can't you simply use the inizialization described in https://en.wikipedia.org/wiki/Xorshift#Initialization ?

random_numbers.c
```diff
@@ -0,0 +1,154 @@
...
+    reseed_counter = (random_number_64() & 0x3ffff) + 1;
+    return (0);
+}
+
+/* The following code of xorshift128p was taken from
+   https://en.wikipedia.org/wiki/Xorshift as of July, 2019
+   and thus is considered public domain.
+ */
+static uint64_t xorshift128p_64 () {
+    if (reseed_counter > 0) {
```
emanuele-f: It seems like the reseed is not part of xorshift+ . I would remove it.

random_numbers.c
```diff
@@ -0,0 +1,154 @@
...
+    /* downcast to uint32_t */
+    return (uint32_t)(t + s);
+}
+
+/* ---  the following code is public/exported  ---
+	internal functions are wrapped for the
+	sake of future further ramification of
+	internal functions */
+
+int32_t random_number_seed (uint64_t seed) {
```
emanuele-f: This is just overhead. Please remove these functions and directly expose the ones above. Better to rename this to n2n_rand and set_seed to n2n_srand

random_numbers.c
```diff
@@ -0,0 +1,154 @@
...
+    uint64_t const s = rn_current_state.b;
+    rn_current_state.a = s;
+    t ^= t << 23;
+    t ^= t >> 17;
+    t ^= s ^ (s >> 26);
+    rn_current_state.b = t;
+    return t + s;
+}
+
+static uint32_t xorshift128p_32 () {
```
emanuele-f: This is just a copy of xorshift128p_64 with a cast, please remove

emanuele-f (changes requested):

random_numbers.c
Logan007: Generally, it would be no problem to use splitmix.

However, one problem remains: xorshiftplus still is a _deterministic_ pseudo number generator, even if it comes with a much longer period. So, same seeds != 0 (called twice to enter 128 bit seed) result in the same series of random numbers.

That's why I implemented the use of some entropy sources in case of seed == 0 – think of it as "high entropy mode". This could also be achieved by feeding that entropy into the regular seed (splitmix or not) which on the other hand would cause more code in edge_utils. I thought it would be more elegant to "hide" all this from the random number generator user.

As I am not able to see how splitmix takes care of stabilization in case of weak seeds (only few bits set), I definitely recommend a few "dry runs" of the random number generator after seeding.

Further action: If there are no objections, I will implement splitmix in random.c for seeding and move the entropy source feeding into edge_utils.

Unfortunately, if we remove the high-entropy seed sources from random.c,  the random number generator will not be able to reseed itself with high-entropy data.

random_numbers.c
Logan007: Reseeding is a well know practice to handle deterministic random number generators. However, it requires entropy in the reseeding process - otherwise, it would still be deterministic.

I can surely remove it, but we would give up some easily won additional security.

random_numbers.c
Logan007: Okay. I just hope the compiler will not complain.

random_numbers.c
Logan007: Okay. I will save this for later when branches to different generators (AES-CTR, OPENSSH) according to some "security_level" would be required.

Logan007:
Thank you for you feedback. I will try to implement the changes somewhen during the upcoming weekend.

random_numbers.c
emanuele-f: We know that rand() is deterministic and returns the same result given a seed. We don't have many expectations on rand() because we know that it is a naive random generator. By mixing different features into a custom random generation I think that we won't know the exact features of such random generator and we might think that it "magical" working great in all the situations and platforms, but this is unlikely to be true since it is a custom generator implemented only in this project. I would just take the standard xorshift+ if it provides substantial advantages over the current rand(), but avoid customized implementations.

random_numbers.c
emanuele-f: The problem is that this forces us to choose a reseed interval and I would not know which is the optimal reseed interval to use and the following comment is not helping me:

```
/* set the reseed_counter to a value between 1 and 2^18
   which roughly is the number of packets per two minutes
   on a 1 Gbit line used to full capacity */
```

Let's stick with standard implementation.

random_numbers.c
emanuele-f: Having two functions like the above is only useful if they actually provide different functionalities. For example, I would have expected that the xorshift128p_64 would use a different algorithm to produce a longer result than xorshift128p_32 . But this is not the case here, so this is only confusing. Type casting should be done by the callers if needed

emanuele-f:
Thank you for your efforts. Please check out the new comments above. The short answer is that we should try to be as standard as possible and avoid custom implementations because we don't have time to test them properly and little possibility for peer review on them. Please also be sure that the xorshift+ cannot go into a "looping" state (e.g. it always returns the same value).

random_numbers.c
Logan007: Done - I left the stabilization loop though.

random_numbers.c
Logan007: Done.

random_numbers.c
Logan007: Done.

random_numbers.c
Logan007: Done - compiler does not complain and seems to take care of casting inside caller's context.

Logan007:
The requested changes were applied.

For not littering `edge_utils.c` 's code, I left out the µsec-fraction part for the seed...

random_numbers.c
```diff
@@ -0,0 +1,55 @@
...
+ */
+static struct rn_generator_state_t rn_current_state
+			       = { .a    = 0x9E3779B97F4A7C15,
+			           .b    = 0xBF58476D1CE4E5B9
+};
+
+int32_t n2n_srand (uint64_t seed) {
+    /* apply splitmix64 algorithm (found on same wiki page as xorshift)
+       twice on the SEED to calcualte A and B of XORSHIFT128PLUS' state */
+    rn_current_state.a = seed;
```
emanuele-f: This seems different from the wikipedia implementation. This code applies:

```
rn_current_state.a = splitmix(seed)
rn_current_state.b = splitmix(seed)
```

whereas the wikipedia code runs two rounds of splitmix :
```
state = {seed}
rn_current_state.a = splitmix(&state)
rn_current_state.b = splitmix(&state)
```
Moreover in `rn_current_state.a = seed` the contant `0x9E3779B97F4A7C15` is not applied. The wikipedia implementation can be also found in other online source written in go:

https://github.com/vpxyz/xorshift/blob/master/xorshift128plus/xorshift128plus.go#L27
https://github.com/vpxyz/xorshift/blob/master/internal/internal.go#L30

emanuele-f (changes requested):

random_numbers.c
Logan007: The addition only gets applied to the state which is _not_ used further in that specific call again - but in the next. For every seeding according to the given pattern, splitmix algorithm is  applied twice in a row only - with initial state of {0}. 

That way, implementation is more linear, i.e. contains less sub routine calls, and there is no need to keep track of a static splitmix state. Isn't the result the same? What did I miss?

If you are not convinced, let me know, I definitely can change it.

Logan007: In a test run, I implemented the wiki-version which delivers the **very same results** as the streamlined implementation above. I personally think that the wiki-implementation kind of clutters the code.

I will upload the wiki-solution in a new commit. Please feel free to choose and incorporate any of the two commits/versions into the dev branch.

Logan007: Done.

Logan007:
The requested changes were applied, please see the [comment](https://github.com/ntop/n2n/pull/170#discussion_r315121198).

Logan007:
It seems that file conflicts have arisen meanwhile. Besides my own recommendation, is there any interest in having this cleaned up compatible to current dev? Please let me know.

I still am concerned about relatively low periodicity and predictable initialization of C's `rand()`. You could try yourself to start two edges on CLI using `-vvvvv` – even remotely – in the very same second by pressing enter at the same time. You will probably see the very same IVs being used on both sides following the trace output.

Logan007:
Closed for preparing a new clean one.

Issue #91, opened 2019-04-14
Use lzo or remove it


emanuele-f:
minilzo is currently compiled but never used

Logan007:
I would love the idea to use some kind of compression.

Logan007:
This [example](https://fossies.org/linux/lzo/minilzo/testmini.c) gives well documented hints on how to use minilzo. 

However, we should take care of incompressible data (OUT_LEN >= IN_LEN) and do without compression in that case, i.e. send the original data. How to flag packets that are (not) compressed? Maybe bit 7 of the byte-sized version field?

Also, compression destination buffer size "OUT_LEN" must be larger than "IN_LEN" N2N_PKT_BUF_SIZE, namely `#define OUT_LEN     (IN_LEN + IN_LEN / 16 + 64 + 3)` according to line 55 of the example. Calling `.fwd` and `.rev` in `edge_utils` should be adopted accordingly.

Logan007:
Making up my mind, I guess it would not be a good idea to publicly indicate payload characteristics, i.e. if it is compressed or not. A much better idea would be to use bit no. 7 of the padding length byte at the end of the payload. That way, the decode function can determine after decryption whether an additional deflate step is required or not.

Logan007:
Please find attached a [patch](https://github.com/ntop/n2n/files/3238490/compression.txt).

emanuele-f:
Using the padding field like this is a bit hacky

Logan007:
Another way to see it is that technically, bit 7 to 4 are not used and still available...  And it saves a dedicated byte field.

Logan007:
We could also indicate compression using the version field filled with a dedicated aes version number. Maybe `2` could stand for "_a compressed version `1`_" (or `0` and `1` respectively if reconsidering #150)?

emanuele-f:
Compression should be a separate field. An host can receive compressed packets but send uncompressed packets

Logan007:
Maybe a `compression_version` field directly after the `aes_version` field? The `compression_version` field could indicate what compression, e.g. compression method and strength, are used and need to be applied after decryption. For outgoing packets, there could be a `#define` just as with the `aes_version`. An 8-bit field should be big large enough in size?

Logan007:
...the field should be part of the encrypted payload, though (see [above](https://github.com/ntop/n2n/issues/91#issuecomment-497091314)) - not giving any hints on payload characteristics.

Logan007:
[...]

PR #172, opened 2019-08-02
made use of minilzo compression

Logan007:
As discussed in #91, the AES mode now implements the already included minilzo compression. It is enabled by default for outgoing packets and can be disabled by removing the `#define COMPRESSION_ENABLED`-line in the `transform_aes.c`-file. The ability to deflate compressed packets from other nodes remains.

A byte-sized compression indicator field was added to prepend the actual data payload. This flag also gets encrypted for an observer is not able to distinguish packet characteristics – encrypted vs. not-encrypted.

The `./benchmark`-utility now indicates extremly exciting AES throughput! The more compressible the data is, the less encryption steps need to be performed resulting in better overall performance.


transform_aes.c
```diff
@@ -38,6 +38,20 @@
...
+
+#define COMPRESSION_ENABLED	/* defines if compression is enabled for outgoing packets; comment out for no compression */
+				/* compressed packets can always be decoded */
+
+/* Work-memory needed for compression. Allocate memory in units
+ * of 'lzo_align_t' (instead of 'char') to make sure it is properly aligned.
+ */
+#ifdef COMPRESSION_ENABLED
+#define HEAP_ALLOC(var,size) lzo_align_t __LZO_MMODEL var [ ((size) + (sizeof(lzo_align_t) - 1)) / sizeof(lzo_align_t) ]
+static HEAP_ALLOC(wrkmem, LZO1X_1_MEM_COMPRESS);
```
emanuele-f, line 52: Is this used?

transform_aes.c
```diff
@@ -76,9 +90,10 @@ static void set_aes_cbc_iv(transop_aes_t *priv, n2n_aes_ivec_t ivec, uint64_t iv
  *
  *  - a 8-bit aes encoding version in clear text
  *  - a 64-bit random IV seed
- *  - encrypted payload.
+ *  - a 8-bit compression indicating field; encrypted, together with the following...
```
emanuele-f: It's better to have a generic "encrypted flags" fields, which will give us future extensibility

transform_aes.c
```diff
@@ -38,6 +38,20 @@
 #define TRANSOP_AES_IV_KEY_BYTES (AES128_KEY_BYTES) /* use AES128 for IV encryption */
 #define TRANSOP_AES_PREAMBLE_SIZE (TRANSOP_AES_VER_SIZE + TRANSOP_AES_IV_SEED_SIZE)
 
+/* AES ciphertext preamble */
+#define TRANSOP_AES_COMPRESSION_SIZE   1
```
emanuele-f: I think this should be set to 0 when compression is disabled

transform_aes.c
```diff
@@ -182,25 +221,46 @@ static int transop_decode_aes( n2n_trans_op_t * arg,
...
                      * enough to prevent segfaults. */
                     traceEvent(TRACE_DEBUG, "padding = %u", padding);
                     len -= padding;
-
-                    memcpy( outbuf, 
-                            assembly, 
-                            len);
+		    len -= TRANSOP_AES_COMPRESSION_SIZE;
+
+		    /* decompress if necessary */
```
emanuele-f, line 243: `#ifdef COMPRESSION_ENABLED`?

transform_aes.c
Logan007: Yes, it is. For compression (line 139).

transform_aes.c
Logan007: Okay. I guess this should be regular part of packet spec then. I will rename the "compression indicating field" in the code to something more generic. Is 8-bit size okay? Or shall we go for 16?

transform_aes.c
Logan007: If we want to keep it as "encrypted flags field" instead, it should remain (unused) in the packet. It would allow compatibility among versions with and without encryption enabled. Let me know.

transform_aes.c
Logan007: For the sake of being able to decompress packets even if a node does not actively makes use of compression, I did not def'ed it out by intent. This allows mixed environments where nodes with some weaker CPU shall not compress - but should be able to decompress in case they are talked to.

But we can switch to the "either ... or"-way which would force all nodes of one network to share the same compression setting. Please advice.

transform_aes.c
emanuele-f: 8 bits are fine for now

Logan007: In the latest commit, I renamed the field and the makros accordingly. Also, I adopted the comments.

Logan007:
`huffman.c` should not be here (yet), still working on it for it will give better compression than minilzo. I hope, these "deletions" affect my commit only; it is not meant to delete those files in the dev branch - of course! :)

transform_aes.c
Logan007: As renamed now, as well defined generic field (8-bit size as discussed earlier), it could indicate other compression methods or some error correction codes or ... any other special data treatment option. 

I guess this is a design decision - do we want to permanently add this data flag field (for compatibility)?

transform_aes.c
Logan007: This too seems to be a design decision.

transform_aes.c
Logan007: Done.

transform_aes.c
Logan007: Done.

Logan007:
I found that there already has been a `N2N_COMPRESSION_ENABLED` definition in `n2n.h` which never got used properly. So, I took that one as a starting point for the newest committed implementation of this pull request.

That means that compression can be disabled in `n2n.h`by just commenting it out. It is present by default.

Please be aware that all edge nodes of one network need to be compiled with the same setting. My experience is that you need to run `make clean` after changing this setting and before recompiling n2n using `make`.

Also, I restored `Makefile.in`. :blush:

This pull request should be complete now.

By the way, talking about the Makefile, just an idea which is way beyond my abilites to implement: To conditionally prevent the `minilzo.o` file being built and linked/compiled in the executables, the `N2N_COMPRESSION_ENABLED` might be a definition to handle somewhere in the Makefile. That way, those who do not want compression could generate smaller binaries.

Logan007:
Closed as [discussed](https://github.com/ntop/n2n/issues/91#issuecomment-520257514) in #91.

Issue #195, opened 2019-12-28
Enable embedding the supernode


pojntfx:
It would be awesome if it would be possible to embed the supernode in an application. I am currently writing some (very basic, I have never written any C before and am just a novice when it comes to Go) [Go bindings for n2n](https://github.com/pojntfx/gon2n). The bindings will also have a command for edges and supernodes and I plan to write a Kubernetes CNI plugin with them. While the public API as defined in `n2n.h` allows for easy embedding of `edge`, it is currently not possible (at least not with my current knowledge of C) to embed the supernode. To do so, I have created a [temporary fork of n2n](https://github.com/pojntfx/n2n) in which I have decomposed the supernode in a similar way that the edge is currently (`edge.c`, `edge_utils.c` => `sn.c`, `sn_utils.c`), and if there is interest in changing the API to allow for embedding I'd be more than happy to start a pull request.

pojntfx:
Alright, I've got a working example for supernode embedding with the new public functions in the [fork](https://github.com/pojntfx/n2n):

```c
#include "n2n.h"

static int keep_running;

int main()
{
    n2n_sn_t sss_node;
    int rc;

    sn_init(&sss_node);
    sss_node.daemon = 0;   // Whether to daemonize
    sss_node.lport = 1234; // Main UDP listen port

    sss_node.sock = open_socket(sss_node.lport, 1);
    if (-1 == sss_node.sock)
    {
        exit(-2);
    }

    sss_node.mgmt_sock = open_socket(5645, 0); // Main UDP management port
    if (-1 == sss_node.mgmt_sock)
    {
        exit(-2);
    }

    keep_running = 1;
    rc = run_sn_loop(&sss_node, &keep_running);

    sn_term(&sss_node);

    return rc;
}
```

This makes it possible to write the [corresponding glue C header](https://github.com/pojntfx/gon2n/blob/master/pkg/supernode.h), [C file](https://github.com/pojntfx/gon2n/blob/master/pkg/supernode.c) and [Go code](https://github.com/pojntfx/gon2n/blob/master/pkg/supernode.go).

pojntfx:
I have also created a seperate branch that adds another example that demonstrates a more advanced embedding of `edge` without the `quick_edge_init` function [in a seperate branch](https://github.com/pojntfx/n2n/pull/1).

The example looks like this and attempts to show all available configuration options:

```c
#include "n2n.h"

static int keep_running;

int main()
{
    n2n_edge_conf_t conf;
[...]

PR #196, opened 2019-12-29
Refactor supernode to allow for embedding and add more embedding examples

pojntfx:
See #195 

This PR allows for the embedding of the supernode in another application (in my specific case, to write [Go bindings for n2n](https://github.com/pojntfx/gon2n)). In addition to that, it also adds more examples for embedding, which were created in order to show a more advanced custom configuration.

The API was changed to allow for the following "high-level start functions" to be created in the bindings:

[Supernode starter](https://github.com/pojntfx/gon2n/blob/master/pkg/supernode.c)

```c
#include "n2n/n2n.h"

int supernode_configure(n2n_sn_t *sss_node, int lport);

int supernode_open_lport_socket(n2n_sn_t *sss_node);

int supernode_open_mgmt_socket(n2n_sn_t *sss_node, int mgmt_port);

int supernode_start(n2n_sn_t *sss_node, int *keep_running);

int supernode_stop(n2n_sn_t *sss_node);
```

This was achieved by decomposing `sn.c` into `sn_utils.c`, `sn.c` and defining the new public functions in `n2n.h`.

The relevant new examples are [example_edge_embed.c](https://github.com/pojntfx/n2n/blob/dev/example_edge_embed.c) and [example_sn_embed.c](https://github.com/pojntfx/n2n/blob/dev/example_sn_embed.c); I have moved the edge embed example with the `quick_edge_init` function to [example_edge_embed_quick_edge_init.c](https://github.com/pojntfx/n2n/blob/dev/example_edge_embed_quick_edge_init.c).

Please take all of this with a grain of salt. I have never written any C before and followed only a simple C tutorial beforehand (I only know Go/JS/Ruby) ;)

### DIFF
--- a/tuntap_osx.c
+++ b/tuntap_osx.c
@@ -46,7 +46,7 @@ int tuntap_open(tuntap_dev *device /* ignored */,
   }
   
   if(device->fd < 0) {
-    traceEvent(TRACE_ERROR, "Unable to open tap device %s", tap_device);
+    traceEvent(TRACE_ERROR, "Unable to open any tap devices /dev/tap0 through /dev/tap254. Is this user properly authorized to access those descriptors?");
     traceEvent(TRACE_ERROR, "Please read https://github.com/ntop/n2n/blob/dev/doc/n2n_on_MacOS.txt");
     return(-1);
   } else {


sn_utils.c
```diff
@@ -0,0 +1,784 @@
...
+    sss->lport = N2N_SN_LPORT_DEFAULT;
+    sss->sock = -1;
+    sss->mgmt_sock = -1;
+
+    return 0; /* OK */
+}
+
+/** Deinitialise the supernode structure and deallocate any memory owned by
+ *  it. */
+extern void sn_term(n2n_sn_t *sss)
```
emanuele-f: Avoid using `extern` as it is implicit in C for function names. Moreover here the function should be `static` and not `extern` as it is only used within this file. Please fix this for all the functions.

emanuele-f (changes requested):
The file sn.c has too many changes and it's very difficult to review. Please make minimal changes (e.g. avoid changing the existing code indentation and when you move code from sn.c to sn_utils.c put that change into a separate commit) in order to make it possible for a quick review of the changes. Please start with a clean pull request

pojntfx:
Oh, yes it looks like I've kept VSCode's formatter on; I'll go ahead and do that once again, but this time without the formatter and minimal changes. Thanks for clearing up the use of `extern` as well.

pojntfx:
Alright, first of all sorry for taking so long, I was busy with some other development. I'll start a clean PR today.

sn_utils.c
pojntfx: Addressing this in the new PR.

PR #199, opened 2020-02-03
Update tuntap_osx.c error message.

Marviel:
Adds clearer explanation of particular tuntap failure.

Previously, it would give a slightly confusing message about "Unable to open tap device /dev/tap254" whenever it failed -- even though the connection process attempted every number from 0 to 254 -- not just 254.

Also added a prompt to check file permissions on failure.


emanuele-f:
Thank you

PR #209, opened 2020-03-16
Enable embedding the supernode, add more embedding examples (clean)

### LINKED ISSUE #195: Enable embedding the supernode

pojntfx:
Closes #195 

Thanks for your patience! This is a clean re-implementation of #195 .

See the embedding features in action here: [pojntfx/gon2n](https://pojntfx.github.io/gon2n/)!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,9 @@ config.*
 Makefile
 autom4te.cache
 edge
+example_edge_embed_quick_edge_init
 example_edge_embed
+example_sn_embed
 supernode
 tools/n2n-benchmark
 tools/n2n-decode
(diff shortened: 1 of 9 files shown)


lucaderi:
Thank you! Great work

PR #213, opened 2020-03-20
AES -- (better) hardware acceleration using openssl's evp interface

### LINKED ISSUE #214: AES: provide (better) hardware acceleration
Reading some hints that openssl's `AES_encrypt_*` interface did not provide hardware acceleration, I dug into its code. Indeed, it seems that it employs a software based encryption only – I did not check the various ASMs for occurences of AES-NI though. However, that already was enough reason to start testing openssl's alternative `evp_*` which is [said](https://www.openssl.org/docs/man1.1.0/man3/EVP_EncryptInit.html) (Notes section) to guarantee the support of available hardware acceleration.

Please find the result in pull request #213 . ~~A tidied update will follow soon (remove one variable and anything else you might find).~~ _[done]_ Also, I have attached a [patch](https://github.com/ntop/n2n/files/4360880/patch.txt) as it might be easier to apply for some testing.

`tools/n2n-benchmark` reports __plus (!) 128 %__ better AES encryption-throughput (397.4 MB/s) on my i7-2860QM (compared to current dev: 174.6 MB/s) which might be an indicator for actual use of hardware acceleration. _[EDIT: data rates from new benchmark runs w/o other applications running in the background]_

Transmission (which also yielded higher throughput) was tested against an original edge built from current dev to ensure compatibility which seems to be flawless.

My guess is that openssl's `AES_encrypt_*` already is quite fast compared to n2n's Twofish as it might use platform-optimized code (pure ASM) if appropriately compiled but not the maybe available, dedicated AES-NI instructions yet.

~~A naturally next step would be to also use openssl's `evp_*` for the remaining IV related AES-ECB step.~~ Applying openssl's `evp_*` also to the one block IV is bogus to throughput (362.4 MB/s) and thus not implemented there. The hardware acceleration does not seem to pay off for a single block to encrypt due to the more complicated `evp_*` setup.

By the way, another significant encryption speed-up could be achieved by using AES in CTR-mode instead of CBC. But for security reasons, it would require to __sharply rethink__ the complete __IV handling__ – especially size, randomness, and its transmission. 

Any thoughts? How do your benchmarks compare?

Logan007:
Reading some hints that openssl's `AES_*_encrypt` interface did not provide hardware acceleration, I dug into its code. Indeed, it seems that it employs a software based encryption only – I did not check the various ASMs for occurences of AES-NI though. However, that already was enough reason to test the alternative `evp_*` which is [said](https://www.openssl.org/docs/man1.1.0/man3/EVP_EncryptInit.html) (Notes section) to guarantee the support of available hardware acceleration.

Please find the result in this pull request. ~~A tidied update will follow soon (remove one variable and anything else you might find).~~ _[done]_

`n2n-benchmark` outputs __plus 120 %__ better throughput on my i7-2860QM (compared to current `dev`) which might be an indicator for actual use of hardware acceleration. 

Transmission (which also yielded higher bandwidth) was tested against an original edge built from current `dev` to ensure compatibility which seems to be flawless.

Closes #214.


Logan007:
Instead of deleting and not using the `evp_plaintext_len` and `evp_ciphertext_len` variables, I used them for an additional length check to compare if openssl's `evp_*` en-/decryption routines deliver the expected number of bytes.

Some typos were removed from the newly introduced  `TRACE_ERROR`s.

Also, I want to give a bit more of an explanation: The `evp_*` encryption and decryption are implemented very close to the C-language examples `encrypt` and `decrypt` as found in [openssl's Wiki](https://wiki.openssl.org/index.php/EVP_Symmetric_Encryption_and_Decryption). Error handling is adopted to n2n's nested code style. `evp_*` related changes to the setup of cipher mode and key can be seen in the `setup_aes_key` routine.

The only addition (besides the aforementioned lengh check) is disabling openssl's default PKCS#7 padding scheme calling `EVP_CIPHER_CTX_set_padding(ctx, 0)` to preserve compatibility to the so far implemented AES. Even though n2n's own un-padding routine can handle PKCS#7 for decryption, it does not work vice versa: openssl's `EVP_DecryptFinal_ex`  throws an error during the decryption of the final block if it receives n2n's `00 ... 00 [len]`-padding where `[len] ... [len]` is expected. However, if it were decided to have openssl also handle the padding (breaking compatibility to current dev), this – along all of n2n's padding handling – would be the line to delete.

Logan007:
The `evp_*` contexts get reused. Init happens only once during `n2n_transop_aes_cbc_init(...)`, they get freed in `transop_deinit_aes(...)`. Contexts need to get reset after each use.

Latest changes just look odd in git diff as indention had to be changed by one level due to the moved context initialization. 

In my limited test environment, it is fully working.

lucaderi:
Thank you. I have made a few fixes for MacOS but everything works

PR #223, opened 2020-04-10
Fix OSX Catalina compilation failure

### LINKED ISSUE #222: Cant compile n2n on Mac Catalina (10.15.4)
I download v2.6-stable on mac catalina, but when i use cmake to compile the source , I got the error :
"ld: cannot link directly with dylib/framework, your binary is not an allowed client of /usr/lib/libcrypto.dylib for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)"

I set the OPENSSL_ROOT_DIR to /usr/local/opt/openssl  but no use .

So Any help would be appreciate !

cnanakos:
Include openssl library path for n2n-benchmark compilation

Fixes https://github.com/ntop/n2n/issues/222

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,7 +92,7 @@ install(TARGETS edge supernode
 include_directories(.)
 
 add_executable(n2n-benchmark tools/benchmark.c)
-target_link_libraries(n2n-benchmark n2n)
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #224, opened 2020-04-14
added list of currently managed communities to supernode's management port output

Logan007:
The management port output was extended by adding a list of currently managed communities `cur_cmnts`. The community names are square-bracketed to visualize leading and/or trailing spaces.

This feature might come in helpful especially in case of an _open_ supernode which has not been locked to a predetermined set of allowed communities by not using the `-c <file>` command line option.


lucaderi:
Thank you

PR #233, opened 2020-04-25
Typo

Logan007:
This pull request fixes a typo that has prevented the proper usage of openSSL's evp_* interface.

Issue #230, opened 2020-04-24
add chachao20-ietf-poly1305 support


xiamr:
There are two algorithms for encryption in n2n, twofish and AES-CBC. But it is not suitable for router CPU without AES-NI support.  Chacha20-ietf-Poly1305, as well as a variant  with an extended nonce: XChaCha20-ietf-Poly1305, both are secure and high performace for such situation and recommanded by libsodium library, so it is worth implementing for n2n.

xiamr:
Could you support it ? Or I will consider to implement my own version.

Logan007:
I plan to implement a lightweight cipher for header encryption (#198). I could try to expand it to its own packet encryption transform – if supported.

For now, I have chosen SPECK, which is extremely simple to implement.

As for Poly1305, I am not sure if n2n can already bear a Message Authentication Code as this seems planned to be taken in a more general approach.

xiamr:
You do not need implement cipher by your own, other libraries such as libsodium provides a friendly API for app development. Choosing a well known cipher is benefit for security, 
For header data, you can only use authentication without encryption. Authentication is essential when packet across untrusted network while firewall can change the content.

Logan007:
I just realized that openSSL [supports](https://www.openssl.org/docs/man1.1.1/man3/EVP_chacha20_poly1305.html) ChaCha20 cipher as `const EVP_CIPHER *EVP_chacha20(void)`.

Now that `transform_aes.c` supports openSSL's `evp_*` interface, testing should be just as easy as changing the cipher in lines `348`, `355`, and `364` accordingly. ChaCha20's fixed key size of 256 bits might require some additional, slight changes to the code. I would be very interested in your results, please share!

A performance comparison running `openssl speed -evp chacha20` and `openssl speed -evp aes-128-cbc` makes ChaCha20 look promising even on AES-NI accelerated CPUs.

[...]

PR #235, opened 2020-04-26
added ChaCha20

Logan007:
This pull request adds ChaCha20 (cc20) stream cipher to the code. It compiles well and it already gets used in `tools/n2n-benchmark`.¹

There has been no discussion yet on how to activate it via cli options and how it could work next to the mutually exclusive other cipher option `-A`. Therefore, for now, to use it for transmission, it needs to be explicitly activated as default cipher by editing the code of `edge.c` at line 273 to `conf->transop_id = N2N_TRANSFORM_ID_CHACHA20;`.

From the author's point of view, this pull request leaves a few TODOs before being considered acceptable: 

- discuss cli option and implement
- make use of a less predictable random number generator
- clean-up

Please review, test and advise.

Solves #230. Further details discussed there.

¹ CBC decryption is much faster than encryption. To therefore get a fair comparison, the tool should be run using `-d` option. Also, pull request #233 should be applied beforehand – otherwise, possibly available AES hardware acceleration remains unused.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -50,7 +50,7 @@ MAN8DIR=$(MANDIR)/man8
 N2N_LIB=libn2n.a
 N2N_OBJS=n2n.o wire.o minilzo.o twofish.o \
 	 edge_utils.o sn_utils.o \
-         transform_null.o transform_tf.o transform_aes.o \
+         transform_null.o transform_tf.o transform_aes.o transform_cc20.o \
          tuntap_freebsd.o tuntap_netbsd.o tuntap_linux.o \
 	 tuntap_osx.o
 LIBS_EDGE+=$(LIBS_EDGE_OPT)
--- a/edge.c
+++ b/edge.c
@@ -143,7 +143,7 @@ static void help() {
 #ifndef __APPLE__
 	 "[-D] "
 #endif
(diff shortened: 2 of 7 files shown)


Logan007:
A command-line option to enable ChaCha20 was added ( `-A4`). Along with it, the scheme to use the `-A` option was extended, maintaining n2n's behavioural compatibility to current dev:

`-A1`: do not use payload encryption (`transform_null`, this is default if neither key nor any `-A_` option is given)
`-A2`: use Twofish (default, if a key is given __and__ none of the `-A_` options is given)
`-A3` or `-A`: use AES-CBC
`-A4`: use ChaCha20

Solitary use of `-A` is marked deprecated and should be removed in some future release. Note: It is not the AES-CBC cipher that gets removed, just use `-A3` instead. With a broader choice of ciphers, the need to explicitly choose and indicate a cipher becomes evident.

From my point of view, this pull request is complete now.

_P. S._ Fun fact: The parameter for to choose a cipher unintentionally got somewhat mnemonic: `A2` is for __2__-Fish, `A3` is for __Ae__-s (leetly), and `A4` is for ch-__A__-ch-__A__ (also leetly).

Logan007:
Just realized that `htobe64(...)` from `endian.h` might not be available on windows and maybe others. We could pull it in from somewhere else as I might need it for header encryption as well. Different approaches are possible, maybe a short form of [this](https://gist.github.com/panzi/6856583).

lucaderi:
Thank you

PR #236, opened 2020-04-26
transform_aes.c -- make openssl_err_as_string() static

Logan007:
This pull request makes openssl_err_as_string() static to avoid conflicts with other ciphers that would also use their own same-named function – those can't rely on the AES cipher being around and offering this function.


lucaderi:
Thank you

Issue #198, opened 2020-01-23
Header Format


Logan007:
Having today's and maybe future featureset (encryption scheme, IV, compression indicator, DH key exchange, FEC, MAC, ...) of n2n in mind, I want to bring up a discussion about standardization of the header preceeding the actual data: its format as well as its encryption.

The header should be the same format for all encryption schemes (as of now: AES and TWOFISH) and allow for other modular extensions to be used, e.g. own encryption or compression schemes.

Any ideas or suggestions?

fengdaolong:
I agree with this proposal.

Logan007:
As found in `doc/HACKING`, current header format seems to be as follows:

```
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   ! Version=2     ! TTL           ! Flags                         !
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 4 ! Community                                                     :
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
 8 ! ... Community ...                                             :
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
12 ! ... Community ...                                             :
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
16 ! ... Community ...                                             !
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
20 ! Source MAC Address                                            :
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
24 :                               ! Destination MAC Address       :
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
[...]

PR #237, opened 2020-05-03
added optional payload compression (lzo) for all transforms

Logan007:
This pull request adds optional payload compression (lzo) for all transforms. It is disabled by default and can be enabled using the new `-z` cli option – symbolizing a squeezed bouncy spring.

There is no need for having all edges enable compression. They all (after applying this pull request) will be able to decompress which is less costly in terms of required CPU power. Compression could be enabled on stronger CPUs only – maybe on desktop grade computers or any other use-case where you need to trade CPU power for bandwidth.

By free-riding on the uppermost bits of the 16-bit _transform_ field (which indicate the encryption type if any) in the packet, this remains compatible to current dev. As soon as major changes disrupt compatibility anyway, _compression_ could easily get its own field by appropriately changing `encode_PACKET(...)` and `decode_PACKET(...)` in `wire.c`.

The code is prepared to easily take your additional compression algorithms. For example, see the `switch` statements in `edge_utils.c`. Those could be chosen by having `-z` accept optional parameters, e.g. `-z2`.

Further discussion in #91 and #198.

Solves #91.

### DIFF
--- a/edge.c
+++ b/edge.c
@@ -175,6 +175,8 @@ static void help() {
 #ifdef N2N_HAVE_AES
   printf("-A                       | Use AES CBC for encryption (default=use twofish).\n");
 #endif
+  printf("-z                       | Enable lzo1x compression for outgoing data packets\n");
+  printf("                         | (default=disabled).\n");
   printf("-E                       | Accept multicast MAC addresses (default=drop).\n");
   printf("-S                       | Do not connect P2P. Always use the supernode.\n");
 #ifdef __linux__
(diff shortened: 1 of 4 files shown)


lucaderi:
Thank you. At some point we need to review all the options and remove some of them in order to simplify the CLI options

Logan007:
Concerning the options, I fully agree.

Maybe, network options `-d`, `-a`, `-s` could be reduced to some `<ipaddress>/<netmask>:<port>%<dev>` single parameter syntax with optional parts, also `-u` and `-g` to some `<uid>:<gid>` syntax (with optional `<uid>` and optional `:<gid>`). Other clusters could be created based on functionality, e.g. `-L` and  `-i` as well as `-M` and `-D`. Would require some additional parsing efforts, though.

Also, features (compression) could just become default scratching the corresponding cli option. Those features could be disabled¹ at compile time using some `-D TINY` – if required.

¹ in case of compression: do not disable the ability to decompress

Issue #157, opened 2019-07-01
flexible IV seed size


Logan007:
As of now, the IV is composed of the 64-bit seed and another 64-bit derivate from the pre-shared secret key which get encrypted as a whole to form the 128 bit IV.

While this offers good security to most users, some users might want to take advantage of a fully 128-bit random seed whereas others might even want to lower the security needs for the sake of shorter packets.

With this in mind, I suggest to implement a more flexible way of handling the IV which would allow for a flexible IV seed length between 0 and 16 bytes, the default could remain at the current value of 8 keeping up compatibility with current dev. It might be worth a thought to even completely disable IV handling if length is set to 0 (and thus saving the additional encryption step).

As this is an extremely relevant security setting, it should be set by consequence-aware users only. That is why I recommend this feature to be controlled by a `#define` in the code.

The implementation would require a few adoptions of data types in `transform_aes.c`. If you want me to, I would be ready to implement it as foreshadowed [here](https://github.com/ntop/n2n/pull/118/commits/df3ae56609b287ff8bff4f8785d7d32386f0c617#diff-1444ba9e46de97ec079a106ed01d0431).

A prerequisite would be a solution to #117 which would allow to easily obtain random numbers of flexible length - concededly even based on `rand()` in the first step.

Any thoughts?

Logan007:
Just added a pull request (#239) that can handle all sizes from `0` to `16`bytes as value for 
`TRANSOP_AES_IV_SEED_SIZE`.

It calls the random number generator (which should be replaced by a stronger one... #117) only as often as necessary. 

The output of the (flexible length) seed to the trace required implementing a short extra routine `to_hex`.

First comparisons (between the `8`-columns) using software AES without evp show either no or no significant difference in performance:

CPU | hardwired `8` | flex `16` | flex `8` | flex `1`  | flex `0`
:--|--:|--:|--:|--:|--:
Cortex A53s | __50.1__ MB/s | 49.3 MB/s | __50.1__ MB/s | 50.9 MB/s | 51.2 MB/s
Core i5 M430 | --- MB/s | --- MB/s | --- MB/s | --- MB/s | --- MB/s
Celeron 3865U | --- MB/s | --- MB/s | --- MB/s | --- MB/s | --- MB/s
i7 2860QM | __182.3__ MB/s | 179.9 MB/s | __181.7__ MB/s | 182.6 MB/s | 183.0 MB/s
[...]

PR #239, opened 2020-05-05
allowed flexibe iv seed size (AES)

Logan007:
This pull request adds the ability to handle all iv seed sizes from `0` to `16` bytes by changing the definition of `TRANSOP_AES_IV_SEED_SIZE` in the `transform_aes.c` file. The current version would only allow up to `8` bytes.

The default value remains set to `8` so it is fully compatible to current dev. 

Only difference: iv seed is output to trace in memory/network order.

Further discussion in #157.

Solves #157.


lucaderi:
Thank you

PR #242, opened 2020-05-09
moved from openSSL to bearSSL

### LINKED ISSUE #231: Can you use bearssl to replace openssl?
bearssl: 
https://bearssl.org/

Bearssl is much better than openssl.
Openssl has too many bugs

Logan007:
This pull request lets n2n's AES encryption use bearSSL instead of openSSL. It is fully compatible to current dev.

Details are discussed in #231.

Closes #231.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -13,27 +13,14 @@ else
 GIT_RELEASE=${N2N_VERSION_SHORT}
 fi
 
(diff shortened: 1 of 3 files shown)


lucaderi:
Please see my comment on https://github.com/ntop/n2n/issues/231. I think it's time to speak a bit and decide what to do next in n2n

PR #246, opened 2020-05-19
added optional payload compression (zstd) for all transforms

Logan007:
This pull request adds optional payload compression using zstd (if installed). If the library is found at compile time, it gets used.

The `-z` cli parameter handles compression as follows:
`-z1` or `-z` LZO compression (always available in n2n)
`-z2` ZSTD compression (if available/compiled)
Omitting any `-z_` parameter leaves the compression disabled for outgoing packets. 

However, an edge node will be able to decompress all incoming compressed packets – lzo (always) and zstd (if compiled). That way, _mixed_ networks in which weaker edges have compression (more costly in terms of CPU cycles) disabled will work.

In n2n, the compression level of zstd defaults to `7`. It is set by a `#define ZSTD_COMPRESSION_LEVEL` in line 173 of the `n2n.h` file.

This implementation does not change the packet layout. The compression indicator still free-rides on the transform indicator as introduced with lzo compression. It should be fully compatible to current dev.

See the (meanwhile closed) discussion of #91 for some more thoughts about compression in n2n.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -13,14 +13,24 @@ else
 GIT_RELEASE=${N2N_VERSION_SHORT}
 fi
 
+N2N_LIBS=
+
+AC_CHECK_LIB([zstd], [ZSTD_compress])
+
+if test "x$ac_cv_lib_zstd_ZSTD_compress" != xyes; then
+  AC_MSG_RESULT(Building n2n without ZSTD support)
+else
+  AC_DEFINE([N2N_HAVE_ZSTD], [], [Have ZSTD support])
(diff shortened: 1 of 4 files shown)


lucktu:
If it is not absolutely necessary (or a major change) , it is recommended that the options be kept to a minimum for n2n is a light vpn.

Logan007:
I fully agree. That's why I re-used the `-z` option which was introduced with lzo-compression. 

I made some [proposals](https://github.com/ntop/n2n/pull/237#issuecomment-626135632) how we could reduce the number of options. What do you think?

Shall we open a new issue to discuss a future cli-option layout?

PR #249, opened 2020-05-27
added Speck

Logan007:
This pull request adds SPECK cipher at 128-bit block size and 256-bit key size in CTR mode to the code. It compiles well and can be used using `-A5` (__5__ is for __S__ peck). Also, it already gets used in `tools/n2n-benchmark` and performs between Twofish (roughly four times as fast) and hardware-accelerated AES (roughly between 60 % and 80% of its rates). In case hardware-accelerated AES is not available, it performs roughly twice as fast as software-based AES – as just seen on Raspberry Pi 3B.

It builds on the patterns used in #235. SPECK is implemented along the official Implementation guide. It does not require external libraries and thus could make a permanently available option. 

This pull request is not complete yet, I will rework it for some optimization, key-setup and big-endian testing ~~– __is there anyone who could compile and run `speck.c`'s commented `main` on a big-endian machine, please?__~~ [_done_] Take it more as a first test and impression and do not use it for productive networks yet as the random numbers are still from C's `rand()`.

Also, I wanted to get more comfortable with the cipher I plan to use for header encryption (#198) and thought I could already share it. Think of it as AES' little sister. How does it work for you?

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -48,9 +48,9 @@ MAN7DIR=$(MANDIR)/man7
 MAN8DIR=$(MANDIR)/man8
 
 N2N_LIB=libn2n.a
-N2N_OBJS=n2n.o wire.o minilzo.o twofish.o \
+N2N_OBJS=n2n.o wire.o minilzo.o twofish.o speck.o \
 	 edge_utils.o sn_utils.o \
-         transform_null.o transform_tf.o transform_aes.o transform_cc20.o \
+         transform_null.o transform_tf.o transform_aes.o transform_cc20.o transform_speck.o \
          tuntap_freebsd.o tuntap_netbsd.o tuntap_linux.o \
 	 tuntap_osx.o
 LIBS_EDGE+=$(LIBS_EDGE_OPT)
--- a/Makefile.in
+++ b/Makefile.in
@@ -7,7 +7,7 @@ GIT_COMMITS=@GIT_COMMITS@
 
 CC?=gcc
 DEBUG?=-g3
-OPTIMIZATION?=-O3
+OPTIMIZATION?=-O3 -march=native
 WARN?=-Wall
 
 #Ultrasparc64 users experiencing SIGBUS should try the following gcc options
--- a/edge.c
+++ b/edge.c
@@ -179,6 +179,7 @@ static void help() {
 #endif
 #ifdef HAVE_OPENSSL_1_1
   printf("-A4                      | Use ChaCha20 for payload encryption. Requires a key.\n");
+  printf("-A5                      | Use Speck    for payload encryption. Requires a key.\n");
 #endif
   printf("-E                       | Accept multicast MAC addresses (default=drop).\n");
   printf("-S                       | Do not connect P2P. Always use the supernode.\n");
@@ -324,6 +325,11 @@ static int setOption(int optkey, char *optargument, n2n_priv_config_t *ec, n2n_e
 	  break;
 	}
 #endif
+      case 5:
+	{
+	  conf->transop_id = N2N_TRANSFORM_ID_SPECK;
+	  break;
+	}
       default:
 	{
  	  conf->transop_id = N2N_TRANSFORM_ID_INVAL;
--- a/edge_utils.c
+++ b/edge_utils.c
@@ -139,6 +139,7 @@ const char* transop_str(enum n2n_transform tr) {
   case N2N_TRANSFORM_ID_TWOFISH: return("twofish");
   case N2N_TRANSFORM_ID_AESCBC:  return("AES-CBC");
   case N2N_TRANSFORM_ID_CHACHA20:return("ChaCha20");
+  case N2N_TRANSFORM_ID_SPECK   :return("Speck");
   default:                       return("invalid");
   };
 }
@@ -247,6 +248,10 @@ n2n_edge_t* edge_init(const tuntap_dev *dev, const n2n_edge_conf_t *conf, int *r
     rc = n2n_transop_cc20_init(&eee->conf, &eee->transop);
     break;
 #endif
+  case N2N_TRANSFORM_ID_SPECK:
+    rc = n2n_transop_speck_init(&eee->conf, &eee->transop);
+    break;
+
   default:
     rc = n2n_transop_null_init(&eee->conf, &eee->transop);
   }
--- a/n2n.h
+++ b/n2n.h
@@ -296,6 +296,7 @@ int n2n_transop_aes_cbc_init(const n2n_edge_conf_t *conf, n2n_trans_op_t *ttt);
 #ifdef HAVE_OPENSSL_1_1
 int n2n_transop_cc20_init(const n2n_edge_conf_t *conf, n2n_trans_op_t *ttt);
 #endif
+int n2n_transop_speck_init(const n2n_edge_conf_t *conf, n2n_trans_op_t *ttt);
 
 /* Log */
 void setTraceLevel(int level);
--- a/n2n_transforms.h
+++ b/n2n_transforms.h
@@ -31,6 +31,7 @@ typedef enum n2n_transform {
   N2N_TRANSFORM_ID_TWOFISH = 2,
   N2N_TRANSFORM_ID_AESCBC = 3,
   N2N_TRANSFORM_ID_CHACHA20 = 4,
+  N2N_TRANSFORM_ID_SPECK = 5,
 } n2n_transform_t;
 
 struct n2n_trans_op;
--- a/speck.c
+++ b/speck.c
@@ -1,26 +1,223 @@
 // cipher SPECK -- 128 bit block size -- 256 bit key size
 // taken from (and modified: removed pure crypto-stream generation and seperated key expansion)
-// https://github.com/nsacyber/simon-speck-supercop/blob/master/crypto_stream/speck128256ctr/ref/stream.c
+// https://github.com/nsacyber/simon-speck-supercop/blob/master/crypto_stream/speck128256ctr/
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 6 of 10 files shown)


skyformat99:
Great work, but what I'm more interested in is the preference for lightweight, fast communication over encryption algorithms.
Is there any way to increase the success rate of p2p direct connection.

Logan007:
I share your aim of fast and lightweight encrypted communication. That is what Speck is for. The code is _extremely_ lightweight (see `speck.c`) and independent from external libraries.

Also, it is quite fast – way faster than current Twofish. I plan to look into the processor specific tricks of the official Implementation Guide to maybe give it another boost on some platforms (x86 SSE, ARM NEON) but I cannot promise anything yet.

If you see any specific issues with p2p success rate, please open a new _Issue_ describing your specific network setup and circumstances of failing p2p. Hopefully, network experts can help then – I am more into cryptographics.

skyformat99:
If this Speck is really light and fast, then it is recommended to replace twofish as the default .

Logan007:
@skyformat99 I had exactly that very same thought – yet, a previous step would require SPECK to be pulled into the code in the first place. I guess that the next _intermediate_ release could include SPECK as option, not default cipher yet as it would break compatibility to older version's command line calling.

In my opinion, the question if to use it as default or not can only be answered after having gained some experience with that cipher. Maybe with the next _major_ release then?

However, I must have lost my natural fear of hardware-accelerated vectorizing SIMD intrinsics  somewhen during the last nights. Speedwise, the SSE version of SPECK in CTR mode looks extremely promising – still not sure if to use super-fast supercop code or more comprehensible implementation following the Implementation Guide. I will publish some code and results soon.

Logan007:
I added SSE support fro SPECK on x86 CPUs along the cybercop implementation and get the following results from `tools/n2n-benchmark`:

CPU | ChaCha20¹ | SPECK Sw | SPECK SSE | SPECK NEON | SPECK AVX
:--|:--:|--:|--:|--:|--:
i7 2860QM | 714.8 MB/s | 294.4 MB/s | 919.8 MB/s | ⁿ/ₐ | ⁿ/ₐ 
i7 7500U | 1329.3 MB/s | 340.6 MB/s | 1021.6 MB/s  | ⁿ/ₐ | ??? MB/s
Cortex A53 | 119.1 MB/s | 67.4 MB/s | ⁿ/ₐ | ??? MB/s  | ⁿ/ₐ

I guess that openSSL employs an AVX-savvy implementation of ChaCha20 which would explain the relative differences diverging on the two i7's. Given those speed-ups, I might go for AVX and NEON... this is going to be fun!

@xiamr @lucktu I would be most interested in how SPECK performs on your MIPS CPUs. Would you be able to download, compile, run the `tools/n2n-benchmark` of this pull request as of up to now (`cd4aada`) and post the results – anytime, whenever it is convenient for you? Thank you!

_P. S.  And how does `d840ba` (changed function call `speck_ctr`, see below) perform on MIPS?_

¹ requires openSSL

Logan007:
Now, this is interesting: While adding NEON support for ARM CPUs on top of `cd4aada`, I found that it performed even slower than pure software SPECK (~ 50 MB/s as opposed to 67 MB/s) on Raspberry Pi 3B+. That's why I changed calling of `speck_ctr` to passing `ctx`-parameter (which holds the round keys) by reference. Indeed, the NEON implementation got faster (~ 64 MB/s). And so got the pure software implementation (~ 99 MB/s).

So, I drew two conclusions: 
- Not every ARM platform seems to benefit from the usage of NEON – this needs verification. __Is anybody able to provide comparative test results for another ARM platform (starting with `d840ba` just turn `__ARM_NEON` in `speck.c` and `speck.h` to something like `__NO_ARM_NEON` to compile the software-only version)?__ ¹
- Passing `struct`s by reference seems to pay off – that's what I thought.

But then, bringing it back to i7 CPUs, the SSE implementation surprisingly got slower (i7 2860QM: ~ 644 MB/s vs. 920 MB/s; i7 7500U: ~ 710 MB/s vs. 1022 MB/s)! Not so the pure software implementation (i7 2860QM: ~ 404 MB/s vs. 294 MB/s; i7 7500U: ~ 471 MB/s vs. 340 MB/s).

Why am I explaining all this? Because, given these results, I am pondering to use platform- or feature-dependent calling of `speck_ctr` in `transform_speck.c` passing `ctx` either by value or by reference – just to get the most speed out of it. That definitely would require explanation.

________________
¹ I even tried to compile using `-Os` as recommended by SPECK Implementation Guide for it would give better performance on some ARM systems – apparently, not for Raspberry Pi 3B+. Maybe, it is because it is a 64-bit system and OS? If you have an opportunity, please post your NEON vs. non-NEON results as well as your system name and OS version (especially bit-width). Thank you! 

If  NEON implementation should turn out to be more a sort of performance killer, I will remove it.

Logan007:
This is some flow... I have just added AVX2-support for SPECK along the supercop benchmark implementation. I got blown away by `tools/n2n-benchmark`'s output on AVX-enabled i7 7500U CPU: __1,752.6 MB/s__! Vectorization rocks!

However, I currently have no plans to add AVX512-support from which server-like computers and maybe some very modern desktop CPUs would benefit even more – I have just no hardware available to test it on.

Other changes include the memory alignment for AVX and SSE to avoid segmentation faults due to unaligned context, especially on AVX (32-byte boundary alignment) – gcc seems to do a 16-byte boundary alignment only/anyway, so at least SSE should be fine even without it. Better safe than sorry: I also implemented it for SSE. The corresponding `_mm_malloc` and `_mm__free` should be available on all AVX and SSE accelerated platforms (x86), even MSVC is said to know about it.

__Please test__ – especially on non x86 platforms – and share any findings or odd speed results. Or just run the `tools/n2nbenchmark` utility... MIPS would be really appreciated. And again, if you use ARM, please do comparative testing with and without NEON support (see [above](https://github.com/ntop/n2n/pull/249/#issuecomment-638754605)). Also, please indicate which commit you used for testing, e.g. `ec34b58`.

Meanwhile, I will be cleaning up the code, creating a proper key setup and pondering how to conditionally compile a by-value or by-reference `ctx` passing, respectively (see above), to maybe get back even some more performance on Intel CPUs – that is a really strange effect. Maybe some caching of the `const` context on stack? Why does it seem to be so successful  especially on x86 CPUs? :thinking:

_P. S. Just one thought on cipher's speed: Probably, I will never ever serve a GBit-link with n2n in my setup. But it is good to know to have a fast cipher so CPU is not kept too busy with en-/decrypting data (only)._

lucktu:
@Logan007: I'm sorry I'm late. I just found out. 

I think we should be very careful with the idea of adding a new encryption, even if it's going to speed things up a lot, but because it's not secure ( I don't like Speck), I think it's the same as not encrypting, for n2n, this is something that is completely unnecessary. 

Thanks to Logan007's efforts, we see more hope. It is recommended that you work towards these areas: increased stability, enhanced p2p, IPv6 support, simplified code, and improved efficiency. 

Once again, thank you for your contributions to n2n, which have been well received.

Logan007:
@lucktu Thank you for your comment!

I don't plan to have any cipher replaced with SPECK, it shall just be another _option_ for those who like to use it. Just to have some choice of flavor to meet everybody's taste. As additional option and with a view to its speed, it _might_ be a good candidate to become the default option. The others probably will still be available. 

I am aware of the controversy that comes along with SPECK. In my opinion, all those discussions need to be taken with a grain of salt. They happened in a somewhat heated setting. Up to now, I did not hear that this cipher were broken, did you? It draws its enormous speed from an admittedly lower security margin (25... 30 percent of the rounds – still the originally setup security margin, as far as I know, unbroken and not diminished by any attacks so far) – two properties that need to be traded for each other.

But what makes me really trust in it is the fact that NSA [recommends](https://apps.nsa.gov/iaarchive/library/ia-guidance/ia-solutions-for-classified/algorithm-guidance/algorithms-support-evolution-ia-needs.cfm) it for official governemnt use in case AES is not available.

If you see a better option – especially for _optional_ (not mandatory) header encryption (#198) which needs to be fast and independent form libraries and thus easy to implement – please let me know. To be honest, it should be really soon and be implementable in a plug-in way as the concept I have proposed for header encryption has already advanced for quite some time.

Let me admit, as a post script, that I am just not a network person, I am more into cryptographics.

Logan007:
Another intermediate commit... I found that the _passing-context-by-value-or-by-reference_ issue seems to only affect the SSE implementation. So, I switched to _passing-context-by-value_ for SSE. Speedwise, it helps  – at least on more modern CPUs:

CPU | ChaCha20¹ | SPECK Sw | SPECK SSE | SPECK NEON | SPECK AVX
:--|:--:|--:|--:|--:|--:
Cortex A53 | 129.4 MB/s | 115.9 MB/s | ⁿ/ₐ | 75.3 MB/s²  | ⁿ/ₐ
i5 M430 | 460.3 MB/s | 260.8 MB/s | 326.9 MB/s³ | ⁿ/ₐ | ⁿ/ₐ
i7 2860QM | 714.8 MB/s | 294.4 MB/s | 919.8 MB/s | ⁿ/ₐ | ⁿ/ₐ 
i7 7500U | 1329.3 MB/s | 340.6 MB/s | 1021.6 MB/s  | ⁿ/ₐ | 1752.2 MB/s


I think speed is not all too bad for a non-library stream cipher. The only thing I am still pondering is whether I should disable NEON. However, that would need more supporting benchmark data. __How does it perform on your platform?__
__________

¹ requires openSSL
² NEON performs slower than pure 64-bit software?! Is this specific to Raspberry Pi 3B+?
³ passes context by value as it generally turned out to be faster with SPECK's SSE implementation; however, passing the context by reference would be slightly faster on this somewhat older CPU: 352.8 MB/s.

lucaderi:
Thank you

PR #254, opened 2020-06-07
fixed A/z parameter handling

Logan007:
This pull request fixes parameter handling as follows:
- Properly allow `-A_` and  `-z_` be used in `.conf`-files. All, `-A_`, `-A=_` as well as `-A` syntax work in `.conf` files now. However, command line requires `-A_` or `-A` and `-z_` or `-z` respectively.
- Restore a former default behavior of defaulting to Twofish if only a key but no cipher was provided.
- Use the more general term _key_ instead of the more special _-k_ way to provide a key (environment N2N_KEY is another option to provide a key)

### DIFF
--- a/edge.c
+++ b/edge.c
@@ -188,18 +188,16 @@ static void help() {
          "                         | causes connections stall when not properly supported.\n");
 #endif
   printf("-r                       | Enable packet forwarding through n2n community.\n");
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #256, opened 2020-06-08
added XORSHIFT128+ random number generator

### LINKED ISSUE #117: Random Numbers

Logan007:
This pull request adds the XORSHIFT128+ random number generator `n2n_rand ()` to n2n. Its periodicity of 2^128 is significantly higher than of C's built-in `rand ()`. It returns 64-bit pseudo random numbers of type `uint64_t` which get down-casted to the size of the receiving data type. Therefore, it should be just as easy as _replace-and-play_.

It is seeded by calling `n2n_srand (seed)` where _seed_ denotes a 64-bit value which should be as unpredictable as possible.

Also, a seed-generator `n2n_seed ()` is provided. It generates a value as random as possible to use as seed for calling `n2n_srand (seed)`. Please note, that the here-in used x86-hardware random number generators (RDRAND and RDSEED) can get used only if compiled with `-march=native`, see `Makefile` line 10, and, if hardware supports, of course…

Also see #170. 

Closes #117.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ add_library(n2n n2n.c
                 transform_null.c
                 transform_tf.c
                 transform_aes.c
+		random_numbers.c
                 tuntap_freebsd.c
                 tuntap_netbsd.c
                 tuntap_linux.c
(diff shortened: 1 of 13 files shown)


lucaderi:
This looks like a good contribution: thank you

Issue #19, opened 2018-06-26
Implement micro-segmentation 


lucaderi:
Identify ways to allow edge nodes to connect via Internet only to the supernode, and route all the rest of traffic via n2n. This would promote the implementation of micro segmented services as well to restrict devices to selected communications. This could be interesting fields like IoT or automotive so that we can use n2n to policy traffic ad the edge and thus distributed the logic to peers.

Logan007:
Such a feature would also require to communicate with the configured DNS server to resolve the supernode's IP address in case of FQDN, especially if used with dynamic IP address and "-b" option.

I also happened to be in need for such a feature. I solved it by adding corresponding routes to the routing table - set the more general default route to use the n2n device and add another specific route to the supernode via the regular network device. The specific routing rule (as it is more specfic) will be used over the more general one to reach the supernode.

Should n2n deal with the device's routing table?

Logan007:
Lately added `-n` command line option might have solved this issue.

PR #258, opened 2020-06-09
Fix default gateway routes

emanuele-f:
The 128.0.0.0/1 route was missing causing some traffic to go outside the VPN

### DIFF
--- a/edge_utils.c
+++ b/edge_utils.c
@@ -2325,10 +2325,12 @@ static int edge_init_routes(n2n_edge_t *eee, n2n_route_t *routes, uint16_t num_r
        *  2. Add the new default gateway route
        *
        * Instead of modifying the system default gateway, we use the trick
-       * of adding a route to the 0.0.0.0/1 network, which takes precedence
-       * over the default gateway (0.0.0.0/0). This leaves the default
-       * gateway unchanged so that after n2n is stopped the cleanup is
-       * easier.
+       * of adding a route to the networks 0.0.0.0/1 and 128.0.0.0/1, thus
+       * covering the whole IPv4 range. Such routes in linux take precedence
+       * over the default gateway (0.0.0.0/0) since are more specific.
+       * This leaves the default gateway unchanged so that after n2n is
+       * stopped the cleanup is easier.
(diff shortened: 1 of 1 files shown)


edge_utils.c
```diff
@@ -2378,6 +2380,14 @@ static int edge_init_routes(n2n_edge_t *eee, n2n_route_t *routes, uint16_t num_r
       custom_route.net_bitlen = 1;
       custom_route.gateway = route->gateway;
 
+      if(routectl(RTM_NEWROUTE, NLM_F_CREATE | NLM_F_EXCL, &custom_route, eee->device.if_idx) < 0)
+	return(-1);
+
+      /* ip route add 128.0.0.0/1 via n2n_gateway */
```
lucaderi, line 2386: Can you elaborate what's special about about 0.0.0.0/1 and 128.0.0.0/1?

lucaderi (changes requested):
Please explain why these networks are special and if this code is related to a specific port or it is general to all platforms

edge_utils.c
emanuele-f: This only works in linux, it is explained in https://github.com/zerotier/ZeroTierOne/issues/178#issuecomment-204599227 . We use 0.0.0.0/1 and 128.0.0.0/1 (which covers the same IPv4 range as 0.0.0.0/0) to avoid touching the system default gateway (0.0.0.0/0). This has the advantage that if n2n is closed abnormally, the `0.0.0.0/1` and `128.0.0.0/1` routes are automatically deleted (since they are bound to the n2n interface) thus restoring the original routing table. In constrast, if we touch the default gateway and n2n is closed abnormally, the system would not have a default gateway route anymore

lucaderi: Can you please put this explanation in a code comment, and as already discussed for other tickets, if this is a Linux-only change try to keep on a separate file ?

emanuele-f:
The code already contains the references which explain it (also check out my comment above). This PR contains a fix for a bug which affects the current n2n users, is there any reason for still delaying merging this?

Logan007:
By the way, doesn't this also solve #19?

PR #259, opened 2020-06-09
Compilation fixes for hin2n

emanuele-f:
This pull request addresses the n2n compilation in Android. It also contains the changes required to make the default gateway work with hin2n (https://github.com/switch-iot/n2n_ntop/pull/4).


emanuele-f:
Update: the hin2n part is now merged, hope that this PR is merged soon to continue the n2n android integration

edge_utils.c
```diff
@@ -281,9 +286,11 @@ n2n_edge_t* edge_init(const tuntap_dev *dev, const n2n_edge_conf_t *conf, int *r
     rc = n2n_transop_cc20_init(&eee->conf, &eee->transop);
     break;
 #endif
+#ifndef __ANDROID_NDK__
```
lucaderi: Why speck is not supported on Android? Better to fix Speck issues rather to add yet another if

edge_utils.c
```diff
@@ -1554,6 +1593,57 @@ static void readFromTAPSocket(n2n_edge_t * eee) {
 
 /* ************************************** */
 
+#ifdef __ANDROID_NDK__
```
lucaderi: Can you please move all the android code in a separate file to avoid many #ifdef in the code?

speck.c
```diff
@@ -6,6 +6,8 @@
 #include <stdlib.h>
 #include <stdint.h>
 
+#ifndef __ANDROID_NDK__
```
lucaderi: Same as above

lucaderi (changes requested):

edge_utils.c
emanuele-f: this is related to the HTOLE64 function which is not available (see #260)

edge_utils.c
emanuele-f: The android code operates in the internal functions of n2n so I think it's not very easy to decouple it, just look at all the `ifdef __ANDROID_NDK__` in edge_utils.c . How do you suggest to make this decoupling?

lucaderi: Move this code to a separate file such as in tuntap for instance, and add empty hooks for platforms that do not need them. I would like to keep the code easy to read instead of having too many #ifdefs

emanuele-f: Are you suggesting to implement a callback api in n2n so an external code (e.g. the android code) can hook specific events (e.g. supernode registration updated?). Again, I find it difficult to turn some of this android specific code into api hooks, e.g. how would you name the hook at https://github.com/ntop/n2n/blob/81a3102661b8071b70686c9194740d2fbf40aaf1/edge_utils.c#L1116 ? Otherwise we can just define some android specific functions and call them in edge_utils.c under `#ifdef __ANDROID_NDK__` (e.g. `n2n_android_check_gateway` in the example above)

speck.c
lucaderi: Please move this android logic on a separate file. You can create empty hooks for platforms that do not need them, but I would like to see clean code. You can rework n2n in C++ if ti can help.

emanuele-f: Please see https://github.com/ntop/n2n/pull/259#discussion_r439337180 . Based on your comments it seems like the whole n2n codebase needs a rework, why do you want the rework to be part of this PR? Can this be just merged so that the android support is fixed and further work discussed separately?

lucaderi:

edge_utils.c
emanuele-f: speck support for android added in f5804db2b055a95c9099f1f809045a3107439477

lucaderi:
Thank you: now it looks better

PR #266, opened 2020-06-17
shortened help text by two lines

Logan007:
This pull request shortens the help text by two lines. As the header encryption will add another cli option, this is an attempt to maintain clarity.


lucaderi:

PR #267, opened 2020-06-18
ifdef'ed openSSL-includes for use with AES present only

### LINKED ISSUE #252: Bug in n2n.h file.
error  line  :
#include <openssl/opensslv.h>
#include <openssl/crypto.h>

The correct ones are as follows:

#ifdef  N2N_HAVE_AES
#include <openssl/opensslv.h>
#include <openssl/crypto.h>
#endif

Logan007:
All credits to @skyformat99.

Compiles well in Linux with openSSL 1.1 present. Not tested with openSSL 1.0 (on Mac?).

Fixes #252.


skyformat99:
thanks ^_^

lucaderi:

PR #269, opened 2020-06-21
Header Encryption

### LINKED ISSUE #198: Header Format

### LINKED ISSUE #257: It is recommended to add key option for supernode
Add a key option for the supernode, use the same cipher suite as the edge for authentication, and encrypt the virtual MAC address, virtual IP, and community name to avoid clear text display. Also added a key option for supernode authentication on the edge.

Logan007:
This pull request adds optional Header Encryption support to supernode and edges. It is enabled at the edges just by using the newly added `-H` command line option.

As the community name serves as key, there is no need for an additional key parameter.

The supernode can handle encrypted communities only if these are predetermined in a file that lists the allowed communities, see `-l <path>` parameter of supernode. This is because the fully encrypted packet header does not contain a hint of any kind to identify the correct key/community, so the supernode needs a limited set of communities to test for.

But the communities listed for the supernode do not have to be encypted, they could remain unencrypted. The decision for or against header encryption is a per-community decision, there are no mixed communities. So, all edges in this community have to be started with  `-H` or without it.

The supernode determines if a community is encrypted or not by evaluating the packets. If the first packet of some community is encrypted, it locks the community's state to encrypted and will accept and send out packets with encrypted headers only (as for this community). So, if a community was marked as encrypted, the supernode needs to be restarted to reset this setting.

The supernode can handle encrypted and unencrypted communities in parallel.

Also note, that re-use of community names for header encrypted communities once used for unencrypted communities is not recommended. Furthermore, if an edge was mistakenly started without the use of `-H` for intended header encryption, the community name (=key) spilled with some packets to the public. In this case, the choice of a new community name is recommended.

Fixes #198. 
Fixes #257.


lucaderi:

Issue #82, opened 2019-03-22
Message Authentication for Whole Packet


Logan007:
As stated in various other posts, message authentication is a feature well worth to consider.  Instead of leaving that to the respective transform-implementations, a whole-packet apporach should be considered taking also the unencrypted MACs into account.

That would require to hash the whole packet and sign the hash or  just encrypt that hash using secret pre-shared key derivate, e.g. the second half of the SHA512ed input key. This feature also requires to agree on one encryption method for signing/authenticating (TF, AES, somethin else...). 

In case of NULL transform, just hash without encryption.

emanuele-f:
Can you please recap the benefits of implementing message authentication over the current AES implementation? I'm reading that [malleability](https://en.wikipedia.org/wiki/Malleability_(cryptography)) could be a potential issue if the attacker manages to assemble multiple cyphertexts in a new one. However, after the changes in bb07f0426e6b83e8851307590620245a6950f80b, this should be more difficult as the cyphertext is now bound to the IV in use, so the modified parts of the cyphertext would be decoded to invalid values by the receiver.

Logan007:
A message authentication code (MAC, for this post) makes sure (as sure as it allows), that the message has not been altered.

As CBC decryption is not as chained as the encryption but could be considered more parallel, even a correctly decrypted last block with some padding length in it does not guarantee that the packet has no changes in its middle. It just means that the last cipher block and the cipher block before probably were not altered on their way.

Also, the packet could just be changed by appending more blocks as there is no length control. There is a chance of 15 / 256 that the last byte gets decrypted to a valid padding-length byte (1 through 15 make sense only). So far, the padding follows W3C padding scheme which does not allow for verification of the padding itself. Another scheme (ANSI, PCKS5, ...) with well defined padding pattern could lower these chances to something probably between 1 / 256 and 2 / 256 - which is still somewhat high. This would not be of any concern when using a MAC.

You are right, another packet that was created under a different IV would not properly decrypt nor would a randomly altered packet. It would result in garbage in the middle or the end of the payload. That is something an attacker might want to do to disturb the user. And it could be detected and prevented by using a MAC.

We could just use a MAC (minimum of 8 bytes recommended by government entities, better 12 or above, I could live with 4 for encryption scheme 1) for the whole packet (including outer and inner headers) for verification. Even padding scheme could remain the same. Every packet not matching its MAC needs to be discarded.

[...]

PR #279, opened 2020-06-29
Packet Checksum

Logan007:
This pull request adds a full packet checksum. For compatibility reasons, the checksum is only calculated and verified for communities having header encryption enabled (`-H` at the edge) as no dedicated field available otherwise. Failing checksum verification – which is performed at the edges _and_ the supernode (even for forwarded only packets) – results in immediately dropping the packet.

The checksum (16 bit) is blended into header encryption's IV (96 bit) by a quick 96-bit block cipher step. The checksum length is that low because we do not know how many more bits of IV will be required for replay prevention which is the next feature to go on. As soon as that is implemented, the checksum length will adopt, i.e. most probably be extended.

Solves #82.


lucaderi:
Thank you

PR #282, opened 2020-06-30
Add the callbacks API

emanuele-f:
Callbacks allow n2n API users to extend n2n without modifying the upstream code. This also allows us to move the hin2n code, which is specific of android, out of the n2n repository.


lucaderi:
The idea is good but I am wondering if moving to C++ can be a better solution

PR #284, opened 2020-06-30
edge.c: fix incorrect bad gateway check

disconnect3d:
It seems the previous `route.net_addr == INADDR_NONE` check was incorrect and was probably a result of copy-pasting the previous if statement and not changing the condition.


lucaderi:
Thank you

PR #285, opened 2020-06-30
call lzo_init() only if required

Logan007:
I encountered failing calls to `lzo_init()` on one of my older x86 machines and was not able to track it down. I never saw it before, not even on that machine. 

As of now, `lzo_init()` gets called no matter if lzo compression (`-z` or `-z1`) is enabled. So, this pull request makes sure that `lzo_init()` gets called only if lzo compression is enabled.


lucaderi:
Thank you

PR #286, opened 2020-07-01
Fix invalid sendto when supernode name resolution fails

emanuele-f:
Even though the supernode resolution failed, sendto was called with
and invalid socket, resulting in the following error:

ERROR: sendto failed (13) Permission denied


lucaderi:
Thank you

PR #289, opened 2020-07-01
Fix cmake compilation errors.

fengdaolong:
1. CMakeLists.txt added to automatically obtain the revised version number from the git repository.
2. Adjusted some header files to fix Windows platform compilation errors.
3. Eliminate some compilation warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,15 @@ cmake_minimum_required(VERSION 2.6)
 include(CheckFunctionExists)
 SET(CMAKE_VERBOSE_MAKEFILE ON)
 
-# N2n information
+# N2n release information
+set(N2N_VERSION "2.7.0")
+set(N2N_OSNAME ${CMAKE_SYSTEM})
+execute_process(
+        COMMAND git --version
+        WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
+        OUTPUT_VARIABLE GIT_EXIST
+)
+if (GIT_EXIST)
 execute_process(
         COMMAND git rev-list --count HEAD
         WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -51,51 +71,67 @@ endif()
 #message(${N2N_OPTION_AES})
 ## DEBUG FOR CMAKE
 
+INCLUDE_DIRECTORIES(.)
+INCLUDE_DIRECTORIES(include)
+if(DEFINED WIN32)
+  INCLUDE_DIRECTORIES(win32)
+  INCLUDE_DIRECTORIES("D:/Program Files/MinGW/opt/include/" "D:/Program Files/MinGW/x86_64-w64-mingw32/include/")
+  LINK_DIRECTORIES("D:/Program Files/MinGW/opt/lib/" "D:/Program Files/MinGW/x86_64-w64-mingw32/lib/")
```
lucaderi: I don't think it's a good idea to hardcode paths

lucaderi (changes requested):
Please remove hardcoded paths

CMakeLists.txt
fengdaolong: For Windows, the path of the header file and the library are uncertain. These two absolute paths need to be changed according to the actual situation at the time of construction. For the time being, there is no other better solution. If they are not coordinated here, can directly comment them.

emanuele-f:
This is not good https://github.com/ntop/n2n/pull/289/commits/f5accc8221f7b296dbdb053192e0225e80cea4dd#diff-af3b638bc2a3e6c650974192a53c7291R8 . git may not be installed or may not be in the search path in windows. 
Also this is not good https://github.com/ntop/n2n/pull/289/commits/f5accc8221f7b296dbdb053192e0225e80cea4dd#diff-af3b638bc2a3e6c650974192a53c7291R39 . You can compile n2n with AES support as explained in https://github.com/ntop/n2n/blob/dev/doc/Windows.md . Basically this PR is breaking the documented windows compilation and making it work only in your environment

fengdaolong:
Unchanged CMakeLists.txt, the binary version number compiled in the linux system also has problems.
Use autogen.sh script and makefile to build, also calls the git command to get the revised.

emanuele-f:
autogen.sh should not be called in windows as it only works under mingw . The version number indeed is a problem, maybe we can check if git is on the search path before running the git command, otherwise fall back to the hard-coded version number

lucaderi:
Thanks

PR #290, opened 2020-07-01
Move android code to hin2n

emanuele-f:


lucaderi:
What about removing all the Android code al all if you moved this port elsewhere?

emanuele-f:
Let's recap the current hin2n structure:

- the hin2n repository contains the android java source and the NDK code needed to glue n2n and the VPNService.
- hin2n requires the original n2n source in order to compile, which is located *outside* hin2n. Currently it points to https://github.com/switch-iot/n2n_ntop/tree/ntop-dev which is a 1:1 fork of this repository (ntop/n2n), so they are interchangeable.

Keeping the android and the n2n code separate (which occurred far before this PR was approved since the hin2n development continued on the hin2n repository) is logic as they are two big projects. The same choice as been taken by syncthing (https://github.com/syncthing/syncthing-android) within the same organization. In essence hin2n is not a fork of n2n, it uses n2n as a dependency.

So if you support hin2n as an official n2n project I think it is good to keep references to the android code in n2n (or to refactor them if possible) as they provide compatibility.

switch-st:
I will  try to remove android code gradually.

#306 
This pull request is remove part of android arp code.

#315 
This pull request is remove android define and log code.

And I'll remove other arp code soon. In this process, additional callbacks may need to be added. Hope your support.

Thank you.

Issue #280, opened 2020-06-29
payload with REGISTER?


Logan007:
While trying to understand the packet exchange which still is confusing to me, I came across the following code in the `sn.c` file around line 660 that seems to handle REGISTER packets from edges in the supernode:

```
        /* Copy the original payload unchanged */
        encode_buf(encbuf, &encx, (udp_buf + idx), (udp_size - idx));
```

In the code, I did not find any other reference to REGISTER datagrams that would carry additional payload. According to my understanding, the socket information is already decoded inside `decode_REGISTER()`. So, I installed an  _alarm bell_ 

`if( (udp_size - idx) ) fprintf (stdout, "!!!\n");`

which did not ring a single time in my admittedly somewhat limited setup. Can we safely delete the above mentioned call to `encode_buf()` and save some CPU cycles? If so, I gladly will bring up a corresponding pull request. If not, I will be deeply embarrassed and will study the code much harder… :wink:

What did I miss?

PR #294, opened 2020-07-02
no payload with REGISTER

Logan007:
This pull request removes a call to `encode_buf()` for payload that once might have been expected with REGISTER but was never observed – neither in assembling REGISTER datagrams nor on-line.

Solves #280.


lucaderi:
Thanks

PR #297, opened 2020-07-03
Change n2n_edge_callbacks payload_size parameter to pointer.

switch-st:
Change n2n_edge_callbacks payload_size parameter to pointer. So the packet payload can be modified.


emanuele-f:
Please note that this only allows the packet size to be reduced.

PR #300, opened 2020-07-04
enabled packet checksum

### LINKED ISSUE #295: observed packet drop for second community with encrypted headers in sn
will investigate

Logan007:
This pull request (re-)enables packet checksum again.

Turns out that there was nothing wrong with the checksum, just with author's network setup :facepalm: 

If you want to test just make sure that _all_ network edges _and_ the supernode get updated to that version with packet checksum feature and not just one…

Fixes #295.


lucaderi:
Thanks

PR #303, opened 2020-07-05
Add support for switched MAC addresses

emanuele-f:
This patch allows n2n to properly route MAC addresses located behind the edge nodes. The map "sw_macs" maps such "switched" MAC addresses to the edge node to use. When sending a packet, the destination MAC is first searched in sw_macs. If a match is found, the destination MAC reported into the N2N packet metadata is replaced with the edge node MAC. This allows a supernode, which has no knowledge of the switched MAC, to properly route the packet to the destination edge.

### DIFF
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -65,9 +65,9 @@
 
 #define DEFAULT_MTU   1290
 
-#define HASH_ADD_PEER(head,add)				\
+#define HASH_ADD_MAC(head,add)				\
   HASH_ADD(hh,head,mac_addr,sizeof(n2n_mac_t),add)
-#define HASH_FIND_PEER(head,mac,out)		\
+#define HASH_FIND_MAC(head,mac,out)		\
   HASH_FIND(hh,head,mac,sizeof(n2n_mac_t),out)
 #define N2N_EDGE_SN_HOST_SIZE   48
 #define N2N_EDGE_NUM_SUPERNODES 2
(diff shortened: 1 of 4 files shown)


lucaderi:
The main issue of this PR is that you want to implement bridging by in essence propagating the Mac address (that can be thousand) by creating a large/huge broadcast collision domain. This is not a good idea IMHO. You put pressure on all edges as well the supernode. Much better to use a different approach that allows you to bridge networks by encapsulating MACs n ad different protocol (see for instance VXLAN or other protocols) than to use the approach you have implemented that is the simplest considering the existing code base. Due to this, I am not in favour of accepting this contribution as is.

emanuele-f:
This pr allows n2n to work at l2 as a user would expect. Since n2n is a l2 VPN, this pr could actually be considered a *fix* as the ability to switch packets at l2 is currently broken. This pr does not force or suggests bad practices like creating huge broadcast domains, the user is responsible for the network topology. Regarding the effect on the nodes, this pr does not affect the supernode at all. If we really want to reason about the supernode, the fewer devices register with it, the fewer memory is needed so this pr can only improve the situation as the switched mac reuse the existing edge connection. The only effect of this pr is on the edge nodes, as the additional lookup on the switched macs hash table is an extra cost to pay. Please consider that providing proper l2 support in n2n opens a lot of new connectivity possibilities, like the ability to bridge n2n to an interface with no ip address as in a virtual bridge

skyformat99:
This should be useful when Edge is added to the bridge.

lucaderi:
My argument is to promote bridging in a clean way. I know this PR does not break anything but it potentially paves the way to breaking everything. If you want to implement bridging, then we need to do it properly. Adding entropy on existing code won't help. Bottom line: I am open to implement bridging if properly implemented, not the way this PR implements it.

lucaderi:
Closed for inactivity. Will reopen if necessary

Zouuup:
To me, this looks like the best possible implementation for the task at hand, maybe it's best to leave concerns over putting lots of pressure on super/edge nodes to end-users with a warning, the way I understood it's just a matter of having very large arp tables, which I think is pretty much manageable and expected on large VPN platforms...

PR #305, opened 2020-07-07
* fix -> Fix read tap device failed when OS wake up from sleep

skyformat99:
1)Fix compilation errors on windows
2)fix variable struct name
3)add sleep()

### DIFF
--- a/src/edge.c
+++ b/src/edge.c
@@ -273,7 +273,7 @@ static void setPayloadEncryption( n2n_edge_conf_t *conf, int cipher) {
 
 /* *************************************************** */
 
-static int setOption(int optkey, char *optargument, n2n_priv_config_t *ec, n2n_edge_conf_t *conf) {
+static int setOption(int optkey, char *optargument, n2n_tuntap_priv_config_t *ec, n2n_edge_conf_t *conf) {
   /* traceEvent(TRACE_NORMAL, "Option %c = %s", optkey, optargument ? optargument : ""); */
 
   switch(optkey) {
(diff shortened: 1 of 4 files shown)


src/edge_utils.c
```diff
@@ -2389,6 +2388,35 @@ static int edge_init_routes(n2n_edge_t *eee, n2n_route_t *routes, uint16_t num_r
   }
 #endif
 
+#ifdef WIN32
```
lucaderi, line 2391: Can you please cleanup this code so that each platform has a clean way of injecting rules?

src/edge.c
```diff
@@ -182,8 +182,8 @@ static void help() {
   printf("-S                       | Do not connect P2P. Always use the supernode.\n");
 #ifdef __linux__
   printf("-T <tos>                 | TOS for packets (e.g. 0x48 for SSH like priority)\n");
-  printf("-n <cidr:gateway>        | Route an IPv4 network via the gw. Use 0.0.0.0/0 for the default gw. Can be set multiple times.\n");
 #endif
+  printf("-n <cidr:gateway>        | Route an IPv4 network via the gw. Use 0.0.0.0/0 for the default gw. Can be set multiple times.\n");
```
lucaderi, line 186: This does not seem to be available for all platforms. In order to accept your patch your work must be split in two pieces: 
- add rounting
- the fix from sleep. Please submit two pull requests

lucaderi (changes requested):

skyformat99:
A new request will be resubmitted in addition, and this will be closed.

PR #306, opened 2020-07-07
move android arp code to n2n_edge_callbacks

switch-st:
1. Move android arp code to n2n_edge_callbacks. This will reduce the complexity of n2n code.
2. Add extra comment to n2n_edge_callbacks.


lucaderi:
thanks for the cleanup

PR #309, opened 2020-07-17
add sleep() for fix compile error on windows

skyformat99:
add sleep() for fix compile error on windows.

### DIFF
--- a/include/edge_utils_win32.h
+++ b/include/edge_utils_win32.h
@@ -21,8 +21,11 @@
 
 #ifdef WIN32
 
+#define WIN32_LEAN_AND_MEAN
+
 #include <process.h>
 #include <n2n.h>
+
 
 /* Multicast peers discovery disabled due to https://github.com/ntop/n2n/issues/65 */
 #define SKIP_MULTICAST_PEERS_DISCOVERY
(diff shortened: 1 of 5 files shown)


include/edge_utils_win32.h
```diff
@@ -21,9 +21,12 @@
 
 #ifdef WIN32
 
+#define WIN32_LEAN_AND_MEAN
```
lucaderi, line 24: What is this define? Why is necessary?

skyformat99: ref1:   https://issues.apache.org/jira/browse/THRIFT-2150
When both Windows.h and Winsock2.h are included, it is easy to cause compilation conflicts.
ref2:  https://msdn.microsoft.com/en-us/library/windows/desktop/aa383745(v=vs.85).aspx
   the explanation at the end of the article.

lucaderi:
Thanks

fengdaolong:
If it must be "#define WIN32_LEAN_AND_MEAN", why not do it like below.

#define WIN32_LEAN_AND_MEAN
#include <windows.h>
#undef WIN32_LEAN_AND_MEAN

PR #313, opened 2020-07-18
Clean up redundant code in sn.c

fengdaolong:
Clean up redundant code contained in both sn.c and sn_utils.c


Logan007:
I have always wondered why it has been doubled, but never dared to question it… :smile:

fengdaolong:
> I have always wondered why it has been doubled, but never dared to question it… 😄

After cleaning up, reading will be much more refreshing.😂😂😂😂😂😂😂

lucaderi:
Thanks

PR #315, opened 2020-07-20
delete android define and log code

switch-st:
- delete android define and log code

- close trace file if needed

Close trace file is needed by hin2n. Hin2n use `traceFile` to write logs. And Need to close trace file when try to stop app.


lucaderi:
Thanks

PR #318, opened 2020-07-21
add main loop callback and move android arp code to this callback

switch-st:
- add main loop callback
- move last android arp code to this callback


lucaderi:
Thanks: this further cleanups the code

Issue #281, opened 2020-06-30
Is there a system to manage the N2N network?


jonahzheng:
I want to monitor the status of all connections edge.

Logan007:
The edges as well as the supernode provide basic information via their respective localhost udp management port. They default to 5644 and 5645.

You can request that information by just sending a new line (press enter) under running the following command (localhost only):

`netcat -u  localhost 5644` (edge) or
`netcat -u  localhost 5645`(supernode).

The edge even understands some simple commands, try `help` instead of just enter. 

Even though the supernode lists the currently supported and/or active communities, it only lists the total number of edges. I like the idea of listing the connected edges by their MACs under the communities. It would require an additional loop inside [this](https://github.com/ntop/n2n/blob/dev/src/sn.c#L346) loop. Are you able to add that code and want to provide a pull request?

jonahzheng:
thank you, I'll try

Logan007:
@jonahzheng I did not see your reply... I already did it. Not beautiful but parsable.

tlsalex:
what is the package name for netcat , I run :
```
root@xxx yum.repos.d]# yum provides */netcat
Loaded plugins: langpacks, versionlock
Excluding 1 update due to versionlock (use "yum versionlock status" to show it)
RackTables-0.20.14-1.el7.noarch : A data-center asset management system
Repo        : epel
Matched from:
Filename    : /usr/share/RackTables/gateways/netcat
```
Should not this RackTables

tlsalex:
I guess it is this package: nmap-ncat
after the installation, I issue:

```
[root@bdc ~]# ss -uln
State      Recv-Q Send-Q                                 Local Address:Port                                                Peer Address:Port
UNCONN     0      0                                                  *:49152                                                          *:*
UNCONN     0      0                                                  *:49152                                                          *:*
UNCONN     0      0                                          127.0.0.1:5645                                                           *:*
UNCONN     0      0                                          127.0.0.1:5645                                                           *:*
UNCONN     0      0                                          127.0.0.1:2099                                                           *:*
UNCONN     0      0                                                  *:68                                                             *:*
UNCONN     0      0                                          127.0.0.1:323                                                            *:*
UNCONN     0      0                                              [::1]:323                                                         [::]:*
[root@bdc ~]# ncat -u localhost 5645
```
The cursor get stuck there.
it returns me nothing.
Did I do something wrong ?

Logan007:
Just send a new line by pressing the Enter key.

[...]

PR #320, opened 2020-07-21
fix supernode coredump when accessing management socket

switch-st:
When accessing the supernode management port, it is possible that the `resbuf` array is out of bounds due to too many communities and edges.

Use `SAFE_SNPRINTF ` to protect `resbuf`.

Or increasing the length of the `resbuf` array to solve this problem.

### DIFF
--- a/src/sn_utils.c
+++ b/src/sn_utils.c
@@ -388,7 +388,7 @@ static int sendto_mgmt(n2n_sn_t *sss,
 
   if (r <= 0) {
     ++(sss->stats.errors);
-    traceEvent (TRACE_ERROR, "process_mgmt : sendto failed. %s", strerror (errno));
+    traceEvent (TRACE_ERROR, "sendto_mgmt : sendto failed. %s", strerror (errno));
     return -1;
   }
   return 0;
--- a/src/sn_utils.c
+++ b/src/sn_utils.c
@@ -15,6 +15,11 @@ static ssize_t sendto_sock(n2n_sn_t *sss,
                            const uint8_t *pktbuf,
                            size_t pktsize);
 
+static int sendto_mgmt(n2n_sn_t *sss,
+                       const struct sockaddr_in *sender_sock,
+                       const uint8_t *mgmt_buf,
+                       size_t mgmt_size);
+
 static int try_broadcast(n2n_sn_t * sss,
 		         const struct sn_community *comm,
 			 const n2n_common_t * cmn,
@@ -295,10 +300,11 @@ static int process_mgmt(n2n_sn_t *sss,
     char resbuf[N2N_SN_PKTBUF_SIZE];
     size_t ressize = 0;
     uint32_t num_edges = 0;
-    ssize_t r;
+    uint32_t num = 0;
     struct sn_community *community, *tmp;
     struct peer_info * peer, *tmpPeer;
     macstr_t mac_buf;
+    n2n_sock_str_t sockbuf;
 
     traceEvent(TRACE_DEBUG, "process_mgmt");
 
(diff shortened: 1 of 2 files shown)


fengdaolong:
The original ressize has been protected, the code here does not need to be changed.

switch-st:
It will crash on linux.

But is it necessary to print communities and edges information here? If it's necessary, may be we can use dump_registrations function here.

fengdaolong:
It is necessary to print edge information, including: MAC, external network IP port, tun IP.

fengdaolong:
If the data sent by the socket here exceeds the maximum MTU size, consider whether it needs to be sent in batches.

Logan007:
My 2 [cents](https://github.com/ntop/n2n/issues/281#issuecomment-662022072). In addition, in case we are able to keep the detailed output, it would be great if we could keep the complete information and would not need to cut it off. So, I would be in for batches.

Logan007:
Would it help to insert a copy the _send-to-socket_ block of `sn_utils.c:363-370` just along with a `ressize = 0;` and parenthesized in an `if(ressize >= N2N_SN_PKTBUF_SIZE - max_output_item_len) {...}` statement at line 358?

switch-st:
> Would it help to insert a copy the _send-to-socket_ block of `sn_utils.c:363-370` just along with a `ressize = 0;` and parenthesized in an `if(ressize >= N2N_SN_PKTBUF_SIZE - max_output_item_len) {...}` statement at line 358?

@Logan007 This solution will also cause the array to go out of bounds, When the number of edges in each community exceeds 100.

How about this? Send to socket once for every edges?
``` c
static int process_mgmt(n2n_sn_t *sss,
                        const struct sockaddr_in *sender_sock,
                        const uint8_t *mgmt_buf,
                        size_t mgmt_size,
                        time_t now)
{
    .........

    ressize += snprintf(resbuf+ressize, N2N_SN_PKTBUF_SIZE-ressize,
                        "cur_cmnts %u\n", HASH_COUNT(sss->communities));
    HASH_ITER(hh, sss->communities, community, tmp) {
      ressize += snprintf(resbuf + ressize, N2N_SN_PKTBUF_SIZE - ressize,
                          "community: %s\n", community->community);
      sendto_mgmt(sss, sender_sock, resbuf, ressize);
      ressize = 0;

      num = 0;
      HASH_ITER(hh, community->edges, peer, tmpPeer) {
        if(peer->sock.family == AF_INET)
        ressize += snprintf(resbuf + ressize, N2N_SN_PKTBUF_SIZE - ressize,
                            "\t[id: %u][MAC: %s][edge: %s][last seen: %lu sec ago]\n",
                            ++num, macaddr_str(mac_buf, peer->mac_addr),
                            sock_to_cstr(sockbuf, &(peer->sock)), now-peer->last_seen);

        sendto_mgmt(sss, sender_sock, resbuf, ressize);
        ressize = 0;
      }
    }

    ressize += snprintf(resbuf+ressize, N2N_SN_PKTBUF_SIZE-ressize,
                        "\n");
    sendto_mgmt(sss, sender_sock, resbuf, ressize);

    return 0;
}
```

``` c
static int sendto_mgmt (n2n_sn_t *sss,
						const struct sockaddr_in *sender_sock,
						const uint8_t *mgmt_buf,
						size_t mgmt_size)
{
  ssize_t r = sendto (sss->mgmt_sock, mgmt_buf, mgmt_size, 0 /*flags*/,
			  (struct sockaddr *)sender_sock, sizeof (struct sockaddr_in));

  if (r <= 0) {
    ++(sss->stats.errors);
    traceEvent (TRACE_ERROR, "process_mgmt : sendto failed. %s", strerror (errno));
	  return -1;
  }
  return 0;
}
```

I just copy the code of function `dump_registrations `. This output format may be clearer.

image

fengdaolong:
The sock_to_cstr function can be used directly to resolve the IP address of the peer.
`sock_to_cstr(sock_buf, &peer->sock));`

switch-st:
> The sock_to_cstr function can be used directly to resolve the IP address of the peer.
> `sock_to_cstr(sock_buf, &peer->sock));`

Great! :thumbsup:

Logan007:
That is basically what I was suggesting… but just `sendto_mgmt()` if really necessary, i.e. not enough space for the next output (item or line) left in the array. But yeah, on the other hand, I guess that serving the management socket once in a while will not harm performance too much anyway.

Your idea of using `dump_registrations()` is terrific and the output looks so much more readable. You should commit your changes! :+1:

switch-st:
> That is basically what I was suggesting… but just `sendto_mgmt()` if really necessary, i.e. not enough space for the next output (item or line) left in the array. But yeah, on the other hand, I guess that serving the management socket once in a while will not harm performance too much anyway.
> 
> Your idea of using `dump_registrations()` is terrific and the output looks so much more readable. You should commit your changes! 👍

OK, I'll update this PR.

switch-st:
Pull request has been updated.

src/sn_utils.c
```diff
@@ -346,30 +352,46 @@ static int process_mgmt(n2n_sn_t *sss,
...
+                       const uint8_t *mgmt_buf,
+                       size_t mgmt_size)
+{
+  ssize_t r = sendto(sss->mgmt_sock, mgmt_buf, mgmt_size, 0 /*flags*/,
+                     (struct sockaddr *)sender_sock, sizeof (struct sockaddr_in));
 
-    return 0;
+  if (r <= 0) {
+    ++(sss->stats.errors);
+    traceEvent (TRACE_ERROR, "process_mgmt : sendto failed. %s", strerror (errno));
```
Logan007: An absolutely minor comment (of which I am about to be sorry to even make it :smile:) : With the new `sendto_mgmt()` in place, we could replace trace output `process_mgmt…` with `sendto_mgmt…`.

Logan007 (changes requested):
Looks good!

src/sn_utils.c
switch-st: @Logan007 It's my careless. Thank you. :wink:

fengdaolong: Logan007 is a careful person.😄

Logan007: With all the diverse talents assembled here, we will push it forward! United we are strong! :smile: :muscle:

Logan007 (approved):
Looks better :)

(practicing github's review functionality)

lucaderi:
Thanks

PR #321, opened 2020-07-22
fix build error for supernode when declared N2N_HAVE_DAEMON

switch-st:
Variable `useSyslog ` is not visible in sn.c. Use `setUseSyslog()` instead of `useSyslog`.


lucaderi:
Thanks

PR #322, opened 2020-07-22
Update purge logic of communities and edges for supernode

switch-st:
There are two problems about the purge logic of supernode.

1. It will access the purge code every time, even though the trigger time is not reached.

2. It will update `last_purge_edges` after calling `purge_expired_registrations` function for the first community in the loop , when the trigger time is reached. Then `last_purge_edges` has updated, and `purge_expired_registrations` function will not work for other communities. In fact, the code here is only work for the first community.

### DIFF
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -35,6 +35,9 @@
 #define IFACE_UPDATE_INTERVAL           (30) /* sec. How long it usually takes to get an IP lease. */
 #define TRANSOP_TICK_INTERVAL           (10) /* sec */
 
+#define PURGE_REGISTRATION_FREQUENCY   30
+#define REGISTRATION_TIMEOUT           60
+
 #define ETH_FRAMESIZE 14
 #define IP4_SRCOFFSET 12
 #define IP4_DSTOFFSET 16
--- a/src/n2n.c
+++ b/src/n2n.c
@@ -22,9 +22,6 @@
 
 #include <assert.h>
 
-#define PURGE_REGISTRATION_FREQUENCY   30
-#define REGISTRATION_TIMEOUT           60
-
 static const uint8_t broadcast_addr[6] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF };
 static const uint8_t multicast_addr[6] = { 0x01, 0x00, 0x5E, 0x00, 0x00, 0x00 }; /* First 3 bytes are meaningful */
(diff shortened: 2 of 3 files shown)


Logan007:
Touching that portion of the code brings up a question that I have already pondered for some time:

Do you think it would be feasible to sort the communities' hash list once in a while, e.g. directly at the end of purging using `HASH_SORT`? Sorting in terms of packet-count mostly used communities upfront could help the supernode performing better during handling incoming packets.

fengdaolong:
> Touching that portion of the code brings up a question that I have already pondered for some time:
> 
> Do you think it would be feasible to sort the communities' hash list once in a while, e.g. directly at the end of purging using `HASH_SORT`? Sorting in terms of packet-count mostly used communities upfront could help the supernode performing better during handling incoming packets.

Your ideas are always useful❤️

Logan007:
Aaawwww…

switch-st:
> Touching that portion of the code brings up a question that I have already pondered for some time:
> 
> Do you think it would be feasible to sort the communities' hash list once in a while, e.g. directly at the end of purging using `HASH_SORT`? Sorting in terms of packet-count mostly used communities upfront could help the supernode performing better during handling incoming packets.

I think it is effective for encrypted header communities. 

In order to prevent frequent shaking of communities hash table, may be we should test the call interval of `HASH_SORT`.

Logan007:
> I think it is effective for encrypted header communities.

That is exactly what I had in mind :smile:

> In order to prevent frequent shaking of communities hash table, may be we should test the call interval of HASH_SORT.

That is a very good idea! 

Time or counter? 

Shall we handle it inside this pull request or make it a new one?

switch-st:
> > I think it is effective for encrypted header communities.
> 
> That is exactly what I had in mind 😄
> 
> > In order to prevent frequent shaking of communities hash table, may be we should test the call interval of HASH_SORT.
> 
> That is a very good idea!
> 
> Time or counter?
> 
> Shall we handle it inside this pull request or make it a new one?

May be we should make it a new pull request.

Logan007:
First, I want to finish replay protection, will have it done by mid of next week. Then I might find some time to do this. But I have no objections if you want to do it.

switch-st:
> First, I want to finish replay protection, will have it done by mid of next week. Then I might find some time to do this. But I have no objections if you want to do it.

I can do some test work when you finish this `HASH_SORT ` feature. :grin:

lucaderi:
Thanks

PR #324, opened 2020-07-23
fix build warning of edge_init_routes

switch-st:
Compiling on non-linux and non-windows will trigger a warning on `edge_init_routes` function.

Just like this:

`src/edge_utils.c:2422:1: warning: control reaches end of non-void function [-Wreturn-type]`


lucaderi:
Thanks

PR #325, opened 2020-07-24
add management port command option to supernode

switch-st:
Add management port command option to supernode.  This feature is useful for running multiple supernodes on a machine.


lucaderi:
Thanks

PR #329, opened 2020-07-27
Adds a trace message to mark connection type between edges.

skyformat99:
Adds a trace message to mark connection type between edges.


Logan007:
I especially like your way of indicating and differentiating the different connection types `[P2P]` vs. `[PsP]`!

Are there more places in the code where this could be applied to the trace output?

lucaderi:
Thanks

PR #330, opened 2020-07-27
added crypto doc

Logan007:
For review and further consideration.

Will most probably contain some typos.


lucaderi:
Documentation is welcome! Thanks

PR #331, opened 2020-07-27
retrying syscall and calls to hardware random number generators

Logan007:
This pull request addresses an issue which was brought forward to the n2n team in private conversation:

The syscall to `getrandom()` and also Intel's hardware number generators `RDRND` and `RDSEED` could fail. In case of failure, they now get another chance(s) by retrying for a given number of times.

### DIFF
--- a/src/random_numbers.c
+++ b/src/random_numbers.c
@@ -16,7 +16,15 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


lucaderi:
Thanks

PR #332, opened 2020-07-27
sorting encrypted communities

Logan007:
This pull request adds a periodical sorting of supernode's communities' hash list by descending order of number of encrypted packets sent during the last period.

This will have the most often used encrypted community iterated first by HASH_ITER and thus especially help to find the community key faster in most cases (which is found by iterating through the communities and trying their hash of name as key).

Period length is defined as `SORT_COMMUNITIES_INTERVAL` in `n2n_define.h` file and defaults to 90 seconds.


lucaderi:
Thanks

PR #335, opened 2020-07-29
Indicates successful connection between the edge and SN nodes

skyformat99:
Indicates successful connection between the edge and SN nodes.


lucaderi:
Thanks

PR #337, opened 2020-07-29
moved some #DEFs to central n2n_define.h

Logan007:


Logan007:
...and also removed two duplicates.

fengdaolong:
I like people who do things rigorously.❤

Logan007:
Yeah, I had to work up all courage to do it.

PR #342, opened 2020-07-31
For supernode, the regex filter rule is added to the -c option

skyformat99:
For private supernode,  the -c option can be added for security reasons.
However, only full-word matches are allowed.
This patch adds regular expression filtering rules that are very convenient and previously compatible.

for example:
# supernode  -c config.ini  -l 7788   -v

The contents of the file config.ini are as follows:
abc[0-9]
eee*

Then the connection of the node is as follows: (xx.xx.xx.xx : supernode ip)
# edge -a 10.1.1.10 -c abc0    -k 00    -l xx.xx.xx.xx:7788        ===>   Connection successful
# edge -a 10.1.1.11 -c abcxx  -k zz   -l xx.xx.xx.xx:7788          ===>   Connection refused,  does not match  abc[0-9]
# edge -a 10.1.1.10 -c eeeabc    -k 00    -l xx.xx.xx.xx:7788     ===>   Connection successful

### DIFF
--- a/src/random_numbers.c
+++ b/src/random_numbers.c
@@ -98,9 +98,9 @@ uint64_t n2n_rand () {
    state yet, a call to   n2n_srand ( n2n_seed() )   would do. */
 uint64_t n2n_seed (void) {
 
-  uint64_t seed;
-  uint64_t ret;
-  size_t i;
+  uint64_t seed=0;
+  uint64_t ret=0;
+  size_t i=0;
 
 #ifdef SYS_getrandom
   int rc = -1;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ add_library(n2n STATIC
         src/tuntap_netbsd.c
         src/tuntap_linux.c
         src/tuntap_osx.c
+        src/n2n_regex.c
         )
 
 if(DEFINED WIN32)
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -358,6 +358,13 @@ struct sn_community
   UT_hash_handle hh; /* makes this structure hashable */
 };
 
+struct sn_community_filter
+{
+  char community_rule[N2N_COMMUNITY_SIZE];  // Filtered regular expressions
+
+  UT_hash_handle hh; /* makes this structure hashable */
+};
+
 typedef struct n2n_sn
 {
     time_t start_time; /* Used to measure uptime. */
@@ -373,6 +380,7 @@ typedef struct n2n_sn
 #endif
     int lock_communities; /* If true, only loaded communities can be used. */
     struct sn_community *communities;
+    struct sn_community_filter *communities_filters; // Filtered regular expressions list
 } n2n_sn_t;
 
 /* ************************************** */
--- a/include/n2n_regex.h
+++ b/include/n2n_regex.h
@@ -1,0 +1,59 @@
+/*
+ *
+ * Mini regex-module inspired by Rob Pike's regex code described in:
+ *
+ * http://www.cs.princeton.edu/courses/archive/spr09/cos333/beautiful.html
+ *
+ *
+ *
+ * Supports:
+ * ---------
+ *   '.'        Dot, matches any character
+ *   '^'        Start anchor, matches beginning of string
+ *   '$'        End anchor, matches end of string
+ *   '*'        Asterisk, match zero or more (greedy)
+ *   '+'        Plus, match one or more (greedy)
+ *   '?'        Question, match zero or one (non-greedy)
+ *   '[abc]'    Character class, match if one of {'a', 'b', 'c'}
+ *   '[^abc]'   Inverted class, match if NOT one of {'a', 'b', 'c'} -- NOTE: feature is currently broken!
+ *   '[a-zA-Z]' Character ranges, the character set of the ranges { a-z | A-Z }
+ *   '\s'       Whitespace, \t \f \r \n \v and spaces
+ *   '\S'       Non-whitespace
+ *   '\w'       Alphanumeric, [a-zA-Z0-9_]
+ *   '\W'       Non-alphanumeric
+ *   '\d'       Digits, [0-9]
+ *   '\D'       Non-digits
+ *
+ *
+ */
+
+#ifndef _N2N_REGEX_
+#define _N2N_REGEX_
+
+#ifdef __cplusplus
+extern "C"{
+#endif
+
+#include <stdio.h>
+
+/* Typedef'd pointer to get abstract datatype. */
+typedef struct regex_t* re_t;
+
+
+/* Compile regex string pattern to a regex_t-array. */
+re_t re_compile(const char* pattern);
+
+
+/* Find matches of the compiled pattern inside text. */
+int  re_matchp(re_t pattern, const char* text, int* matchlenght);
+
+
+/* Find matches of the txt pattern inside text (will compile automatically first). */
+int  re_match(const char* pattern, const char* text, int* matchlenght);
+
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
--- a/src/n2n_regex.c
+++ b/src/n2n_regex.c
@@ -1,0 +1,502 @@
+/*
+ *
+ * Mini regex-module inspired by Rob Pike's regex code described in:
+ *
+ * http://www.cs.princeton.edu/courses/archive/spr09/cos333/beautiful.html
+ *
+ *
+ *
+ * Supports:
+ * ---------
+ *   '.'        Dot, matches any character
+ *   '^'        Start anchor, matches beginning of string
+ *   '$'        End anchor, matches end of string
+ *   '*'        Asterisk, match zero or more (greedy)
+ *   '+'        Plus, match one or more (greedy)
+ *   '?'        Question, match zero or one (non-greedy)
+ *   '[abc]'    Character class, match if one of {'a', 'b', 'c'}
+ *   '[^abc]'   Inverted class, match if NOT one of {'a', 'b', 'c'} -- NOTE: feature is currently broken!
+ *   '[a-zA-Z]' Character ranges, the character set of the ranges { a-z | A-Z }
+ *   '\s'       Whitespace, \t \f \r \n \v and spaces
+ *   '\S'       Non-whitespace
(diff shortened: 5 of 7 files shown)


src/random_numbers.c
```diff
@@ -98,9 +98,9 @@ uint64_t n2n_rand () {
    state yet, a call to   n2n_srand ( n2n_seed() )   would do. */
 uint64_t n2n_seed (void) {
 
-  uint64_t seed;
-  uint64_t ret;
-  size_t i;
+  uint64_t seed=0;
+  uint64_t ret=0;
```
Logan007, line 102: If possible, please leave those two uninitialized. As they serve to gather some entropy and collect randomness for seeding, any previous arbitrary value from memory adds up to it. Does your compiler complain?

src/random_numbers.c
```diff
@@ -98,9 +98,9 @@ uint64_t n2n_rand () {
...
+  size_t i=0;
```
Logan007, line 103: This is not neccessarily required because `i` is initialized at the beginning of the loop. Does your compiler complain?

Logan007 (changes requested):
It is a great idea to optionally only accept new communities with a certain name pattern! I really like it. I also think you identified all the important points in the code where changes would be required.

As far as I can see it, the way it is implemented does however not allow for encrypted communities anymore.

So, we need to find a way that can let these features co-exist.

Logan007:
Maybe, we could keep the `-c` (loading the explicitly allowed **c**ommunities) and internally mark them as "locked" on a per community base (new field) in addition to the the global `sss->lock_communities` (which could be set by some additional  parameter, such as `-d` for _**d**o not accept any further communities_). Little adaptions to the purge algorithm would be required.

If `-d` is not set, additional communities can be added, internally marked as "unlocked". A warning would be required in case `-d` is specified with empty community list. 

This would also allow to use some communities with (optionally) encrypted headers (those loaded in `-c`) and additionally have new communities come in. Their names could undergo the pattern matching according to the regular expressions if present and loaded by some (also new) `-r` option.

A valid question would be how `-r` shall behave if a `-d` is present.

The good think about this approach is that it allows for basic syntax with or without `-c` – and not necessarily having the user care about the details of the more complicated, advanced options such as `-d` and `-r`.

What do you think?

src/random_numbers.c
skyformat99: Yes, when compiling with Vistudio Studio2015, it crashes in Debug mode.

src/random_numbers.c
skyformat99: sorry, You're right about the i variable.

skyformat99:
> Maybe, we could keep the `-c` (loading the explicitly allowed **c**ommunities) and internally mark them as "locked" on a per community base (new field) in addition to the the global `sss->lock_communities` (which could be set by some additional parameter, such as `-d` for _**d**o not accept any further communities_). Little adaptions to the purge algorithm would be required.
> 
> If `-d` is not set, additional communities can be added, internally marked as "unlocked". A warning would be required in case `-d` is specified with empty community list.
> 
> This would also allow to use some communities with (optionally) encrypted headers (those loaded in `-c`) and additionally have new communities come in. Their names could undergo the pattern matching according to the regular expressions if present and loaded by some (also new) `-r` option.
> 
> A valid question would be how `-r` shall behave if a `-d` is present.
> 
> The good think about this approach is that it allows for basic syntax with or without `-c` – and not necessarily having the user care about the details of the more complicated, advanced options such as `-d` and `-r`.
> 
> What do you think?


thanks for your reply.
About some communities with (optionally) encrypted headers,  I will confirm the result after the test.

skyformat99:
@Logan007 

About the use of  -H，Is there anything to note about the detailed usage？
Use the latest version of Git, and then compile and run：

PC1:    supernode   -l 7788  -v

PC2:    edge -a 10.1.1.10 -c abc -k 123  -l xx.xx.xx.xx:7788   -H
PC3:    edge -a 10.1.1.11 -c abc -k 123  -l xx.xx.xx.xx:7788   -H


However, there is no connection between the two edges.
What is wrong with that?

Logan007:
Should be as easy as using `-H`. Just make sure that
- the supernode is current, too
- **the community is listed in supernode's `-c <file>` list of allowed communities**
- the community name has not been used without `-H` before, not on a single edge (supernode decides on first packet of that community if encrypted or not and sticks to it); otherwise restart supernode

Description of internals can be found in [`docs/Crypto.md`](https://github.com/ntop/n2n/blob/dev/doc/CRYPTO.md).

skyformat99:
> Should be as easy as using `-H`. Just make sure that
> 
> * the supernode is current, too
> * **the community is listed in supernode's `-c <file>` list of allowed communities**
> * the community name has not been used without `-H` before, not on a single edge (supernode decides on first packet of that community if encrypted or not and sticks to it); otherwise restart supernode
> 
> Description of internals can be found in [`docs/Crypto.md`](https://github.com/ntop/n2n/blob/dev/doc/CRYPTO.md).

thanks,
It feels a little too restrictive.
If  can make use of the auth field, header encryption does not matter.
typedef struct n2n_REGISTER_SUPER
{
    n2n_cookie_t        cookie;         /* Link REGISTER_SUPER and REGISTER_SUPER_ACK */
    n2n_mac_t           edgeMac;        /* MAC to register with edge sending socket */
    n2n_auth_t          **auth**;           /* Authentication scheme and tokens */
} n2n_REGISTER_SUPER_t;

For example, JWT methond is used to transfer token.

Logan007:
The basic idea of header encryption is to fully disguise communication and make it undetectable for DPI. So, with full header encryption, there is no indicator left in the packet that would point to n2n as packet format. Not in a single one (well, but size, but we could append a random number of bytes… just a thought).. None of the packet formats is detectable by form or content of fields. Header encryption shall completely hide the meta-data (community names, MAC addresses, IP addresses of sockets, even tokens) from eavesdroppers. **Authentication** in this case is a side-effect only caused by shared knowledge of the same secret key. This feature can only work for community names known beforehand. Thus, they have to be provided via cli (`-c`) to the supernode.

It addresses the long standing [bullet](https://www.ntop.org/products/n2n/) _"n2n packets need to include some junk to avoid DPI applications to detect them"_ (n2n Security section).

I am aware that the way it is implemented, if enabled, header encryption does not allow for not-named-yet free communities anymore, because it uses the `-c` parameter. That is why I suggested to lock the communities fed by `-c` on a per-community basis and make the global locking a new distinct feature (`-d`?) which would just not allow any new community – only those fixed-name allowed by `-c` or by some `-r` filter. With that as a basis, regular expressions can be added as an additional parameter (`-r`?) without any need to worry about locked or unlocked communities.

**Another view** to regular expressions would be that some of them (those without any special characters such as `[].?`…) in fact already represent a non-changing community name. If we are able to identify those and mark them as "HEADER_ENCRYPTION_UNKNOWN" (the others get "HEADER_ENCRYPTION_DISABLED") but still must be locked on a per community basis (just like the regular `-c` feature does), we could have these two features co-exist, too. Plus, it would require no additional parameter. This alternative may seem not much more complicated to code but way more appealing (no extra cli-options).

What do you think?

lucaderi:
@skyformat99 @Logan007 I would like to plan 2.8 release this week. Please decide about this PR and then let me know if there is something else open to fix. Ideally non-bugs will need to go into the next release IMHO

Logan007:
In my opinion, there is a way to let these two features co-exist. I would prefer the second way lined out above. It would require some coding though. However, it is not a bug but a very welcomed new extra feature.

@skyformat99 Do you think we can finish it during the next say one or two days? We would have to hurry… :smile: Otherwise, we will make it part of post-release dev and later **2.10** or **3.0**.

skyformat99:
> In my opinion, there is a way to let these two features co-exist. I would prefer the second way lined out above. It would require some coding though. However, it is not a bug but a very welcomed new extra feature.
> 
> @skyformat99 Do you think we can finish it during the next say one or two days? We would have to hurry…  Otherwise, we will make it part of post-release dev and later **2.10** or **3.0**.


@Logan007
Yes, I do agree with you.  Can you complete all the code？ 
The code in this commit is free to use, and you can also submit new pull request.
When you are finished, I will close this pull request.
thanks.

Logan007:
Well, at least I could prepare the required changes inside (additional per-community locking, revise `HEADER_ENCRYPTION_*` flag handling if required).

I would need your help then for further steps to put regExp on top on that as I have not been deeply involved with those.

Okay, so let's give it a shot! I will start right away.

skyformat99:
In fact, non-regular expression filtering is also possible.
For example, filtering rules like MQTT also meet most requirements.
abc/+
abc/#
abc*

Maybe that's a good idea, too.

Logan007:
You mean like meta-filtering the expressions/names supplied in the `-c` file for regular expressions using regular expressions again? Close to mind-bending – that would be sooo cool!

What would the regular expression that filters regular names only look like? It should not only allow `a…z` and `A…Z` and `0…9` but also unusual characters (not used as special character inside regular expressions such as `é`) – so it should exclude only the regExp special characters. How would such a regular expression look like?

Logan007:
Working on it... might require a few more hours. No extra regular expression for distinguishing required, I will use `strpbrk()`. That will exlucde the `.^$*+?[]\` characters from regular community names. As this is a change in behaviour, this feature might be worth being delayed after the release. Not sure about that. However, I fill finish it, let's decide then.

@skyformat99 Did you write the `n2n_regex.c` code yourself? Or, where did you get it from – especially under what license? We would need to prepend it to the file.

skyformat99:
@Logan007 

ref:  https://github.com/kokke/tiny-regex-c

In addition, you can use -c file to make the difference.
The file format is as follows:

1 abcdfix              ===>     1    Is for fixed match
0 abc[0-9]             ====>  0   Regular expression matching

1 and 0 represent different types of matches,  Divide by space and the following name , Such a -c file can represent both.
What do you think?

Logan007:
Thank you. 

The good thing is: No difference required! That would also be too complicated to explain to users. The supernode will treat it either as fixed community name or as regular expression if detected. Only, that we loose some allowed characters for community name.

Logan007:
Please check out #349. What do you think?

skyformat99:
thanks，very good.
I will test it.

skyformat99:
see #349 
close it.

PR #345, opened 2020-08-02
revised readme and crypto docs

Logan007:
typos, corrections, …

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ In order to start using n2n, two elements are required:
 
 A virtual network shared between multiple edge nodes in n2n is called a _community_. A single supernode can relay multiple communities and a single computer can be part of multiple communities at the same time. An encryption key can be used by the edge nodes to encrypt the packets within their community.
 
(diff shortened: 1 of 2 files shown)


lucktu:
"In order to start using n2n, two elements are required:"?  I think so:
In order to start using n2n, three elements are required:
A _supernode_: it allows edge nodes to announce ...
B _edge_ nodes: the nodes which will ...
C _the third condition is that udp is allowed.

Logan007:
With the latest commit, I have pointed to udp in one of the paragraphs.

lucktu:
Well, that works, too.

Because some large companies or isps restrict or even prohibit udp traffic, n2n can not be used at this time.

lucaderi:
Thanks

Logan007:
P. S.  @lucktu In the meanwhile, as long as TCP remains unsupported, maybe [this](http://manpages.ubuntu.com/manpages/xenial/en/man1/udptunnel.1.html) kind of helper tools (and there surely are others around) might prove helpful.

PR #343, opened 2020-07-31
fix register supernode loop

switch-st:
see issue: https://github.com/ntop/n2n/issues/341

Issue #341, opened 2020-07-30
may not be correct that register to  supernode in function update_supernode_reg.


switch-st:
It's may not be correct that register to  supernode in function update_supernode_reg of line 934 to 944.

https://github.com/ntop/n2n/blob/2ea8dd53b7c5ed5390f3df42df3d92f43682c5c6/src/edge_utils.c#L934-L944

It will send register messages to both two supernodes. After that, `eee->supernode` will change to the last supernode address. Then edge will send messages to this address even if this address is not valid now. 

Also `send_register_super` function will change `eee->last_cookie`'s value every time. So call `send_register_super` function twice is not correct.

Just like this log:
![image](https://user-images.githubusercontent.com/7710981/88923204-eb746f00-d2a3-11ea-9eeb-96a2b7f82596.png)

test cmd: `./edge -a static:192.168.235.20  -c xxxx -k xxxx -l switchy.wang:10086 -l switchy.wang:9999 -f -A3 -vvv`

The second supernode `switchy.wang:9999` is invalid. Just first supernode `switchy.wang:10086` responded. But the cookie is not match with `eee->last_cookie` of edge.

I'm not sure about the purpose of why send register message to both supernode here. But i think that the logic here is incorrect.

Logan007:
I think I see what you are talking about. I have no idea why the supernodes are iterated using that `for` loop. Especially as they already get _slowly_ iterated and retried over time along the `eee->*` fields `sup_attempts` and `conf.sn_idx`.

Would it help to
- comment / delete lines 909, 935 as well as 943 and
- replace `sn_idx` by `eee->conf.sn_idx` in lines 936 and 936?

I guess that is what you have had in mind, right? Good catch!

skyformat99:
Every once in a while, the Edge sends a registration message to all the SuperNode nodes to maintain the Edge list.
An invalid SN node may become valid again at sometime in the future.
So It should send a registration message to all  SN nodes.

skyformat99:
For cookies,  can put the  last_cookie  value to the structure of  n2n_edge_conf:
typedef struct n2n_edge_conf {
  n2n_sn_name_t       sn_ip_array[N2N_EDGE_NUM_SUPERNODES];
 ###  **n2n_cookie_t        last_cookie[N2N_EDGE_NUM_SUPERNODES];**            /**< Cookie sent in last REGISTER_SUPER. */
  n2n_route_t	      *routes;		      /**< Networks to route through n2n */
[...]

PR #346, opened 2020-08-04
fix register supernode loop - Another way

skyformat99:
see issue: #341
see pull requests : #343

It's exactly the same logic as the original code, 
Send registration messages to all supernodes in a loop.
The cookie is only updated at the loop beginning.


lucaderi:
Thanks

PR #347, opened 2020-08-04
On Windows, the Edge node exits when the network card is disabled and…

skyformat99:
On Windows, the Edge node exits when the network card is disabled and opened again.
this pull request fix its bug.


lucaderi:
Thanks

PR #349, opened 2020-08-04
added regular expression support

Logan007:
This pull request adds regular expression support to the supernode's community file intake.

Starting from #342'2 idea, this one also allows for encrypted communities by just using their name. However, the `.^$*+?[]\` are not allowed in regular community names anymore as their pure occurence is interpreted as indicator for a regular expression.

For example, a `communities` file provided to the supernode by `-c communities` with the two line

```
Logan00[0-7]'s
HQ
```

would only allow `-c Logan000\'s` … `-c Logan007\'s` (008 stays off) and ` -c HQ` at the edge.

Note that only fixed-name communities, i.e. not those defined by a regular expression, can be used with encrypted headers (`-H` at the edge). The fixed-name communities are internally marked as _non-purgeable_ so they do not get purged even if no peers are present anymore.

The `sss->lock_communities` field is more to be read as "_we do not allow any **new** communities but those that match a regular expression if given_". We still need it as it also indicates a communitywise completely free supernode if reset.

Implementation uses a new HASH_LIST of regular expressions that every newly presented community is checked against (packet handling of REGISTER_SUPER).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,6 +110,7 @@ add_library(n2n STATIC
         src/tuntap_netbsd.c
         src/tuntap_linux.c
         src/tuntap_osx.c
+        src/n2n_regex.c
         )
 
 if(DEFINED WIN32)
--- a/community.list
+++ b/community.list
@@ -18,4 +18,20 @@ ntop[0-1][0-9]
 #
 #   *  Only fixed-name communities are supported for header encryption (-H)
 #
-
+#   *  Regular expression support the following placeholders
+#      '.'        Dot, matches any character                           (meaningless, as full matches only)
+#      '^'        Start anchor, matches beginning of string            (meaningless, as full matches only)
+#      '$'        End anchor, matches end of string
+#      '*'        Asterisk, match zero or more (greedy)
+#      '+'        Plus, match one or more (greedy)
+#      '?'        Question, match zero or one (non-greedy)
+#      '[abc]'    Character class, match if one of {'a', 'b', 'c'}
+#      '[^abc]'   Inverted class, match if NOT one of {'a', 'b', 'c'}  (feature is currently broken)
+#      '[a-zA-Z]' Character ranges, the character set of the ranges { a-z | A-Z }
+#      '\s'       Whitespace, \t \f \r \n \v and spaces
+#      '\S'       Non-whitespace
+#      '\w'       Alphanumeric, [a-zA-Z0-9_]
+#      '\W'       Non-alphanumeric
+#      '\d'       Digits, [0-9]
+#      '\D'       Non-digits
+#
--- a/community.list
+++ b/community.list
@@ -1,5 +1,21 @@
 #
 # List of allowed communities
+# ---------------------------
+#
+#      these could either be fixed-name communities such as the following lines ...
 #
 mynetwork
 netleo
+#
+#      ... or regular expressions that a community name must fully match
+#      such as ntop[0-1][0-9] for communities from "ntop00" through "ntop19"
+# 
+ntop[0-1][0-9]
+#
+#   *  Note that fixed-name communities may not contain one of the following characters
+#      . ^ $  * + ? [ ] \
+#      as otherwise, they are interpreted as regular expression
+#
+#   *  Only fixed-name communities are supported for header encryption (-H)
+#
+
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -349,6 +349,7 @@ typedef struct sn_stats
 struct sn_community
 {
   char community[N2N_COMMUNITY_SIZE];
+  uint8_t             purgeable;              /* indicates purgeable community (fixed-name, predetermined (-c parameter) communties usually are unpurgeable) */
   uint8_t	      header_encryption;      /* Header encryption indicator. */
   he_context_t        *header_encryption_ctx; /* Header encryption cipher context. */
   he_context_t        *header_iv_ctx;	      /* Header IV ecnryption cipher context, REMOVE as soon as seperate fields for checksum and replay protection available */
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -155,6 +155,7 @@ typedef struct ether_hdr ether_hdr_t;
 #include "pearson.h"
 #include "portable_endian.h"
 #include "speck.h"
+#include "n2n_regex.h"
 
 #ifdef WIN32
 #define N2N_IFNAMSIZ            64
@@ -359,6 +360,13 @@ struct sn_community
   UT_hash_handle hh; /* makes this structure hashable */
 };
 
+struct sn_community_regular_expression
+{
+  re_t rule;         // compiles regular expression
+
+  UT_hash_handle hh; /* makes this structure hashable */
+};
+
 typedef struct n2n_sn
 {
     time_t start_time; /* Used to measure uptime. */
@@ -372,8 +380,9 @@ typedef struct n2n_sn
 	uid_t userid;
 	gid_t groupid;
 #endif
-    int lock_communities; /* If true, only loaded communities can be used. */
+    int lock_communities; /* If true, only loaded and matching communities can be used. */
     struct sn_community *communities;
+    struct sn_community_regular_expression *rules;
 } n2n_sn_t;
 
 /* ************************************** */
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -71,6 +71,10 @@
 				 	           bits of transform_id; will be obsolete as soon as compression gets
 						   its own field in the packet. REVISIT then. */
 
+/* (un)purgeable community indicator (supernode) */
+#define COMMUNITY_UNPURGEABLE		0
+#define COMMUNITY_PURGEABLE		1
+
 /* Header encryption indicators */
 #define HEADER_ENCRYPTION_UNKNOWN       0
 #define HEADER_ENCRYPTION_NONE          1
--- a/include/n2n_regex.h
+++ b/include/n2n_regex.h
@@ -1,0 +1,80 @@
+/**
+ * (C) 2007-20 - ntop.org and contributors
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not see see <http://www.gnu.org/licenses/>
+ *
+ */
+
+// taken from https://github.com/kokke/tiny-regex-c
+// under Unlicense as of August 4, 2020
+
+/*
+ *
+ * Mini regex-module inspired by Rob Pike's regex code described in:
+ *
+ * http://www.cs.princeton.edu/courses/archive/spr09/cos333/beautiful.html
+ *
+ *
+ *
+ * Supports:
+ * ---------
+ *   '.'        Dot, matches any character
+ *   '^'        Start anchor, matches beginning of string
+ *   '$'        End anchor, matches end of string
+ *   '*'        Asterisk, match zero or more (greedy)
+ *   '+'        Plus, match one or more (greedy)
+ *   '?'        Question, match zero or one (non-greedy)
+ *   '[abc]'    Character class, match if one of {'a', 'b', 'c'}
+ *   '[^abc]'   Inverted class, match if NOT one of {'a', 'b', 'c'} -- NOTE: feature is currently broken!
+ *   '[a-zA-Z]' Character ranges, the character set of the ranges { a-z | A-Z }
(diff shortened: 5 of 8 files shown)


skyformat99:
There were problems with the test.

supernode -c config.ini -l 7788 -v
file config.ini:
abc[0-9]
eee*
ddd

edge -a 10.1.0.1  -c abcef000  -k zzz  -l xx.xx.xx.xx:7788 
Expectation: Connection failed
Actual result: Successful connection

edge -a 10.1.0.1  -c ddd1 -l  xx.xx.xx.xx:7788
Expectation: Connection failed
Actual result: Successful connection

skyformat99:
We should discuss the specific requirements:
1 When SN does not add the -c option, there is no restriction on community name.
2. When SN adds -c config-File, community name is restricted and only those that pass the filter condition can be connected.

config-file includes two format of rules:
1) abc  :  does not contain special characters, full text match, can support -H encryption.
2) fa[0-9]  :  contains special characters, regular expression filtering, does not support -H encryption.

For example, Contents of the config-file:
abc1
abc3
fa [0-9]
cnc [0-9]

So,   edge 
-c abc1    accept , Support for -H encryption
-c abc2    NG
-c abc3    accept,  Support for -H encryption
-c fa00    accept,  -H encryption is not supported
-c cnc00  accept,  -H encryption is not supported
-c cnczz   NG

Logan007:
I am with  completely with you concerning the specs, that is exactly the way it is supposed to be. So, I guess that evaluation of matches is wrong, in `sn_utils.c`'s loop

```
HASH_ITER(hh, sss->rules, re, tmp_re) {
      allowed_match = re_matchp(re->rule, cmn.community, &match_length);
      if(allowed_match != -1)
        break;
    }
```
Is evaluation of return values from `re_matchp()` correct?

skyformat99:
> I am with completely with you concerning the specs, that is exactly the way it is supposed to be. So, I guess that evaluation of matches is wrong, in `sn_utils.c`'s loop
> 
> ```
> HASH_ITER(hh, sss->rules, re, tmp_re) {
>       allowed_match = re_matchp(re->rule, cmn.community, &match_length);
>       if(allowed_match != -1)
>         break;
>     }
> ```
> 
> Is evaluation of return values from `re_matchp()` correct?

It should be correct, because I tested  #342.
If we all agree on this specification, 
so we only need to add the STRPBRK () function to #342 , then  support the -H encryption method.
I will try to fix it.

Logan007:
It might be easier to adopt this PR to not store the patterns but the original strings and then use `re_match()` and not `re_matchp()` for matching. Will be slower, though (as patterns will be re-compiled every time tested).

#342 has not taken precautions against purging fixed-name communites yet.

Logan007:
@skyformat99 

Maybe, we should take more time. I do not think it will harm to delay this feature after the release. Let us take the appropriate time to make it a clean, fast and working solution.

skyformat99:
> @skyformat99
> 
> Maybe, we should take more time. I do not think it will harm to delay this feature after the release. Let us take the appropriate time to make it a clean, fast and working solution.

yes,That's right.
I think I've fixed the problem, and I haven't submitted PR yet.
https://github.com/skyformat99/n2n-1.git
I did a quick test , there is no problem at all,  Could you help to test it?

In fact, only  when the first time a SuperNode Register Message is sent, community name  filtering is performed,
So it doesn't affect efficiency.

Logan007:
I am having a look at the code and will do some testing.

Logan007:
Looking at the code (no testing yet), I have the following questions and hints:
- How will matched communities added be purged lateron? They should as supernode should handle the list of communities economically being re-ITERated all the time.
- What is the appropriate length for a regular expression in its string form?
- Performance-wise, I still am in fond of having the patterns pre-compiled to make actual matching faster. It helps scalability.
- Not sure if `num_communities` should also count the regular expressions.

However, I also took a look into the `n2n_regex.c` code to understand why the HASH_LIST of pre-compiled expressions does not seem to work in your tests. I think that I found that `n2n_regex.c` has a flaw and does not handle multiple compiles correctly due to only one static array inside the compile-function always re-using it again. It does not create per-expression instances. That is why it worked in my example (one regular expression) and did not work in your example (several regular expressions).

I would suggest that we fix `n2n_regex.c`, then #349 should be working.

Logan007:
I think I got it working. Every expression compilation now gets its own space in memory. Also, I modified the ITERation more according to your scheme, @skyformat99, (see `match` inside the loop) because `allowed_match` alone does not neccessarily indicate a **full** match (starting position of the match == 0, could also be an infix or suffix). Thus, even though it might look odd (comparison to `-1` and to `0`, compiler will fix), I implemented it in a way one could easily remove the "full match" condition if required.

For must users, full matches might be more intuitive.

Here, all tests went smoothly.

skyformat99:
thanks，very good
I will try test it.

skyformat99:
file sn_utils.c 

line 819:
===================================================================================
   if (!comm) {
      HASH_ITER(hh, sss->rules, re, tmp_re) {
        allowed_match = re_matchp(re->rule, cmn.community, &match_length);

        if( (allowed_match != -1)
          && (match_length == strlen(cmn.community)) // --- only full matches allowed (remove, if also partial matches wanted)
          && (allowed_match == 0)) {                 // --- only full matches allowed (remove, if also partial matches wanted)
          match = 1;
          break;
        }
      }
    }


Should be changed to  ====>

===================================================================================
if (!comm  &&  sss->lock_communities) {
      HASH_ITER(hh, sss->rules, re, tmp_re) {
        allowed_match = re_matchp(re->rule, cmn.community, &match_length);

        if( (allowed_match != -1)
          && (match_length == strlen(cmn.community)) // --- only full matches allowed (remove, if also partial matches wanted)
          && (allowed_match == 0)) {                 // --- only full matches allowed (remove, if also partial matches wanted)
          match = 1;
          break;
        }
      }
      
      if (match !=1) { 
          return  -1;     // When the names don't match, directly return.
      }
    }

Logan007:
I see two changes you suggest:
1. The condition of first `if (!comm && sss->lock_communities)`
2. Fast track `return`

I fully agree to number **1**. :+1: 

But I am not sure about the second one. Even though I like the idea of the fast track`return`, we should offer the code the opportunity to run until the traceEvent "unallowed community" output. I would not want to duplicate the same error-output.

Does it otherwise work well for you?

skyformat99:
for safety, return -1 directly.
If trace is displayed, 0 will be returned.
If there is any update in the future, it is easy to introduce errors
If you want to display the trace, you can put it before return -1

Logan007:
Hmmm, the circumstances are the same (not allowed community presented, non-matching name, neither fixed-name nor regular expression). Shouldn't the behavior be the same?

skyformat99:
return 0 and return-1 should have different meanings.
And several places in the first half of this function are return-1.

Logan007:
Oh, now I see what you are talking about… :facepalm:  Then, we need to change the return value in case of an unaccepted fixed-name community too. Do you see an appropriate place to do it (error message **and** return value) together for both cases – rejected fixed-name **and** regular-expressed community?

In whatever way we solve it, these two cases should go down the same path through the code. I agree that the return code should be `-1` in both cases but I am not able to see any reason why non-matching regex-communities should be treated differently from non-matching fixed-name communities.

skyformat99:
You're right, it should return -1 if there's a mismatch (both of fix name & regex).
But your code now returns 0 in the case of a  regex  mismatch.

Logan007:
> But your code now returns 0 in the case of a regex mismatch.

Yes, that  is the original behaviour in case of a mismatch. I think we can change it to -1.

Logan007:
So, how about this: 

We let the code run through (no quick return) to handle it the same and do a  `return -1` right after the error message:
```
…
      traceEvent(TRACE_DEBUG, "Tx REGISTER_SUPER_ACK for %s [%s]",
                 macaddr_str(mac_buf, reg.edgeMac),
                 sock_to_cstr(sockbuf, &(ack.sock)));
    } else {
      traceEvent(TRACE_INFO, "Discarded registration: unallowed community '%s'",
                 (char*)cmn.community);
      return -1;
    }
    break;
  }
  case MSG_TYPE_QUERY_PEER: {
…
```

skyformat99:
> So, how about this:
> 
> We let the code run through (no quick return) to handle it the same and do a `return -1` right after the error message:
> 
> ```
> …
>       traceEvent(TRACE_DEBUG, "Tx REGISTER_SUPER_ACK for %s [%s]",
>                  macaddr_str(mac_buf, reg.edgeMac),
>                  sock_to_cstr(sockbuf, &(ack.sock)));
>     } else {
>       traceEvent(TRACE_INFO, "Discarded registration: unallowed community '%s'",
>                  (char*)cmn.community);
>       return -1;
>     }
>     break;
>   }
>   case MSG_TYPE_QUERY_PEER: {
> …
> ```

Very good, thanks
My suggestion is to add return-1 in both places.
Of course, we have to respect Lucaderi's opinion in the end.

skyformat99:
Also,  need to add a memory-free statement to the function void sn_term(n2n_sn_t *sss).
like this:
HASH_ITER(hh, sss->rules, re, tmp_re) {
    HASH_DEL(sss->rules, re);
    free(re);
  }


This statement may not be correct,  
about  re->rule = re_compile(line);    
Where is the calloc memory released?   Refer to the end of the function  re_compile().

skyformat99:
In addition,  static int load_allowed_sn_community(n2n_sn_t *sss, char *path) ：
Add the following statement:

uint32_t num_regex = 0;

When a new regular expression exists, then
num_regex++;


At the end of the function, it should be changed to
if (num_regex>0 || num_communities>0 )
{
   sss->lock_communities = 1;
}


Otherwise, there will be a problem if the configuration file does not have any valid  line.

Logan007:
> Also, need to add a memory-free statement to the function void sn_term(n2n_sn_t *sss)

You are right.

Logan007:
> uint32_t num_regex = 0;

> When a new regular expression exists, then
num_regex++;

You are right. We should count them.

> At the end of the function, it should be changed to
if (num_regex>0 || num_communities>0 )
{
sss->lock_communities = 1;
}
>
> Otherwise, there will be a problem if the configuration file does not have any valid line.

That is debatable. If the user explicitly specifies `-c <file>`, she probably expects to have closed-community supernode. If there happens to be not a single valid line in that file, she should be warned. But it should not default to start up a completely free supernode – which would be completely opposite to the users's expectation in this case.

_P. S._ I just saw that invalid file will be handled returning `-1`, too, without exiting and resulting in a free supernode. I will handle it the same way (warning as a minimum, return `-1`) for now. We just need to keep in mind that the return value does not get evaluated (orderly exit on `-1` in `setOption()`? That would be different pull request…).

Logan007:
> My suggestion is to add return-1 in both places.

I would not really care if it did not handle the very same case. But, if for any reason the error message needs to be changed or should ever additional handling be required in treating this case, it must be changed in two places.

However, you invented this feature, we do it your way!

skyformat99:
> > My suggestion is to add return-1 in both places.
> 
> I would not really care if it did not handle the very same case. But, if for any reason the error message needs to be changed or should ever additional handling be required in treating this case, it must be changed in two places.
> 
> However, you invented this feature, we do it your way!

thanks for your work.
But it's in the wrong place.
The following code is correct: (  in the    if(!comm && sss->lock_communities) {  ....   }    )

if(!comm && sss->lock_communities) {
      HASH_ITER(hh, sss->rules, re, tmp_re) {
        allowed_match = re_matchp(re->rule, cmn.community, &match_length);

        if( (allowed_match != -1)
          && (match_length == strlen(cmn.community)) // --- only full matches allowed (remove, if also partial matches wanted)
          && (allowed_match == 0)) {                 // --- only full matches allowed (remove, if also partial matches wanted)
          match = 1;
          break;
        }
      }
      // ---- mismatch regex, return -1
      if (match != 1) {
          traceEvent(TRACE_INFO, "Discarded registration: unallowed community '%s'",
              (char*)cmn.community);
          return -1;
      }
    }


Otherwise, an error will be returned when -c conf-file  is not added.

skyformat99:
sorry,  another error:
file sn.c  line 103  The correct one is:

if (  (num_regex + num_communities) < 1 )
  {
    traceEvent(TRACE_WARNING, "File %s does not contain any valid community names or regular expressions", path);
    return -1;
  }

skyformat99:
The bottom one is also wrong :
HASH_ITER(hh, sss->rules, re, tmp_re) {
    HASH_DEL(sss->rules, re);
    free(re);
  }

The correct ones are shown below:

HASH_ITER(hh, sss->rules, re, tmp_re) {
    HASH_DEL(sss->rules, re);
    if (NULL!=re->rule) {
        free(re->rule);
    }
    free(re);
  }

There are two functions that need to be fix:
static int load_allowed_sn_community(n2n_sn_t *sss, char *path)  ,
void sn_term(n2n_sn_t *sss)

Logan007:
> But it's in the wrong place.

You are right. Thank you for keeping tabs on the code!

> sorry, another error:

Okay. But I have put it yet another way.

> The bottom one is also wrong

Oh my bad… must be the weekend!

> There are two functions that need to be fix:
static int load_allowed_sn_community(n2n_sn_t *sss, char *path) ,
void sn_term(n2n_sn_t *sss)

Yes. That should be in a different pull request. ~~Maybe~~ Most probably, we will also find other parts in the code where return value handling can be done better.

skyformat99:
Thank you for your work, this will be a perfect solution.
I will close #342 .

lucaderi:
Thank you

Issue #247, opened 2020-05-19
Is there a faster implementation of Twofish available?


Logan007:
n2n's default cipher Twofish seems to be the slowest in all comparisons using `tools/n2n-benchmark`. Older [comparisons](https://panthema.net/2008/0714-cryptography-speedtest-comparison/) suggest that Twofish does not necessarily has to be slower than software AES.

Would anybody volunteer to check for faster implementations? I found [this](https://github.com/drewcsillag/twofish) older one which is [referenced](https://www.schneier.com/academic/twofish/download.html) (fourth line from the bottom) by Twofish's author Bruce Schneier. In a really quick attempt, it did not let me compile – as opposed to [this fork](https://github.com/fudanchii/twofish) of it.

I guess it could be moved to `twofish.c` if comparative testing should speedwise prove it superior.

lucktu:
By default, twofish is indeed slower than AES in most cases.
If you want to do some tests, please enter the QQ group: 5804301

Logan007:
@lucktu Thank you for offering testing. The currently used Twofish implementation shows roughly 55 percent of software AES' (openSSL) performance only¹ – on Intel as well as ARM. What platforms do you have available? MIPS, SPARC and maybe RISC-V would be interesting. I would be curious how current Twofish compares to software AES on those.

However, somebody would need to port the above mentioned fork (or an even faster implementation – let us know) to `twofish.c` maintaining full compatibility for – as a next step – conducting a comparison between these different Twofishes under the same conditions on different platforms.
___________________
¹ This might not be a too fair comparison as openSSL's software implementations (not the hardwardware accelerated) surely employ highly platform-optimized  implementations (possibly MMX, SSE, AVX, NEON) – a road n2n itself probably does not go down. Anyway, if there is a faster standard C Twofish implementation available, we should use it.

lucktu:
I'm a little busy at the moment. I don't have much time for tests. 

But I've done a lot of testing before, both here and on my blog (mainly for mips) .

Logan007:
@lucktu Good to hear that you have MIPS available for testing... we might need it later whenever another implementation of Twofish shows up.

From [this example](https://github.com/ntop/n2n/issues/230#issuecomment-619683650) I get that Twofish compared to AES seems to perform even worse on MIPS (only 45 percent of AES' performance). Another reason to look out for a faster implementation.

Let me add, as a postscript, that your n2n-related articles on your webpage are quite an impressive resource!

[...]

PR #357, opened 2020-08-12
simplified twofish's key handling

Logan007:
This pull request changes, i.e. simplifies Twofish's key handling _from_ using the repeated user password as key _to_ using a hash of the user password as key. Hashing happens in the `transform_tf.c` file conforming to the other transforms and also following the idea of having just the pure cipher inside the `twofish.c` file as it would allow for an easier replacement with another implementation if ever required.

Finally, hashing the key can be considered more secure as it makes more unpredictable use of the key space. Just think of the usually reset MSB of common ASCII-coded passwords.

To test this enhancement in current dev branch, all participating edges must be updated because compatibility is not maintained.

_P. S._ This pull request also solves the security related issue of repeating keys being interpreted as equal (due to repetition): Just try Twofish `-A2` with (literally) `-k blablabla` at one edge and `-k bla` (only) at another edge, they will talk to each other!

### DIFF
--- a/include/twofish.h
+++ b/include/twofish.h
@@ -62,7 +62,6 @@ typedef uint8_t  uint8_t;
 
 /* Constants */
 
-#define TwoFish_DEFAULT_PW		"SnortHas2FishEncryptionRoutines!" /* default password (not more than 32 chars) */
 #define TwoFish_DEFAULT_PW_LEN		32
 #define TwoFish_MAGIC			"TwoFish"			/* to indentify a successful decryption */
 
@@ -133,13 +132,13 @@ typedef struct
  *	This routine generates a global data structure for use with TwoFish,
  *	initializes important values (such as subkeys, sBoxes), generates subkeys
  *	and precomputes the MDS matrix if not already done.
- *	
- *	Input:	User supplied password (will be appended by default password of 'SnortHas2FishEncryptionRoutines!')
  *
- *  Output:	Pointer to TWOFISH structure. This data structure contains key dependent data.
+ *      Input:  User supplied key of correct length (TwoFish_KEY_LENGTH, 256 bits = 32 bytes by default)
+ *
+ *      Output:	Pointer to TWOFISH structure. This data structure contains key dependent data.
  *			This pointer is used with all other crypt functions.
  */
-TWOFISH *TwoFishInit(const uint8_t *userkey, uint32_t keysize );
+TWOFISH *TwoFishInit(const uint8_t *userkey);
 
 
 /*	TwoFish Destroy
--- a/src/transform_tf.c
+++ b/src/transform_tf.c
@@ -180,6 +180,7 @@ int n2n_transop_twofish_init(const n2n_edge_conf_t *conf, n2n_trans_op_t *ttt) {
   transop_tf_t *priv;
   const u_char *encrypt_key = (const u_char *)conf->encrypt_key;
   size_t encrypt_key_len = strlen(conf->encrypt_key);
+  uint8_t key_hash[32];
 
   memset(ttt, 0, sizeof(*ttt));
   ttt->transform_id = N2N_TRANSFORM_ID_TWOFISH;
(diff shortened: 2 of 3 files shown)


skyformat99:
Very good.
If you can speed it up, that would be great, now Twofish's processing is very slow.

fengdaolong:
Some people say that Twofish is slow, if compatibility is not considered, can we remove Twofish that affects speed, only need to keep the best encryption methods. or choose one of the fastest encryption algorithms as the default encryption.

Logan007:
For some people, choosing an encryption algorithm is a touchy decision. Some would never use anything approved or even developed by NSA (AES, SPECK) and would prefer Bruce Schneier's Twofish. Others even might like or not like the way Daniel J. Bernstein implemented and promoted his ChaCha20. Even others think that they all are safe to use (in my opinion, they are) and do not care for anything but speed (which equals heat, battery life, throughput, time for other tasks) and use the fastest one. So, I think it is crucial to offer a choice of different flavors as reasonable as possible. The four ciphers we have got now should serve everybody's requirements. I would not remove Twofish, it would be missing as a distinct option.

But indeed, I am planing a complete rework of the Twofish implementation. It starts with this little change here and will probably comprise implementation of [CTS](https://en.wikipedia.org/wiki/Ciphertext_stealing#CBC_implementation_notes) mode inside `transform_tf.c` (so `twofish.c` only will take care of the basic ECB and CBC mode crypto operations). Afterwards, in another step, I will replace the `twofish.c` with another implementation which [seems](https://github.com/ntop/n2n/issues/247#issuecomment-667912323) to be faster. On top of that, if I still am motivated and it proves beneficial, I will try to bring an SSE-accelerated implementation which might (!) help to further speed-up decryption.

The Twofish transform will not be compatible afterwards anymore, not even to this Pull Request because it seems that the CTS scheme that can currently be found in `twofish.c` in `_TwoFish_BlockCrypt()` around line 662 is not exactly the same as the one I plan for. I did not dare yet to dig into that portion of the code but did some comparative testing – (hackily) implemented my scheme in one edge and tried to have it communicate with the original.

So yeah, this is just the beginning of reworking Twofish implementation. I want to do it step by step in reasonably-sized chunks for the sake of transparency. It might even happen that I suggest CTS to AES before I implement it for Twofish. But those are just plans, let's see how it all works out.

skyformat99:
> For some people, choosing an encryption algorithm is a touchy decision. Some would never use anything approved or even developed by NSA (AES, SPECK) and would prefer Bruce Schneier's Twofish. Others even might like or not like the way Daniel J. Bernstein implemented and promoted his ChaCha20. Even others think that they all are safe to use (in my opinion, they are) and do not care for anything but speed (which equals heat, battery life, throughput, time for other tasks) and use the fastest one. So, I think it is crucial to offer a choice of different flavors as reasonable as possible. The four ciphers we have got now should serve everybody's requirements. I would not remove Twofish, it would be missing as a distinct option.
> 
> But indeed, I am planing a complete rework of the Twofish implementation. It starts with this little change here and will probably comprise implementation of [CTS](https://en.wikipedia.org/wiki/Ciphertext_stealing#CBC_implementation_notes) mode inside `transform_tf.c` (so `twofish.c` only will take care of the basic ECB and CBC mode crypto operations). Afterwards, in another step, I will replace the `twofish.c` with another implementation which [seems](https://github.com/ntop/n2n/issues/247#issuecomment-667912323) to be faster. On top of that, if I still am motivated and it proves beneficial, I will try to bring an SSE-accelerated implementation which might (!) help to further speed-up decryption.
> 
> The Twofish transform will not be compatible afterwards anymore, not even to this Pull Request because it seems that the CTS scheme that can currently be found in `twofish.c` in `_TwoFish_BlockCrypt()` around line 662 is not exactly the same as the one I plan for. I did not dare yet to dig into that portion of the code but did some comparative testing – (hackily) implemented my scheme in one edge and tried to have it communicate with the original.
> 
> So yeah, this is just the beginning of reworking Twofish implementation. I want to do it step by step in reasonably-sized chunks for the sake of transparency. It might even happen that I suggest CTS to AES before I implement it for Twofish. But those are just plans, let's see how it all works out.

Thanks,  That's a good idea.

lucaderi:
Thanks

PR #358, opened 2020-08-15
removed minor transform version

### LINKED ISSUE #273: unencrypted minor crypto version field in transforms
With header encryption in place, the only identifiable information are the unencrypted parts of the transforms.

Apart from the null transform `-A1`, the other transforms expose an unencrypted part by having a minor version number (for now, usually `0x01`) right at the beginning of the payload, probably always in the same place (+/- optional socket information). This makes n2n-packets well distinguishable from others. It should be removed or handled differently, e.g. encrypted but resolvable for all future versions.

These changes to the payload encryption schemes would break compatibility and therefore should be saved for a later version (3.0?).

Logan007:
This pull request removes the minor version field from all crypto transforms – saving 1 byte per packet and removing a detectable n2n packet characteristic.

By the way, Twofish transform lost another 4 bytes weight by removing the unused SA number.

Resolves #273.

### DIFF
--- a/src/transform_aes.c
+++ b/src/transform_aes.c
@@ -25,7 +25,6 @@
 #include <openssl/evp.h>
 #include <openssl/err.h>
 
-#define N2N_AES_TRANSFORM_VERSION       1  /* version of the transform encoding */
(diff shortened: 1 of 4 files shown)


lucaderi:
Thanks

PR #359, opened 2020-08-15
split PACKET's transform field to give way for separate compression field

Logan007:
This pull request splits PACKET's former 16-bit _transform_ field into two 8-bit fields, namely _compression_ and _transform_. Room for 256 compression and the same number of encryption schemes should be reasonable.

Also, the internal compression numbering scheme was updated as the former scheme considered compatibility in the shared field (before, `0` coded _no compression_).


lucaderi:
This is actually a good idea

fengdaolong:
From now on, we should pay attention to the protocol version number "N2N_PKT_VERSION". When the edge protocol version does not match, we will give the user a clear prompt.

Logan007:
I plan to increment that version soon (#375).

If we increase now, user do not get a clue (from the error transform id error message) on edges that still need to upgrade. Increasing the version will just discard the packets (as it is implemented now). Also, I want to implement a few more packet-format changing things (mainly within the crypto schemes) before I would increase from 2 to 3.

PR #360, opened 2020-08-15
updated minilzo

Logan007:
This pull request updates minilzo to current version especially to fix an issue which sometimes (!) arises during `lzo_init()` probably due to memory alignment.


Logan007:
… but the issue persists – sometimes, during start-up, `lzo_init()` makes the edge exit due to an error. I have not been able yet to track it down further. Anyway, this update does not hurt.

lucaderi:
Memory alignment is sometime a problem on ARM platforms

PR #361, opened 2020-08-15
merged additional building information

Logan007:
This pull request concentrates additional building information (different platforms, special options) in one new document `doc/Building.md`.


lucaderi:
Thanks

PR #363, opened 2020-08-16
extended edge's management port output

Logan007:
This pull request extends edge's management port output by a list of connected edges. It is adopted from the latest repective changes to the supernode's management port output. All credits to @fengdaolong!

In foresight, it also outputs the *tun_tap* IP which internally is not written to the the internal _peer structure_'s field yet.

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -201,7 +201,6 @@ n2n_edge_t* edge_init(const n2n_edge_conf_t *conf, int *rv) {
       traceEvent(TRACE_ERROR, "LZO compression error");
       goto edge_init_error;
     }
-
 #ifdef N2N_HAVE_ZSTD
   // zstd does not require initialization. if it were required, this would be a good place
 #endif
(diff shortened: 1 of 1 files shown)


lucaderi:
Thanks

fengdaolong:
@Logan007 
Thank you!
I am currently working on the function of making the edge management terminal display the IP of the tuntap adapter. There are currently two solutions:
1. Extract the association relationship between MAC and IP from the arp table of the system kernel;
2. Modify the edge transmission message body, and the edge actively reports its own IP;

I have implemented the first scheme, but I found that different system APIs are different and the compatibility is not ideal; I also implemented the second scheme. At present, the message logic in the edge code is rather messy. I plan to sort out the message logic and code structure.

fengdaolong:
I will submit the code for the second solution later.

Logan007:
Perhaps, ARP is too complicated to access and very platform dependant. I agree!

I think this point is related to the auto IP address questions.

What do you think about the proposal using QUERY_PEER (#367) with an additional IP address field? The edge could get/collect and record the IP addresses of the other peers from the PEER_INFOs as they arrive.

PEER_INFOs would need to be triggered from time to time, though. Maybe an ITERation through all known nodes from time to time sending out QUERY_PEER (edge) or PEER_INFO (supernode)? This could happen in the supernode as it only happens once then (n² = n messages for each edge). Otherwise, 2 x n² (each edge asking and getting n answers producing 2 packets each).

Issue #367, opened 2020-08-16
auto IP


Logan007:
The freshly added auto IP feature (#355) can make the supernode assign IP addresses cross-community from the same sub-network's address range. That is an issue as different communities usually mean different sub-networks (sometimes even several edges on the very same computer).

Is it possible to modify this feature such that IP addresses are assigned on a per-community basis according to their very own (predefined?) sub-networks?

fengdaolong:
If different communities use different subnets, there are two solutions:
1. The super node loads a configuration file. The content of the file is the correspondence between the complete community name and the IP subnet. When the community name exactly matches, the corresponding IP is automatically assigned;
2. Strengthen your recently implemented regular expression filtering community name function. Each community filtering condition corresponds to an IP subnet. When the community name meets the regular expression rules, the corresponding IP is automatically assigned.

For super nodes, it is impossible to determine the community name of the edge. I think the second solution is relatively better.

Logan007:
> correspondence between the complete community name and the IP subnet

Perhaps, if we leave it un-determined for now, each (new) community can get its own new subnet range from which it draws the new IPs.

Extension to config-file reading could be done at some later point in time, e.g. some special syntax `<community name> <IP sub-network>` delimited by a blank (that would exclude blanks from community names).

> For super nodes, it is impossible to determine the community name of the edge. I think the second solution is relatively better.

The name is transmitted in the COMMON section of each packet, even the REGISTER_SUPER (if IP address proposal is sent back using REGISTER_SUPER_ACK). Even if not in the hash list yet (first edge registering), it should be readable from `cmn.community`.

> recently implemented regular expression filtering community name function

Yes, I guess that would be the place, somewhere during the processing of the REGISTER_SUPER. Alternatively,  the QUERY_PEER could be extended by some IP address field delivering a proposed IP address if asked about the very same peer (srcMac == queryMac)? The edge would need to query the supernode about itself (after registration) if it does not carry a fixed IP address.

fengdaolong:
> Perhaps, if we leave it un-determined for now, each (new) community can get its own new subnet range from which it draws the new IPs.

A new community is created and new IPs can be extracted from it, but what if there are multiple different IP subnets in the community? For example, there are two edges, the IP is `192.168.5.7/24`, `172.16.88.99/24`, then which one should I choose?

[...]

PR #369, opened 2020-08-16
Enhanced edge management console function

fengdaolong:
1. Edge can view the IP address and socket information of the connected tap adapter through the management console (127.0.0.1:5644);
2. Repair two memory leaks;
3. Fix several code errors and redundant logic;

### DIFF
--- a/src/wire.c
+++ b/src/wire.c
@@ -251,42 +251,44 @@ int decode_sock( n2n_sock_t * sock,
     return (idx-idx0);
 }
 
-int encode_REGISTER( uint8_t * base,
-                     size_t * idx,
-                     const n2n_common_t * common,
-                     const n2n_REGISTER_t * reg )
-{
-    int retval=0;
-    retval += encode_common( base, idx, common );
-    retval += encode_buf( base, idx, reg->cookie, N2N_COOKIE_SIZE );
-    retval += encode_mac( base, idx, reg->srcMac );
-    retval += encode_mac( base, idx, reg->dstMac );
-    if ( 0 != reg->sock.family )
-    {
-        retval += encode_sock( base, idx, &(reg->sock) );
-    }
 
-    return retval;
+int encode_REGISTER(uint8_t *base,
+                    size_t *idx,
+                    const n2n_common_t *common,
+                    const n2n_REGISTER_t *reg) {
+	int retval = 0;
+	retval += encode_common(base, idx, common);
+	retval += encode_buf(base, idx, reg->cookie, N2N_COOKIE_SIZE);
+	retval += encode_mac(base, idx, reg->srcMac);
+	retval += encode_mac(base, idx, reg->dstMac);
+	if (0 != reg->sock.family) {
+		retval += encode_sock(base, idx, &(reg->sock));
+	}
+	retval += encode_uint32(base, idx, reg->dev_addr.net_addr);
+	retval += encode_uint8(base, idx, reg->dev_addr.net_bitlen);
+
+	return retval;
 }
 
-int decode_REGISTER( n2n_REGISTER_t * reg,
-                     const n2n_common_t * cmn, /* info on how to interpret it */
-                     const uint8_t * base,
-                     size_t * rem,
-                     size_t * idx )
-{
-    size_t retval=0;
-    memset( reg, 0, sizeof(n2n_REGISTER_t) );
-    retval += decode_buf( reg->cookie, N2N_COOKIE_SIZE, base, rem, idx );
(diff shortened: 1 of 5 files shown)


fengdaolong:
Function screenshot:
image

fengdaolong:
@Logan007  When this PR is merged, you can also optimize this feature, especially the redundant message code left over from the edge history, which takes a lot of time to straighten out.

fengdaolong:
In the future, the management console will also display each edge: transmission bytes, transmission speed, online and offline status, and so on.

Logan007:
I am not sure what part of the pull request goes in which of the 3 categories you mentioned above. I will have to go through it in more detail, see our discussion at #367

src/wire.c
```diff
@@ -251,42 +251,44 @@ int decode_sock( n2n_sock_t * sock,
...
+                    const n2n_REGISTER_t *reg) {
+	int retval = 0;
+	retval += encode_common(base, idx, common);
+	retval += encode_buf(base, idx, reg->cookie, N2N_COOKIE_SIZE);
+	retval += encode_mac(base, idx, reg->srcMac);
+	retval += encode_mac(base, idx, reg->dstMac);
+	if (0 != reg->sock.family) {
+		retval += encode_sock(base, idx, &(reg->sock));
+	}
+	retval += encode_uint32(base, idx, reg->dev_addr.net_addr);
```
Logan007, line 267: Will the edges be able to exploit this information, e.g. for directly injecting it into ARP table (that is probably already handled by OS)? So, this is just for internal information?

Maybe this is a good occasion to send out edge properties and information via PEER_INFO? :wink:

fengdaolong: When the edge requests data from the peer, the edge first initiates a query to the peer through the ARP protocol according to the MAC address, and then writes it into the ARP table. This is done automatically by operating system kernel.

Logan007: So, do I get it correctly that sending the IP address in this packet here is more of a _nice-to-send_ character? If so, that is why I would suggest the PEER_INFO…

fengdaolong:
At present, the functional difference between REGISTER_SUPER and PEER_INFO is not big. But I think they should assume different functional roles in the future. REGISTER_SUPER focuses on transmitting static data, such as the n2n version number of the edge, the tap adapter IP address, operating system details, etc. These data will not change frequently; and PEER_INFO should transmit dynamic data, such as: edge CPU usage, memory usage, etc. These data change in real time.

src/wire.c
fengdaolong: At present, the functional difference between REGISTER_SUPER and PEER_INFO is not big. But I think they should assume different functional roles in the future. REGISTER_SUPER focuses on transmitting static data, such as the n2n version number of the edge, the tap adapter IP address, operating system details, etc. These data will not change frequently; and PEER_INFO should transmit dynamic data, such as: edge CPU usage, memory usage, etc. These data change in real time.

Logan007: In my understanding, REGISTER_SUPER (and REGISTER as well) are to introduce nodes to each other only (if our vision becomes reality, those two packets might become just a simple REGISTER one day). It must contain everything that is required to start working. In my opinion, everything else excessing this minimum should probably go into PEER_INFO.

Concerning the PEER_INFO, I would not distinguish by static and dynamic. For example, PEER_INFO could provide the feature set of supported (de-)compression which is static. As soon as another edge knows about de-compression ability (keeping track of other edge's features), it could start sending compressed packets (if itself supports it).

fengdaolong: This consideration is reasonable.🍻

fengdaolong:
If there are not many questions about this PR, I suggest our BOSS @lucaderi merge it first, and then make improvements, because too many people expect this feature.
What do you think? @Logan007

src/wire.c
Logan007: I think I might add this just-developed compression-suggestion to my to-do list :smile:

Logan007:
I would be okay with it. In my limited test environment, it seems to work well. But again, I am not able to cover all scenarios…

But hey, wait: Is it compatible to the network-range code #372?

Logan007:
Also, what are the exact changes to the logic you did apply?

fengdaolong:
>Also, what are the exact changes to the logic you did apply?

The redundant logic was originally planned to be greatly changed, but the PR did not change much. It was only written in the comments.

fengdaolong:
>But hey, wait: Is it compatible to the network-range code #372?

It is compatible in theory.

PR #372, opened 2020-08-18
let auto ip feature handle several sub-networks

Logan007:
Following the discussion at #367, this pull request adds the ability to handle several sub-networks within the auto ip address feature.

A description is included in the `community.list` file.

Supernode's `-d` parameter saw some changes to accept a range of networks, e.g. `-d 10.0.1.0-10.0.5.0/24` but it can safely be ommited as default range is `10.128.0.0-10.255.255.0/24` offering 32768 (or so) sub-networks each of 254 (or so) addresses.

This is heavily based on @fengdaolong 's work trying to be as faithful as possible to the original.

Please test.

### DIFF
--- a/src/sn_utils.c
+++ b/src/sn_utils.c
@@ -218,12 +218,12 @@ int sn_init(n2n_sn_t *sss) {
   sss->mport = N2N_SN_MGMT_PORT;
   sss->sock = -1;
   sss->mgmt_sock = -1;
-  sss->min_dhcp_net.net_addr = inet_addr(N2N_SN_MIN_DHCP_NET_DEFAULT);
-  sss->min_dhcp_net.net_addr = ntohl(sss->min_dhcp_net.net_addr);
-  sss->min_dhcp_net.net_bitlen = N2N_SN_DHCP_NET_BIT_DEFAULT;
-  sss->max_dhcp_net.net_addr = inet_addr(N2N_SN_MAX_DHCP_NET_DEFAULT);
-  sss->max_dhcp_net.net_addr = ntohl(sss->max_dhcp_net.net_addr);
-  sss->max_dhcp_net.net_bitlen = N2N_SN_DHCP_NET_BIT_DEFAULT;
+  sss->min_auto_ip_net.net_addr = inet_addr(N2N_SN_MIN_AUTO_IP_NET_DEFAULT);
+  sss->min_auto_ip_net.net_addr = ntohl(sss->min_auto_ip_net.net_addr);
+  sss->min_auto_ip_net.net_bitlen = N2N_SN_AUTO_IP_NET_BIT_DEFAULT;
+  sss->max_auto_ip_net.net_addr = inet_addr(N2N_SN_MAX_AUTO_IP_NET_DEFAULT);
+  sss->max_auto_ip_net.net_addr = ntohl(sss->max_auto_ip_net.net_addr);
+  sss->max_auto_ip_net.net_bitlen = N2N_SN_AUTO_IP_NET_BIT_DEFAULT;
 
   return 0; /* OK */
 }
@@ -337,15 +337,15 @@ static signed int peer_tap_ip_sort(struct peer_info *a, struct peer_info *b) {
 }
 
 
-/** The IP address assigned to the edge by the DHCP function of sn. */
+/** The IP address assigned to the edge by the auto ip address function of sn. */
 static int assign_one_ip_addr(struct sn_community *comm,
                               n2n_ip_subnet_t *ipaddr) {
   struct peer_info *peer, *tmpPeer;
   uint32_t net_id, mask, max_host, host_id = 1;
   dec_ip_bit_str_t ip_bit_str = {'\0'};
 
-  mask = bitlen2mask(comm->dhcp_net.net_bitlen);
-  net_id = comm->dhcp_net.net_addr & mask;
+  mask = bitlen2mask(comm->auto_ip_net.net_bitlen);
+  net_id = comm->auto_ip_net.net_addr & mask;
   max_host = ~mask;
 
   HASH_SORT(comm->edges, peer_tap_ip_sort);
@@ -366,7 +366,7 @@ static int assign_one_ip_addr(struct sn_community *comm,
     }
   }
   ipaddr->net_addr = net_id | host_id;
-  ipaddr->net_bitlen = comm->dhcp_net.net_bitlen;
+  ipaddr->net_bitlen = comm->auto_ip_net.net_bitlen;
 
   traceEvent(TRACE_INFO, "Assign IP %s to tap adapter of edge.", ip_subnet_to_str(ip_bit_str, ipaddr));
   return 0;
@@ -384,8 +384,8 @@ int subnet_available(n2n_sn_t *sss,
 
   HASH_ITER(hh, sss->communities, cmn, tmpCmn) {
     if (cmn == comm) continue;
-    if( (net_id <= (cmn->dhcp_net.net_addr + ~bitlen2mask(cmn->dhcp_net.net_bitlen)))
-      &&(net_id + ~mask >= cmn->dhcp_net.net_addr) ) {
+    if( (net_id <= (cmn->auto_ip_net.net_addr + ~bitlen2mask(cmn->auto_ip_net.net_bitlen)))
+      &&(net_id + ~mask >= cmn->auto_ip_net.net_addr) ) {
         success = 0;
         break;
     }
@@ -395,7 +395,7 @@ int subnet_available(n2n_sn_t *sss,
 }
 
 
-/** The IP address assigned to the edge by the DHCP function of sn. */
+/** The IP address range (subnet) assigned to the community by the auto ip address function of sn. */
 int assign_one_ip_subnet(n2n_sn_t *sss,
                          struct sn_community *comm) {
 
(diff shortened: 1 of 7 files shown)


fengdaolong:
Please change all the "dhcp" words in the functions you implement and the functions I implement to "auto_ip" or "dynamic_ip" to avoid confusion with the real DHCP protocol, because it is not a real DHCP protocol.

Logan007:
Thanks for that hint. Will do immediately!

Logan007:
Any suggestion for supernode's cli option? Shall we switch to `-a` and `--autoip`?

fengdaolong:
> Any suggestion for supernode's cli option? Shall we switch to `-a` and `--autoip`?

OK！

Logan007:
Please have a look.

fengdaolong:
You are a hardworking bee.

Logan007:
I only exploited your hard work and re-framed it. I cannot take any credit for that.

Does it work for you the way it is intended? My testing capacity is limited.

fengdaolong:
Hard work!
I have no time to test in the last few days. If other people have no comments, I suggest to merge first and fix the problem after the test.

fengdaolong:
Are IP subnets random in different communities? Is non-repeatability guaranteed and can cover all subnet ranges?

image

src/sn_utils.c
```diff
@@ -364,13 +366,87 @@ static int assign_one_ip_addr(n2n_sn_t *sss,
...
+                     struct sn_community *comm,
+                     uint32_t net_id,
+                     uint32_t mask) {
+
+  struct sn_community *cmn, *tmpCmn;
+  int success = 1;
+
+  HASH_ITER(hh, sss->communities, cmn, tmpCmn) {
+    if (cmn == comm) continue;
+    if( (net_id <= (cmn->auto_ip_net.net_addr + ~bitlen2mask(cmn->auto_ip_net.net_bitlen)))
```
fengdaolong, line 387: Subnet and mask are reversed and added together. I don't quite understand this calculation method. Can you explain it?

src/sn_utils.c
```diff
@@ -364,13 +366,87 @@ static int assign_one_ip_addr(n2n_sn_t *sss,
...
+
+  mask = bitlen2mask(sss->min_auto_ip_net.net_bitlen);
+  // number of possible sub-networks
+  no_subnets   = (sss->max_auto_ip_net.net_addr - sss->min_auto_ip_net.net_addr);
+  no_subnets >>= (32 - sss->min_auto_ip_net.net_bitlen);
+  no_subnets  += 1;
+
+  // proposal for sub-network to choose
+  net_id  = pearson_hash_32(comm->community, N2N_COMMUNITY_SIZE) % no_subnets;
+  net_id  = sss->min_auto_ip_net.net_addr + (net_id << (32 - sss->min_auto_ip_net.net_bitlen));
```
fengdaolong, line 415: This calculation method is not very clear, and seems to be problematic.

fengdaolong (commented):
Please describe the calculation idea of ​​the subnet pool, which is difficult to understand.

Logan007:
Generally, sub-networks are taken from the default range or the one provided with `-d` to the supernode (if not explicitly specified differently in the `community.list` file).

Basically, every community gets its own sub-network. The sub-network assigned is disjoint from all other sub-networks, even from those explicitly assigned in the `community.list` file. Communities can only have overlapping networks if explicitly specified so in the `community.list` file.

The sub-network assigned depends on the hash of the community name. This way, there is a good chance that the same community finds back to the very same sub-network. This is to mitigate the scenario of all edges quitting the community, losing connection and re-connection later, getting assigned another sub-network (which we want to prevent if possible).

I will add some hints to your review remarks.

> Are IP subnets random in different communities?

Not random but based on the hash of the community name.

>  Is non-repeatability guaranteed …

It is checked for collisions. Collisions are not allowed. 

> …and can cover all subnet ranges?

If a collision is detected, it is searched upward / downwards starting from the hash-based proposal for an available, i.e. non-colliding sub-network. The eventually required search covers the whole range.

src/sn_utils.c
Logan007: The bit-reversed sub-network mask added to the network base address is the highest possible address in that sub-network.

src/sn_utils.c
Logan007: The community name gets hashed. We need to map the hash value to a sub-network.

To do so, we map the hash to a number of the generally available sub-networks (division modulo) which goes into `net_id` for now.

This number of a sub-network now needs to get mapped to a sub-network itself. It happens in the `(net_id << (32 - sss->min_auto_ip_net.net_bitlen)` part: `32 - bitlen` gives the number of bits used at the end of an IP address for the nodes. Shifting the (still: **number** of sub-network) `net_id` left by this amount of bits, gives a sub-network.

To make this subnet start at the minimum sub-network base address, we have to add just this minimum sub-network base address.

Afterwards, that is later in the code, it checks if this range is still available and if not, it searches downwards and upwards starting from there.

Why do you think it is problematic? It should very well stay in the allowed / defined ranges and give a higher probability that a community finds back to the same IP sub-network.

fengdaolong:
The calculation method of the subnet pool is incorrect and needs to be revised, please refer to [rfc1918](https://www.rfc-editor.org/rfc/rfc1918.html).

If you want to fix the subnet pool to a class A address, then all of its private address range is 10.0.0.0-10.255.255.255/8, and then you can divide it into multiple addresses by setting different mask bits in this address range Subnet pools, the following is an example.

If the mask bit is 11 (10.0.0.0/11), you can get 8 subnets, and the number of hosts in each subnet is 2097152, as follows:

`network     start       end             broadcast`
`10.0.0.0    10.0.0.1    10.31.255.254   10.31.255.255`
`10.32.0.0   10.32.0.1   10.63.255.254   10.63.255.255`
`10.64.0.0   10.64.0.1   10.95.255.254   10.95.255.255`
`10.96.0.0   10.96.0.1   10.127.255.254  10.127.255.255`
`10.128.0.0  10.128.0.1  10.159.255.254  10.159.255.255`
`10.160.0.0  10.160.0.1  10.191.255.254  10.191.255.255`
`10.192.0.0  10.192.0.1  10.223.255.254  10.223.255.255`
`10.224.0.0  10.224.0.1  10.255.255.254  10.255.255.255`




If the mask bit is 13 (10.0.0.0/13), you can get 32 subnets, and the number of hosts in each subnet is 524288, as follows:

`network     start       end             broadcast`
`10.0.0.0    10.0.0.1    10.7.255.254    10.7.255.255`
`10.8.0.0    10.8.0.1    10.15.255.254   10.15.255.255`
`10.16.0.0   10.16.0.1   10.23.255.254   10.23.255.255`
`10.24.0.0   10.24.0.1   10.31.255.254   10.31.255.255`
`10.32.0.0   10.32.0.1   10.39.255.254   10.39.255.255`
`10.40.0.0   10.40.0.1   10.47.255.254   10.47.255.255`
`10.48.0.0   10.48.0.1   10.55.255.254   10.55.255.255`
`10.56.0.0   10.56.0.1   10.63.255.254   10.63.255.255`
`10.64.0.0   10.64.0.1   10.71.255.254   10.71.255.255`
`10.72.0.0   10.72.0.1   10.79.255.254   10.79.255.255`
`10.80.0.0   10.80.0.1   10.87.255.254   10.87.255.255`
`10.88.0.0   10.88.0.1   10.95.255.254   10.95.255.255`
`10.96.0.0   10.96.0.1   10.103.255.254  10.103.255.255`
`10.104.0.0  10.104.0.1  10.111.255.254  10.111.255.255`
`10.112.0.0  10.112.0.1  10.119.255.254  10.119.255.255`
`10.120.0.0  10.120.0.1  10.127.255.254  10.127.255.255`
`10.128.0.0  10.128.0.1  10.135.255.254  10.135.255.255`
`10.136.0.0  10.136.0.1  10.143.255.254  10.143.255.255`
`10.144.0.0  10.144.0.1  10.151.255.254  10.151.255.255`
`10.152.0.0  10.152.0.1  10.159.255.254  10.159.255.255`
`10.160.0.0  10.160.0.1  10.167.255.254  10.167.255.255`
`10.168.0.0  10.168.0.1  10.175.255.254  10.175.255.255`
`10.176.0.0  10.176.0.1  10.183.255.254  10.183.255.255`
`10.184.0.0  10.184.0.1  10.191.255.254  10.191.255.255`
`10.192.0.0  10.192.0.1  10.199.255.254  10.199.255.255`
`10.200.0.0  10.200.0.1  10.207.255.254  10.207.255.255`
`10.208.0.0  10.208.0.1  10.215.255.254  10.215.255.255`
`10.216.0.0  10.216.0.1  10.223.255.254  10.223.255.255`
`10.224.0.0  10.224.0.1  10.231.255.254  10.231.255.255`
`10.232.0.0  10.232.0.1  10.239.255.254  10.239.255.255`
`10.240.0.0  10.240.0.1  10.247.255.254  10.247.255.255`
`10.248.0.0  10.248.0.1  10.255.255.254  10.255.255.255`

src/sn_utils.c
fengdaolong: If I have a subnet: `10.8.9.10/24`
According to your calculation: `0x0A08090A + (~0xFFFFFF00) = 0xA08AA89 = 10.8.170.137`
Then according to the calculation result, it is obviously wrong to get the highest address of `10.8.170.137`.

src/sn_utils.c
fengdaolong: To map the subnets according to the hash value of the community name, I think this method is not very meaningful, because the repetition rate is too high, and ultimately rely on search, which reduces the efficiency. It is better to increment the subnets in the subnet pool in turn .

src/sn_utils.c
Logan007: Well, I am not a network expert, but `10.8.9.10/24` does not seem to denote a valid network. I was assuming that networks always have `0`-bits in their node part, so `10.8.9.0/24` would be a valid network (as far as I remember, it is being checked for the zeros at input). Starting from there, the calculation is `0x0A080900 + ~(0xFFFFFF00)` = `0x0A080900 + 0x000000FF` = **`0x0A0809FF`** which seems fine to me.

Also, keep in mind that this is for ranges and not for the actually delivered node IP addresses (I am aware of the broadcast). Your part of the code which delivers node IP addresses from a certain subnet remained unchanged in this repect.

src/sn_utils.c
Logan007: The hash makes kind of sure that it is **not** repeating so easily, so it keeps repetition rate down. Mapping all the possible community names to the 32,768 sub-networks we offer by default, I think there is a low chance for collision – that is a exactly hashes purpose. By the way, for definitely non-repeating assignments, a network can be provided along with the community name in the `community.list` file.

The search is required only in case a collision happens. Starting from the hash-based proposal, it should not take longer than the number of currently used sub-networks (worst-case) to find an available network. It probably would be the same if you started from 0 using an incremental assignment. You also would need to search the pool from 0 to find the next available network (you cannot just increase leaving holes for communities that were abandoned).

Big advantage of hashing the community name: You probably get back your old sub-network even when all nodes from a community are gone for a while or longer. That is something that will not happen with incremental pool or pseudo-random assignment.

Logan007:
> The calculation method of the subnet pool is incorrect and needs to be revised, please refer to rfc1918.

Please see my calculation above. I do not see wrong results.

> If you want to fix the subnet pool to a class A address, then all of its private address range is 10.0.0.0-10.255.255.255/8, 

I decided to use the upper half of that class A net by default because I see a lot of people and public spaces using (mostly from the lower half) for their real world NATed IP address assignments. I did not want to collide.

> and then you can divide it into multiple addresses by setting different mask bits in this address range Subnet pools, the following is an example.

The auto IP feature takes sub-networks from that pool (upper half of that class A network) and provides all /24 networks out of it (by default).

> If the mask bit is 11 (10.0.0.0/11), you can get 8 subnets, and the number of hosts in each sub-net is 2097152, as follows:

If someone wants broader networks than the default /24, this could be done by applying `-d 10.0.0.0-10.224.0.0/11` to the supernode's command line to match your example.

> If the mask bit is 13 (10.0.0.0/13), you can get 32 subnets, and the number of hosts in each subnet is 524288, as follows:

Using `-d 10.0.0.0-10.224.0.0/13` should deliver exactly the same pool of sub-networks.

The `-d` option is not bound to that specific class A networks, it could be used with other address spaces.

Logan007:
Maybe we should write a markup for the `/doc` folder containing all those hints, especially details on how to use the `-d` option?

fengdaolong:
> it could be used with other address spaces.

I always think that this kind of network calculation method is a bit strange. For example, I set the following subnet pool, but the network range error is given.
`./supernode -l 9135 -a 10.0.0.0-10.0.0.255/29`
`[sn.c:242] WARNING: Bad network range '10.0.0.0...10.0.0.255/29' in '10.0.0.0-10.0.0.255/29', defaulting to '10.128.0.0...10.255.255.0/24'`

src/sn_utils.c
fengdaolong: >Big advantage of hashing the community name: You probably get back your old sub-network even when all nodes from a community are gone for a while or longer. That is something that will not happen with incremental pool or pseudo-random assignment.

In this respect, it may indeed be very useful.

src/sn_utils.c
fengdaolong: > but 10.8.9.10/24 does not seem to denote a valid network. 

In order to illustrate the problem, I wrote this address casually. In fact, in order to improve the fault tolerance, usually in this example, the IP and subnet mask are ANDed to clear the lowest bit "10" to 0.

Logan007: But this kind of address is invalid and thus not accepted by the program. The case you mentioned will not happen.

Logan007:
> `-a 10.0.0.0-10.0.0.255/29`

Again, same as above: `10.0.0.255/29` does not denote a valid network due to missing zero-bits in the node portion, thus the error message. Keep in mind that `-a` requires a range of **networks** and not node's IP addresses, so it is provided a starting network `10.0.0.0` (it shares the `/29`) and the last assignable network, e.g. `10.0.0.248/29` or `10.255.255.248/29` to make full use of the class A network.

I do not think that n2n should make assumptions on user input and correct them or re-interpret them. Better throw an error message.

fengdaolong:
OK, I know.
For the subnet range described above (10.0.0.0-10.0.0.255), the correct way to write it is `-a 10.0.0.0-10.0.0.248/29` (full expression: 10.0.0.0/29 - 10.0.0.248/29) instead of `-a 10.0.0.0-10.0.0.255/29` .
It contains a total of 32 subnets, and each subnet has 8 hosts.
All doubts originated from my writing errors.

Logan007:
So, no error message then for you?

I think our discussion shows the need of an explanatory document. We should draft a short `doc/AutoIP.md` somewhen soon.

fengdaolong:
Yes, explanatory documents are important, but automatic fault tolerance of programs is also very important, such as  `10.8.9.10/24`  I mentioned above.
Fault tolerance issues like `10.0.0.0-10.0.0.255/29` also need to be dealt with. If it is solved, no documentation is needed.
The solution is actually very simple. The IP entered in the `-a` parameter is ANDed with the subnet mask, and the host number is cleared to become the network number. In this way, you can enter the IP range at will without worrying about input errors.

Remember, good programs or products don't need to read documentation.

Logan007:
> Fault tolerance issues like 10.0.0.0-10.0.0.255/29 also need to be dealt with.

> > I do not think that n2n should make assumptions on user input and correct them or re-interpret them. Better throw an error message.

However, I have no objections if you want to change it.

> If it is solved, no documentation is needed.

I am not sure about that. The two of us are quite deep into it and may both not require documentation. But everybody else and the generations to come (…) might.

Also, do you think that the information in `community.list` is sufficient? At some point, we need to point the user to it, maybe supernode's help text for `-c` could be extended by some _see file 'community.list'_ or so.

fengdaolong:
>I have no objections if you want to change it.

I am just a suggestion, but I will follow your ideas.

About `community.list` , I haven't tested it yet. I have been very busy recently. The automatic IP multi-subnet function is just a simple experience today.
I still think that if there are no major problems, submit them first and then improve.

PR #373, opened 2020-08-18
seeded random number generator in supernode

Logan007:
This pull request adds the so far missing seeding of pseudo random number generator to the supernode – edge already has it.


Logan007:
_P. S._ This could also go into 2.8 stable as it does not change compatibility but significantly increases security (IV generation for header encryption in supernode).

lucaderi:
Backported to stable https://github.com/ntop/n2n/commit/62955193391b4e8ef67e879a75e1215c9078ea64

PR #380, opened 2020-08-23
changed aes transform to cipher text stealing mode

### LINKED ISSUE #238: What about Ciphertext Stealing (CTS) for AES-CBC?
Extending AES's CBC scheme by [Ciphertext Stealing](https://en.wikipedia.org/wiki/Ciphertext_stealing) (CTS), AES¹ packets would not need to be padded anymore. There would be savings of up to 16 bytes per data packet.

As far as I can judge, data confidentiality would not be affected. Only the (anyway very light) integrity check-ability gained from correct/reasonable padding information is dropped. This will be cushioned by some future packet-wide Message Authentication Code. Performance penalty should hardly be noticeable – I would create a proof of concept along this [scheme](https://en.wikipedia.org/wiki/Ciphertext_stealing#CBC_implementation_notes) to evaluate.

As CTS breaks compatibility, it should be implemented – if desired – together with other major changes (v3.0?) only. Or shall we make it an additional minor variant `N2N_AES_TRANSFORM_VERSION` of AES, namely `2`?

What do you think?

¹ Twofish could benefit from it, too.

### LINKED ISSUE #378: Rework AES
I took some time to review AES transform and plan for the following changes:

- With header encryption disguising internal MACs in place, I am way less concerned about payload's first plain text block. Thus, I would like to remove the AES's IV encryption. This saves the single AES-ECB step in IV generation and thus may speed things up a bit.

- ~~As I want to stick with flexible IV size and as it shall not be partly constant (the non-transmitted part will be constant as we skip the scrambling encryption step), I will revert to the _nonce_-like scheme already used in Twofish but using a 64 bit value by default.~~ I will implement full-size IV adding 8 more bytes to the packet.

- I will move the openSSL-related code into a new `aes.c` file to relieve the `transform_aes.c` from too specific code. This would allow to later eventually integrate some own AES code in `aes.c` more easily.

- I will replace the current key hashing with Pearson hashing for being independent from openSSL in this point. Also, this is a prerequisite for somewhen implementing own AES code.

- However, I would like to keep the general [scheme](https://github.com/ntop/n2n/issues/101#issuecomment-496020090) of password length-triggered key-length.

- Implement the [CTS](https://en.wikipedia.org/wiki/Ciphertext_stealing#CBC_implementation_notes) scheme. CTS will save between 1 and 16 bytes per packet as padding is not required anymore.

- The crypto-context shall be provided by `aes.c` to `transform_aes.c` allowing to flexibly handle it on transform-level while its internal structure depends on low level only.

On average, this will save 0.5 bytes per packet (added 8, saved (1+16)/2) and speed things up a bit. It will break compatibility to current dev for AES though – joining current disruptions.

Any thoughts or hints?

Logan007:
This pull request changes (rewrites) the aes transform from _cbc with padding_ to _cipher text stealing_ using the underlying cbc and ecb modes straight forward along [this](https://en.wikipedia.org/wiki/Ciphertext_stealing#CBC_implementation_notes) scheme. It still relies on openSSL.

The aes cipher's internals now are gathered in the new file `aes.c` whereas the cts-scheme considered higher-level logic residing in `transform_aes.c`.

A random number (128 bit, block sized) is prepended to the plaintext instead of using the IV (using a zeroed IV) to make sure that the minimum packet size of one block is reached (a requirement of cipher text stealing scheme). The regular packet size however will probably never fall below this limit as they carry header and stuff. So, it is debatable to **shorten the random number's length, maybe to 64 bit saving another 8 bytes per packet** – open an _Issue_ if you feel we should discuss. Just be aware, low level programmer, should you ever work with shorter packets.

Slight cipher-bound speed-ups in the 10 to 20 percent range are observed  as determined by `tools/n2n-benchmark`.

This pull request breaks compatibility to current _dev_ – breaking changes are limited to aes transform.

Fixes #378.
Fixes #238.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,7 @@ add_library(n2n STATIC
         src/transform_aes.c
         src/transform_cc20.c
         src/transform_speck.c
+        src/aes.c
         src/speck.c
         src/random_numbers.c
         src/pearson.c
--- a/doc/Crypto.md
+++ b/doc/Crypto.md
@@ -18,11 +18,11 @@ The following chart might help to make a quick comparison and decide what cipher
 | Cipher | Mode | Block Size | Key Size         | IV length |Speed | Built-In | Origin |
 | :---:  | :---:| :---:      | :---:            | :---:     |:---: | :---:    | ---    |
 |Twofish | CTS  | 128 bits   | 256 bit          | 32 bit    | -    | Y        | Bruce Schneier |
-|AES     | CBC  | 128 bits   | 128, 192, 256 bit| 64 bit    | O..+ | N        | Joan Daemen, Vincent Rijmen, NSA-approved |
+|AES     | CBC  | 128 bits   | 128, 192, 256 bit| 128 bit   | O..+ | N        | Joan Daemen, Vincent Rijmen, NSA-approved |
 |ChaCha20| CTR  | Stream     | 256 bit          | 128 bit   | +..++| N        | Daniel J. Bernstein |
 |SPECK   | CTR  | Stream     | 256 bit          | 128 bit   | ++   | Y        | NSA |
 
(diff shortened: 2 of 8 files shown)


Logan007:
I would require some feedback from testing it with native openSSL **1.0** support. I am not sure if my environment gives the correct results ~~("segmentation fault")~~.  _-- probably fixed by new commit, see below --_

PR #389, opened 2020-08-26
reworked twofish implementation and replaced it by a faster version

### LINKED ISSUE #247: Is there a faster implementation of Twofish available?

### LINKED ISSUE #350: Twofish can't handle packets sized under 12 bytes properly
Whenever packet size falls below 12 bytes, Twofish scheme decrypted packets will have 12 (16 byte block size _minus_ 4 byte nonce size) bytes in size (last unused bytes probably zeroed).

`twofish.c`'s handling of not fully filled one-block packets causes this. Longer packets that do not match block size do not see this problem as a CTS scheme is applied.

Every solution to this will break compatibility. I will think of something, maybe give Twofish a complete overhaul. But not for the upcoming release. For now, just be aware.

Logan007:
The Twofish cipher's internals were reworked using Drew Scillag's implementation as a basis which is considerably faster. They can be found in the new files `tf.c` and `tf.h` whereas the cts-scheme considered higher-level logic has been moved into  the`transform_tf.c` file.

A random number (128 bit, block sized) is prepended to the plaintext (instead of former 32-bit number only) so the IV is zeroed. Hence, we made sure that the minimum packet size of one block is reached (a requirement of cipher text stealing scheme). The regular packet size however will probably never fall below this limit as they carry header and stuff. **So, it is debatable to shorten the random number's length, maybe to 64 bit saving 8 bytes per packet** – open an Issue if you feel we should discuss. Just be aware, low level programmer, should you ever work with shorter packets.

Cipher-related speed-ups in the 50 to 60 percent range are observed on Intel CPUs using `tools/n2n-benchmark`. Other CPUs might well go beyond.

This pull request breaks compatibility to current _dev_ – breaking changes are limited to the Twofish transform.

Fixes #247.
Fixes #350.


Logan007:
Fixed `src/tf.c` whose `tf_decrypt_cbc(…)` did not allow for shared `in` and `out` previously.

Should be working now.

Please test and if possible, please share your speed-ups compared to previous Twofish implementation, do not forget to name your platform. Mine are as follows (all with `./configure CFLAGS="-O3 -march=native"`):

```
                                           -d
i7-2860QM    98.3 --> 151.0 MB/s      48.1 -->  78.6 MB/s

i7-7500U    119.3 --> 187.0 MB/s      61.0 --> 102.1 MB/s

Cortex A53   22.8 --> 47.0 MB/s       11.2 -->  23.8 MB/s    !!!
(RPi 3B+)
```

skyformat99:
good

lucaderi:
very good

PR #390, opened 2020-08-26
Avoid leaking memory if the buffer is uncompressable

hamishcoleman:


Logan007:
Good catch!

hamishcoleman:
Ha - it wasn't hard once I noticed my resident size was in the gigs!

Logan007:
Oh my… gigs?! Your edge must have seen a lot of throughput.

Your finding should definitely go into _dev_ branch as well.

hamishcoleman:
Yeah, stress/stability testing with iperf3 will do that!

I had a look at dev and saw that the code had not changed, so the patch should apply cleanly

fengdaolong:
After the compression_buffer is released, it needs to be set to null, otherwise the patch will be useless.

fengdaolong:
After the `compression_buffer` is free, It needs to be assigned to `NULL`, otherwise the patch may be useless.

fengdaolong:
There are many similar problems in the n2n project, which need to be dealt with one by one.

hamishcoleman:
In this case, the variable goes out of scope immediately after the free(), so the assignment to NULL is not needed

PR #395, opened 2020-08-29
twofish transform code clean-up amendment

Logan007:
Minor code clean-up in Twofish transform and cipher **not** breaking compatibility to current _dev_.


Logan007:
I wish I could type…

PR #397, opened 2020-08-30
made aes always available

Logan007:
This pull request makes AES always available by adding a public domain software implementation as default in case no openSSL can be found. It basically is the implementation that can also be found in openSSL 1.0 (but was slightly tuned for some more speed using one of Drew Scillag's tricks :wink:).

So, if executable size is of concern, it will be a much easier decision now to do without especially static openSSL – only will thou have to abstain from ChaCha20 though… which still relies on openSSL 1.1.

This is fully compatible to current dev.


Logan007:
In my opinion, we could abandon the openSSL **1.0** integration then (keeping openSSL **1.1** support only). I will prepare it inside this pull request.

Let me know if you disagree.

fengdaolong:
In the compilation configuration, OpenSSL can be used as an option, and the user decides whether to use it.

Logan007:
Yes, that would be great if we were able to achieve it in a comfortable way. Until then, I just will stick with auto detection (which might meet most users' needs and thus should be the default) and let manual editing `include/config.h` (defines) and `Makefile` (libs) do the job.

Do you have any idea how to better automate it? It probably must be some parameter to `./configure` whose handling must be coded in the `configure.seed` file.

Logan007:
> In my opinion, we could abandon the openSSL 1.0 integration then (keeping openSSL 1.1 support only). I will prepare it inside this pull request.

Thinking about it, I will prepare abandoning openSSL **1.0** (keeping **1.1** only) in another pull request (somewhen after this one being accepted). For testing and comparison, **1.0** should stay around a while.

PR #398, opened 2020-08-31
replaced Pearson hashing permutation by aes' s-box

Logan007:
This pull request replaces the Pearson hashing look-up table by [AES' S-box](https://en.wikipedia.org/wiki/Rijndael_S-box). **It breaks compatibility to current _dev_ completely**.

Pearson hashing works with every permutation of [0…255]. This however is a special permutation with careful designed properties. Nevertheless, the most important aspect is that a look-up can be performed extremely fast on AES accelerated hardware such as Intel's x86 AES-NI. It will help a lot speed-wise – mostly for header encryption check-summing for example, the key setup only happens once in a life-time.

I will implement the accelerated version soon. Just want to put this upfront to have all the compatibility breaking changes done as soon as possible to soon sail in calmer waters again.

From my side, this will be the **last major compatibility breaking change** for a while, so we could increase packet version soon (give it another week, or so).

Only two items with _minor_ compatibility breaking potential left on my to-do list: A revision of ChaCha20 will break ChaCha20's compatibility (only). Also, auto-compression detection will change behavior a bit, not so much compatibility.


lucaderi:
I agree to increase the version

PR #400, opened 2020-09-01
reworked cc20

Logan007:
This pull request applies some changes to the ChaCha20 implementation. Some are for speed (in the 10 percent range). Another one applies to the key setup now using Pearson hashing instead of SHA. This is a first step in order to get independent from openSSL.

This pull request is a prerequisite for adding n2n's own basic ChaCha20 code one day for backup in case openSSL is not available or desirable (as already happened to AES).

Due to the changes applied to the key setup, this pull request breaks compatibility to current _dev_ with respect to ChaCha20 only.


Logan007:
I couldn't hold myself back to add a public domain implementation of ChaCha20 to the `src/cc20.c` file. That way, ChaCha20 will always be available even if compiled without openSSL 1.1 support.

It is not the fastest implementation, I will be working on it. Just take it as a start.

PR #401, opened 2020-09-02
increase packet version to 3

### LINKED ISSUE #375: This is not v2 anymore!
Because the latest edge is no longer compatible with the previous one

Logan007:
This pull request increases the packet version to 3. **All edges and supernodes of your network** need to be updated in order to be able to communicate (again).

Fixes #375 

_______

~~Also, an SSE related version correction is included to allow correct compilation on older CPUs (it was meant to go in a separate pull request, but, well, git and me, you know…): Sometimes, `__SSE3__` is indicated even though the CPU does not have SSE3-support. That's why we need to check for `__SSE4_1`__ instead.~~ _will go into another pull request_

### DIFF
--- a/include/n2n_wire.h
+++ b/include/n2n_wire.h
@@ -25,7 +25,7 @@
 #include <stdint.h>
 #endif
 
-#define N2N_PKT_VERSION                 2
+#define N2N_PKT_VERSION                 3
 #define N2N_DEFAULT_TTL                 2       /* can be forwarded twice at most */
 #define N2N_COMMUNITY_SIZE              16
 #define N2N_MAC_SIZE                    6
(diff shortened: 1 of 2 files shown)


fengdaolong:
I think there may be changes in the network protocol in the future, such as the realization of automatic voting for super nodes, although no one has a clear plan for the development of v3.0.

Logan007:
The latest changes broke a lot, so current _dev_ will not work with the version from a week ago anyway. So, we could change the version field now. I agree with you @fengdaolong, nobody knows when the next big bang will happen here. But at least from my side, I do not plan any compatibility breaking changes in the near future. Only some speed-ups… :smile:

Actually, I have drafted a few ideas for leveraging a true peer-to-peer character. On paper, it looked quite wild and in theory, it worked out quite well. Just coding would beyond my current abilities. Also, I did not overcome the flaw of **guaranteeing** to reach **all** nodes with a broadcast. But we can discuss, perhaps in the _Issue_ section? It will take me a while to write it all down. It could serve as a discussion starter.

fengdaolong:
I very much look forward to your writing your ideas for everyone to discuss, because your ideas are always refreshing and you are a very talented person.😄😄😄

PR #404, opened 2020-09-03
compilation fix: correct (S)SSE version

Logan007:
An SSE related version correction is included to allow correct compilation on older CPUs: Just learned about the difference between SSE3 and **S**SSE3…

(separated from #401)

### DIFF
--- a/include/cc20.h
+++ b/include/cc20.h
@@ -43,7 +43,6 @@ typedef struct cc20_context_t {
 
 typedef struct cc20_context {
   uint32_t keystream32[16];
-  uint32_t state[16];
   uint8_t key[CC20_KEY_BYTES];
 } cc20_context_t;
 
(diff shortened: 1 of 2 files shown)


Logan007:
Also, the counter is now limited to the leading 32 bit of the nonce, overflows do not increment the remaining 96 bit of the nonce. This behavior is in accordance with OpenSSL's [`evp_*`](https://www.openssl.org/docs/man1.1.1/man3/EVP_chacha20_poly1305.html) implementation and well following RFC7539 and RFC5116.

Reviewing the ChaCha20 code, I found some speed-up opportunities especially inside the SSE section. Those will follow at a later point in time.

Logan007:
Added some speed-ups to the SSE implementation: A minor one (keeping it all in SSE ~ 10 % speed up) and a huge one (making better use of the eight SSE registers by crypting two streams in parallel – does not fully double the speed because other helper registers now go to memory but totaling to an additional 50 % speed-up anyway).

This ChaCha20 SSE implementation now even beats openSSL's implementation on my i7 2860QM – probably due to openSSL's `evp_*` overhead.

Pondering AVX now…

lucaderi:
You've done a great job here

PR #408, opened 2020-09-05
added aes-ni support to Pearson hashing

Logan007:
This pull request adds hardware accelerated table look-up on **Intel** CPU if available: Pearson hashing as implemented here uses AES' s-box as look-up table. Thus, it could benefit from AES hardware acceleration if available. Most of all, header encryption will benefit.

Also, I removed the `LOW_MEM_FOOTPRINT` define and stuck the code for different hash sizes to the most promising alternative. By consequence, the 16-bit look-up table and especially its initialization calling `pearson_hash_init()` is mandatory 

I tried NEON (on Raspberry Pi) and SSE, both are slowing it down. On ARM, even the VTBL instructions did not help or I just have not been able to properly use them.

As Raspberry Pi's CPU does not provide the AES support, I have no opportunity to test ARM-AES hardware support, so I did not add it.


lucaderi:
@Logan007 If you want I can provide you access to a raspberry, accessible via n2n :-), if interested

Logan007:
Thank you for offering! I already own a Rapsberry Pi whose CPU unfortunately does not support the ARM-hardware crypto extension which has to be licensed by the manufacturer and obviously was too expensive or just out of Raspberry Foundation's focus. It hits all RPi models. So, ARM-AES support will not help the RPi models (and I can't test it on them).

I know that there are some ARMv8 CPUs out there that have this hardware crypto extension (especially for AES, there are other additions, too). One user mentioned the EspressoBin SBC (your collection does not comprise one of those by accident?), but also router boards or even mobiles could have it. It even is supported by C's intrinsics (just like the Intel ones, a bit different). AES cipher as well as the hashing would benefit – but I would not be able to blindly code it without testing it, so I try to find some of this kind of hardware. Remote access would be very sufficient, indeed. 

`echo | gcc -dM -E - -march=native` should list `__ARM_FEATURE_AES` (among a lot of others) if supported.

PR #410, opened 2020-09-06
made openssl & zstd support optional (in configuration)

Logan007:
Given that all ciphers are built-in now, this pull request opts out OpenSSL which from hereon has to explicitly be configured by using

`./configure --with-openssl`

I will adopt the documentation in the other pull request (#409) accordingly.

### DIFF
--- a/configure.seed
+++ b/configure.seed
@@ -25,26 +25,37 @@ else
   N2N_LIBS="-lzstd ${N2N_LIBS}"
 fi
 
-AC_CHECK_LIB([crypto], [AES_cbc_encrypt])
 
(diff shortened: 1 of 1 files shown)


skyformat99:
Very good. thanks.
And does the cmake file need to be modified such?

Logan007:
> And does the cmake file need to be modified such?

Probably yes. It has an AES OPTION section which seems to be always on. I am not sure about CMake syntax, never used it. Do you have a suggestion?

Logan007:
The same behavior now applied to ZSTD support which now has to explicitly be configured before compilation:

`./configure --with-zstd`

I will adopt the documentation in the other pull request (#409) accordingly.

Issue #366, opened 2020-08-16
Windows compilation error "config.h: No such file or directory"


siskai:
`Can you help me with this problem?`

D:\download\n2n-2.8>md build

D:\download\n2n-2.8>cd build

D:\download\n2n-2.8\build>cmake ..
-- Building for: Visual Studio 15 2017
-- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.18363.
-- The C compiler identification is MSVC 19.16.27035.0
-- The CXX compiler identification is MSVC 19.16.27035.0
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe
-- Check for working C compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: D:/download/n2n-2.8/build

D:\download\n2n-2.8\build>"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\amd64\MSBuild.exe" edge.vcxproj /t:Build /p:Configuration=Release
Microsoft (R) Build Engine verze 15.9.21+g9802d43bc3 pro .NET Framework
Copyright (C) Microsoft Corporation. Všechna práva vyhrazena.

Vytváření sestavení bylo zahájeno 16.8.2020 9:35:00.
Projekt D:\download\n2n-2.8\build\edge.vcxproj v uzlu 1 (počet cílů: Build).
[...]

PR #420, opened 2020-09-14
windows compilation fix

### LINKED ISSUE #415: compile error on Visual Studio 2015
image

Logan007:
This pull request removes the `inline` keyword from a wrapper function in the `src/speck.c` file which prevented compilation on MSVS 2015.

Fixes #415.

### DIFF
--- a/src/speck.c
+++ b/src/speck.c
@@ -664,8 +664,8 @@ static int speck_expand_key (const unsigned char *k, speck_context_t *ctx) {
 
 // this functions wraps the call to internal speck_ctr functions which have slightly different
 // signature -- ctx by value for SSE with SPECK_CTX_BYVAL defined in speck.h, by name otherwise
-inline int speck_ctr (unsigned char *out, const unsigned char *in, unsigned long long inlen,
-                      const unsigned char *n, speck_context_t *ctx) {
+int speck_ctr (unsigned char *out, const unsigned char *in, unsigned long long inlen,
+               const unsigned char *n, speck_context_t *ctx) {
 
   return internal_speck_ctr (out, in, inlen, n,
 #if defined (SPECK_CTX_BYVAL)


skyformat99:
thanks.
And can the other one be fixed together?   #366

Logan007:
I am not too confident with configuration – especially not on windows. I guess that there has been a reason for the nested `#ifdefs` in the `n2n.h` file distinguishing between visual C++ and other windows-cases.

The regular `config.h` file s not included with the source because it shall be generated during configuration. The user shall not be able to build without configuration.

Can we be certain that the change from #366 does not break anything? Will it still work as intended with CMake?

If agreed upon, we definitely should put it into a separate pull request in which we might also address the still unresolved CMake related AES/OpenSSL handling – a "config/cmake fix" pull request.

```C
#ifdef WIN32
  #ifdef _MSC_VER
    #include "config.h" /* Visual C++ */
  #else
    #include "win32/winconfig.h"
  #endif
   … other general windows (un)defs …
#else
  #ifndef CMAKE_BUILD
    #include "config.h"
  #endif
#endif
```

skyformat99:
image

You can look at  git's log history.
in fact, delete it without any impact.

Logan007:
I just do not understand why we have different windows config files. Also, I do not understand the difference between those Windows alternatives _MSC_ and _others_ and how they influence configuration or their configuration requirements. Also, I am absolutely not confident with CMake and its configuration requirements. As I do not have any opportunity to test on Windows nor can I really evaluate consequences of Windows-related changes, I better do not change anything in that field.

If you are sure about these changes, please feel free to provide a pull request.

skyformat99:
> I just do not understand why we have different windows config files. Also, I do not understand the difference between those Windows alternatives _MSC_ and _others_ and how they influence configuration or their configuration requirements. Also, I am absolutely not confident with CMake and its configuration requirements. As I do not have any opportunity to test on Windows nor can I really evaluate consequences of Windows-related changes, I better do not change anything in that field.
> 
> If you are sure about these changes, please feel free to provide a pull request.

ok, thanks.

VergLsm:
This should consult an expert @lucaderi

lucaderi:
I personally don't use cmake so I cannot comment on that

PR #427, opened 2020-09-19
removed jitter allowance for systemic packets

Logan007:
This pull request removes time jitter allowance for systemic packets. It hits header-encrypted packets/communities only and shall protect against deviation attempts caused by directly replayed REGISTER or REGISTER_SUPER packets.

Also, the documentation is supplemented accordingly.


lucaderi:
perhaps you should explain what is a systemic packet

Logan007:
Sure! I consider _systemic packets_ all packets that are required to build the p2p network such as SUPER_REGISTER, REGISTER, REGISTER_ACK… of which the source matters as their source IP address might be further used for internal socket data.

The payload carrying data packets of PACKET type and also PEER_INFO are considered _non-systemic_.

PR #428, opened 2020-09-20
consolidated documentation on tap interface configuration

Logan007:


Logan007:
Technically, n2n uses a TAP device (not a TUN device). So, the wording in edge's external message and the help text got adopted accordingly in this context, too, to avoid confusion about the doc which is about **TAP** device configuration.

Issue #403, opened 2020-09-03
It is recommended to standardize the name of each file


skyformat99:
For encrypted files, use  crypt_  as the header of the file name.

crypto_aes.c
crypto_aes.h
crypto_cc20.c
crypto_cc20.h
crypto_tf.h
crypto_tf.c
crypto_speck.h
crypto_speck.c

hash_pearson.h
hash_pearson.c

n2n_wire.h
n2n_wire.c

Compression related, using zip_ for the beginning of the file:
zip_lzoconf.h
zip_lzoconf.c
zip_lzodefs.h

zip_minilzo.h
zip_minilzo.c

Other files start with n2n_  :
n2n_portable_endian.h
n2n_portable_endian.c
n2n_random_num.h
n2n_random_num.c

The advantage of doing this is that the document is clear and easy to understand and improve.

Logan007:
Or maybe even more sub-folders?

skyformat99:
If  use a standardized file name, it might be simpler.

Logan007:
From [the comment](https://github.com/ntop/n2n/pull/431#issuecomment-696029707) in #431:
> #403 Could you please update this?

This requires a more overarching design decision. Project leadership should decide on general file structure and names. We have just seen the split-up in two subfolders `src/` and `include/`.

Just two remarks (in addition to the _folder_-one above):

- Renaming would impact all header includes.

- I would not go with `zip_*` which is a special compression algorithm family but the more general term `compression_*` or `compr_*`.

fengdaolong:
Since @Logan007  has a lot of time, I suggest that Logan007 have project management authority. What do you think? @lucaderi

Logan007:
Erm, I am not sure what your assumptions about my availability are but let me assure you that my time is limited – and will be even more in the future…

Also, I cannot handle git. I would make the repository explode! :wink:

fengdaolong:
But I have already voted for you.🤣🤣🤣

Logan007:
I have not heard about elections going on… :smile:

PR #431, opened 2020-09-21
merged edge's subnet parameter into address parameter

Logan007:
In a first attempt to clean-up command line a bit, edge's `-s` (subnet) was merged into `-a` (address) follwoing cidr-style `-a [static:|dhcp:]IP[/nn]` style (keeping /nn optional, defaulting to /24, removing `-s`).

For example, the new `-a 192.168.8.5/16` corresponds to the former `-a 192.168.8.5 -s 255.255.0.0` 

This saves one parameter and one line of help text.


skyformat99:
very good!
#403  Could you please update this?

Issue #433, opened 2020-09-21
Communities need new federation flag 


fcarli3:
The federation will be a regular community because it will contain all the registered supernodes as nodes. Despite this, we need to add a new `is_federation` flag to `sn_community` structure to check if the current community is the federation or a community of edges.

PR #437, opened 2020-09-22
Add new federation flag to community structure

fcarli3:
Communities have now a new `is_federation` flag to check if a specific community is the federation of supernodes. 
If `is_supernode == FEDERATION_ENABLED` then the current community is the federation, otherwise it's a regular community.
This pr solves #433.

### DIFF
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -63,8 +63,9 @@
 #define N2N_COMPRESSION_ID_ZSTD		3	/* set if '-z2' cli option is present, available only if compiled with zstd lib */
 #define ZSTD_COMPRESSION_LEVEL		7	/* 1 (faster) ... 22 (more compression) */
 
-/* Federation indicator */
-#define FEDERATION_ENABLED     1 
+/* Federation indicators */
+#define IS_FEDERATION        1
+#define IS_NO_FEDERATION     0 
 
 /* (un)purgeable community indicator (supernode) */
 #define COMMUNITY_UNPURGEABLE		0
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -64,8 +64,8 @@
 #define ZSTD_COMPRESSION_LEVEL		7	/* 1 (faster) ... 22 (more compression) */
 
 /* Federation indicators */
-#define IS_FEDERATION        1
-#define IS_NO_FEDERATION     0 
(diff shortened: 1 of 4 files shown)


Logan007:
Do you already want to initialize the new data field (whenever a new/regular community gets added) within this pull request? Or have you planned for that as soon as we start actually using this new data field?

fcarli3:
I've planned to initialize it when we will start to use it

lucaderi:
@fcarli3 I suggest extending this PR as @Logan007 pointed out with at least some initialization code. Then I don't think you need a 32 bit integer for this flag. Please review your PR

fcarli3:
I have reviewed my PR, initializing the new field and using a smaller data type for the flag

Logan007 (approved):
Looks good to me!

include/n2n_define.h
```diff
@@ -63,6 +63,10 @@
 #define N2N_COMPRESSION_ID_ZSTD		3	/* set if '-z2' cli option is present, available only if compiled with zstd lib */
 #define ZSTD_COMPRESSION_LEVEL		7	/* 1 (faster) ... 22 (more compression) */
 
+/* Federation indicators */
+#define IS_FEDERATION        1
```
lucaderi: Can you please define an enum datatype with all the possible states instead of using #defines ?

src/sn_utils.c
```diff
@@ -974,6 +974,7 @@ static int process_udp(n2n_sn_t * sss,
 	if(comm) {
 	  strncpy(comm->community, (char*)cmn.community, N2N_COMMUNITY_SIZE-1);
 	  comm->community[N2N_COMMUNITY_SIZE-1] = '\0';
+	  comm->is_federation = IS_NO_FEDERATION;
```
lucaderi: As this code is similar to sn.c perhaps is better to create an initialization function?

lucaderi (changes requested):

fcarli3:
I have added an enum data type for `is_federation` flag value and a function to initialise some common fields of a new community

include/n2n_define.h
fcarli3: I've added an enum datatype with the two values

src/sn_utils.c
fcarli3: I've created `comm_init()` function that intialises some fields of a new community

PR #446, opened 2020-10-02
added (de-)compression to benchmark tool

Logan007:
This pull request adds the compression algorithms to the benchmark tool (LZO1x and ZSTD if compiled).

As compression and decompression speeds extremely differ, it is implemented in a way that a simple `tools/benchmark` will test the compression whereas `tools/benchmark -d` will run the decompression only. This way, the result gets much more meaningful.

Encryption and (encryption & decryption) benchmark remains nearly untouched: two calls to *de*code functions were moved into the *de*cryption part.

The output has been enhanced by output data size (for ciphers: that includes the PACKET overhead) and its format has been slightly adopted.


Logan007:
Also added Pearson hashing to the benchmark tool as of its 32-bit version because I ponder to soon use the 32 bit version for header encryption check-summing (currently: 16 bit).

lucaderi:
Great contribution

PR #459, opened 2020-10-10
AES-NI speed-up

Logan007:
This pull request speeds-up **AES-NI** (x86 hardware accelerated AES) implementation in the following ways:

- CBC encryption is kept as much as possible in SSE code leading to an encryption speed-up in the one-percent-range (… already a few MB/s!)
- CBC decryption also is kept as much as possible in SSE and was parallelized resulting in the clearly-above-ten-percent-range (… a few more MB/s!)

Check your personal speed-up using `tools/benchmark` and especially `tools/benchmark -d` (combined encryption and decryption).

_P. S._ I found that we do not have a tool to measure decryption speed only… that might be worth an addition to the benchmark tool. Some other day… perhaps.

### DIFF
--- a/src/aes.c
+++ b/src/aes.c
@@ -365,48 +365,92 @@ int aes_ecb_encrypt (unsigned char *out, const unsigned char *in, aes_context_t 
 }
 
 
-#define fix_xor(target, source) *(uint32_t*)&(target)[0] = *(uint32_t*)&(target)[0] ^ *(uint32_t*)&(source)[0]; *(uint32_t*)&(target)[4] = *(uint32_t*)&(target)[4] ^ *(uint32_t*)&(source)[4]; \
-                                *(uint32_t*)&(target)[8] = *(uint32_t*)&(target)[8] ^ *(uint32_t*)&(source)[8]; *(uint32_t*)&(target)[12] = *(uint32_t*)&(target)[12] ^ *(uint32_t*)&(source)[12];
 
 
 int aes_cbc_encrypt (unsigned char *out, const unsigned char *in, size_t in_len,
                      const unsigned char *iv, aes_context_t *ctx) {
 
-  uint8_t tmp[AES_BLOCK_SIZE];
-  size_t i;
-  size_t n;
+  size_t n = in_len / 16; // number of blocks
+  in_len &= 15;           // remaining bytes
 
-  memcpy(tmp, iv, AES_BLOCK_SIZE);
+  __m128i ivec = _mm_loadu_si128((__m128i*)iv);
 
-  n = in_len / AES_BLOCK_SIZE;
-  for(i=0; i < n; i++) {
-    fix_xor(tmp, &in[i * AES_BLOCK_SIZE]);
-    aes_internal_encrypt(ctx, tmp, tmp);
-    memcpy(&out[i * AES_BLOCK_SIZE], tmp, AES_BLOCK_SIZE);
+  while(n) {
+    __m128i tmp = _mm_loadu_si128((__m128i*)in);
+    in += 16; n--;
+    tmp = _mm_xor_si128(tmp, ivec);
+
+    tmp = _mm_xor_si128       (tmp, ctx->rk_enc[ 0]);
+    tmp = _mm_aesenc_si128    (tmp, ctx->rk_enc[ 1]);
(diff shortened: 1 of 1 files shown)


src/aes.c
```diff
@@ -365,48 +365,129 @@ int aes_ecb_encrypt (unsigned char *out, const unsigned char *in, aes_context_t
...
+  in_len &= 15;           // remaining bytes
+
+  __m128i ivec = _mm_loadu_si128((__m128i*)iv);
+
+  while(n) {
+    __m128i tmp = _mm_loadu_si128((__m128i*)in);
+    in += 16; n--;
+    tmp = _mm_xor_si128(tmp, ivec);
+
+    tmp = _mm_xor_si128       (tmp, ctx->rk_enc[ 0]);
```
lucaderi, line 381: What do you think about using a loop?

src/aes.c
```diff
@@ -365,48 +365,129 @@ int aes_ecb_encrypt (unsigned char *out, const unsigned char *in, aes_context_t
...
+    __m128i tmp1 = _mm_loadu_si128((__m128i*)in); in += 16;
+    __m128i tmp2 = _mm_loadu_si128((__m128i*)in); in += 16;
+
+
+    __m128i old_in1 = tmp1;
+    __m128i old_in2 = tmp2;
+
+    n--; n--;
+
+    tmp1 = _mm_xor_si128       (tmp1, ctx->rk_dec[ 0]); tmp2 = _mm_xor_si128       (tmp2, ctx->rk_dec[ 0]);
```
lucaderi, line 430: Same loop question here

src/aes.c
```diff
@@ -365,48 +365,129 @@ int aes_ecb_encrypt (unsigned char *out, const unsigned char *in, aes_context_t
...
+    _mm_storeu_si128((__m128i*) out, tmp2); out += 16;
   }
 
-  return n * AES_BLOCK_SIZE;
+  if(n) {
+    __m128i tmp = _mm_loadu_si128((__m128i*)in);
+    __m128i old_in = tmp;
+    in += 16; n--;
+
+    tmp = _mm_xor_si128       (tmp, ctx->rk_dec[ 0]);
```
lucaderi, line 530: and here

src/aes.c
Logan007: Haven't thought about those for a while! :wink:  – Will replace.

src/aes.c
Logan007: – Will replace.

src/aes.c
Logan007: Not here, this handles the last block **if** present which in turn renders the increments and decrements even completely superfluous. – Will remove.

Logan007:
Curiosity got triggered… if we offer 4 parallel AES-NI rails (for data left to encrypt of 4 or more blocks), `tools/n2n-benchmark -d` shows another 5 to 10 percent performance gain for "encryption plus decryption" **on some CPUs**¹. That can only be explained with less data dependencies between the single instructions (the rails are interleaved) inside the CPU pipeline.

______
¹ i7 2860QM significantly benefits while i7 7500U seems to be almost indifferent…

src/aes.c
Logan007: With the new commit (4 rails) the `if` still fits in better here as the following code is passed only once (1 block left) or never (0 blocks left).

Logan007:
Applying a slight modification (hacky: skip the encryption by `if (!num_packets){…}`, then calling it with `-d`) to the benchmark tool, I measured the decryption speed only.

Decryption speed using current _dev_ is around 1,166 MB/s (much higher than expected for single lane, not sure if the compiler is able to better optimize the code or less dependency). However, this pull request lifts it up to 2,118 MB/s. Decryption speed on i7 7500U (encryption around 1,101 MB/s) jumps from 2,358 MB/s to Even 3,177 MB/s!

Just for the sake of completeness, encryption speed is around 623 MB/s on i7 2860QM.  CBC indeed is so much faster to decrypt than to encrypt.

I am going to add a separate decryption speed output to the benchmark tool (in a separate pull request). Such amazing numbers deserve to be displayed.

Hardware acceleration rocks!

lucaderi:
Great work!

Issue #451, opened 2020-10-03
Which supernodes intialize the federation list


fcarli3:
We could initialize the federation list using a parameter like edge's `-l` that adds several supernodes to the edge if given several times.  As the supernode has already `-l` parameter, we might use another letter (e.g. change the supernode's current `-l` in  `-p` to match the edge's port parameter `-p`).

Additionally, we would need to mark the supernodes given by command line as "unpurgeable", they may never be purged as they are considered anchors. The peer_info needs a corresponding flag (e.g. `is_purgeable` like the field of a community) and the purge logic needs a slight update towards this idea: "Never purge those nodes even if older than the limit and always retry those for re-registration, no matter how old they are"

fcarli3:
@Logan007  I've thought about how implement the function that adds the supernode to the federation list. How could we convert a supernode's name ? Because my idea is to do a function that takes the string, splits it in 'supernode' and 'port' and then fulfills a new `peer_info` structure for the provided supernode. That way, we will need to know at least `mac_addr`, `dev_addr` and `sock`

Logan007:
I guess the edge's current `-l` should already do it for `-l <ip:port>`.

But you have just made a great point here: How to add the MAC address to the peer_list (at the edge) if we only pass IP and port?!

We probably need to leave the MAC for the anchor supernodes (those passed through `-l`) reset (zero MAC) and treat them in a special way (they also wear that "unpurgeable flag" which could be considered synonym here for anchor – or shall we use a zero MAC for that purpose?).

Inserting freshly received supernodes from REGISTER_SUPER_ACK needs special handling then because supernodes in the list could have zero-MAC. How to solve this? In case of zero-MACs present, we might need to compare the socket as well and replace the zero-MACs as soon as we receive their data.

Or: We accept that we might have anchor supernodes twice in our list: One with zero MAC (from command line) and another one with real MAC (received later from REGISTER_SUPER_ACK from other supernodes).

[...]

PR #460, opened 2020-10-10
Initialize the federation of supernodes

fcarli3:
Supernode has now a new parameter `-l <supernode:port>` that takes IP or name of an anchor known supernode that will be insert in the federation if not present yet. The original `-l <local port>` parameter has changed in `-p <local port>`.

I've used `supernode2sock()` function to convert a supernode name and obtain its IP, its port and its socket. This isn't a new function, I have only renamed the original `supernode2addr()` function and declare it in `include/n2n.h`.

I've created a new `add_sn_to_federation_by_mac_or_sock()` function that search a supernode in the federation list by MAC or socket (a new anchor supernode could have zero MAC but it could be already in the federation, we can find it searching by socket). 

I've also changed the type of `federation` field in `n2n_sn_t` from `char *` to `struct sn_community *`. That way, we always have a pointer to the federation of supernodes and we don't have to search it every time we need it. 

This PR solves #451.

### DIFF
--- a/include/n2n_wire.h
+++ b/include/n2n_wire.h
@@ -169,8 +169,6 @@ typedef struct n2n_REGISTER_SUPER_ACK {
   uint8_t              num_sn;         /**< Number of supernodes that were send
 					* even if we cannot store them all. If
 					* non-zero then sn_bak is valid. */
-  n2n_sock_t           sn_bak;         /**< Socket of the first backup supernode */
-  n2n_mac_t	       mac_addr;
 } n2n_REGISTER_SUPER_ACK_t;
 
 
--- a/src/sn.c
+++ b/src/sn.c
@@ -236,11 +236,6 @@ static int setOption(int optkey, char *_optarg, n2n_sn_t *sss) {
     size_t length;
     int rv;
 
-    if(_optarg == NULL){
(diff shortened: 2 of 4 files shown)


src/sn.c
```diff
@@ -327,7 +366,7 @@ static int setOption(int optkey, char *_optarg, n2n_sn_t *sss) {
 static const struct option long_options[] = {
 					     {"communities", required_argument, NULL, 'c'},
 					     {"foreground",  no_argument,       NULL, 'f'},
-					     {"local-port",  required_argument, NULL, 'l'},
+					     {"local-port",  required_argument, NULL, 'p'},
```
lucaderi, line 369: For the future please avoid changing CLI conventions

include/n2n_wire.h
```diff
@@ -169,8 +169,6 @@ typedef struct n2n_REGISTER_SUPER_ACK {
   uint8_t              num_sn;         /**< Number of supernodes that were send
 					* even if we cannot store them all. If
 					* non-zero then sn_bak is valid. */
-  n2n_sock_t           sn_bak;         /**< Socket of the first backup supernode */
```
lucaderi, line 172: This caused encode_REGISTER_SUPER_ACK to fail

Issue #463, opened 2020-10-12
Slow rate and packet loss in old version of openwrt


small-5:
The latest commit has problems of slow speed and packet loss in the old version of openwrt (kernel version 3.14.79, uclibc), as well as commit 644ea6c82ab3671a6b9aed4494e07b38a823308b. Tried to change the encryption method, but did not solve it. It felt like it was caused by MTU. I didn't know how to find out the cause. There was no problem with the [commit 73fa50a37dc23a91a130335583494c97a2355a95](https://github.com/ntop/n2n/commit/73fa50a37dc23a91a130135583494c97a2355a95) a long time ago, but there were too many changes submitted in the middle. I couldn't find out what the change caused. I hope it can be fixed,Thank you!

Logan007:
If you compiled yourself, would a binary search towards the breaking commit be an option for you? That could really help to look into this.

If you suspect MTU, you might want to play around with edge's `-M` cli parameter.

small-5:
I have checked the MTU of the virtual network card, which is the default 1290. But the symptoms are similar to the loss of large packets, because if I just ping the virtual IP, there will be basically no packet loss, but when there is data transmission, It will start to lose packets, and the speed is significantly slower than the previous version.

Logan007:
Does `sudo ping -f <IP address>` also show packet loss?

And what about different packet sizes `sudo ping -f -s <size> <IP address>`? Can you nail it to a certain packet size?

small-5:
I used the -H parameter, I will test it later according to your suggestion and tell you the result.Thank you!

Logan007:
Then, please also without `-H`.

small-5:
After testing, it is found that this problem is related to the -H parameter. Enabling the -H parameter will cause packet loss and slow speed problems. This problem has nothing to do with the encryption method. And when some edge nodes enable the -H parameter, and some edge nodes do not enable the -H parameter, it will cause a segmentation fault problem on the supernode, just like my previous feedback: https://github.com/ntop/ n2n/issues/453.

small-5:
After testing, it is found that this problem is related to the -H parameter. Enabling the -H parameter will cause packet loss and slow speed problems. This problem has nothing to do with the encryption method. And when some edge nodes enable the -H parameter, and some edge nodes do not enable the -H parameter, it will cause a segmentation fault problem on the supernode, just like my previous feedback: https://github.com/ntop/n2n/issues/453.

small-5:
[...]

PR #464, opened 2020-10-13
Fix segfault in issue #463

fcarli3:
This PR fixes segfault error in issue #463


small-5:
After a simple test, the segfault should be solved, thank you!

lucaderi:
Thank you

PR #470, opened 2020-10-19
added pure decryption speed output

### LINKED ISSUE #462: benchmark tool does not show decryption-only speed
As seen [here](https://github.com/ntop/n2n/pull/459#issuecomment-706592268), encryption and decryption speed can differ a lot.

I plan to adopt the benchmark tool to always output encryption speed and (pure) decryption speed, thus getting rid of `-d` parameter as well as the "mixed" (encryption plus decryption) speed as it smears encryption and decryption speed too much.

Logan007:
This pull request changes the output of `tools/n2n-benchmark` by adding the decryption-only speed. It removes the mixed speed (of encryption and decryption) along with the `-d` command line option as it smears especially the cases in which decryption is much faster than encryption such as  hardware-accelerated AES.

Fixes #462

### DIFF
--- a/tools/benchmark.c
+++ b/tools/benchmark.c
@@ -18,7 +18,7 @@
 
 #include "n2n.h"
 
-#define DURATION     3   // test duration per algorithm
+#define DURATION     2.5   // test duration per algorithm
 
 
 /* heap allocation for compression as per lzo example doc */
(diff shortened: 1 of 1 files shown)


lucaderi:
Thank you

PR #474, opened 2020-10-21
enabled gratuitous arp packets

Logan007:
This pull request fixes and enables the code for sending gratuitous ARP packets. Those are required to immediately announce eventually changed MAC addresses throughout the network.

Also, the supernode better handles changed MAC addresses by not creating a new entry but using the old one if not purged yet.

MAC addresses are assigned randomly if not assigned at start-up (`-m`). In case of a quickly performed edge restart, the remaining edges have not had the new MAC-IP address relation in their ARP tables yet, thus the gratuitous ARP packet.

Plans for future extension are to send gratuitous ARP packets only if required (the supernode can detect and inform the edge via REGISTER_SUPER_ACK) whereas the current version sends them along with every REGISTER_SUPER.

### DIFF
--- a/src/edge.c
+++ b/src/edge.c
@@ -647,59 +647,6 @@ static int loadFromFile(const char *path, n2n_edge_conf_t *conf, n2n_tuntap_priv
 
 /* ************************************** */
 
-#if defined(DUMMY_ID_00001) /* Disabled waiting for config option to enable it */
-
-static char gratuitous_arp[] = {
-				0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, /* Dest mac */
-				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, /* Src mac */
-				0x08, 0x06, /* ARP */
-				0x00, 0x01, /* Ethernet */
-				0x08, 0x00, /* IP */
-				0x06, /* Hw Size */
-				0x04, /* Protocol Size */
-				0x00, 0x01, /* ARP Request */
-				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, /* Src mac */
-				0x00, 0x00, 0x00, 0x00, /* Src IP */
-				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, /* Target mac */
-				0x00, 0x00, 0x00, 0x00 /* Target IP */
-};
-
-/* ************************************** */
-
-/** Build a gratuitous ARP packet for a /24 layer 3 (IP) network. */
-static int build_gratuitous_arp(char *buffer, uint16_t buffer_len) {
-  if(buffer_len < sizeof(gratuitous_arp)) return(-1);
-
-  memcpy(buffer, gratuitous_arp, sizeof(gratuitous_arp));
-  memcpy(&buffer[6], device.mac_addr, 6);
-  memcpy(&buffer[22], device.mac_addr, 6);
-  memcpy(&buffer[28], &device.ip_addr, 4);
-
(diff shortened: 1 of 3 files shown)


src/edge.c
```diff
@@ -647,59 +647,6 @@ static int loadFromFile(const char *path, n2n_edge_conf_t *conf, n2n_tuntap_priv
...
-
-static char gratuitous_arp[] = {
-				0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, /* Dest mac */
-				0x00, 0x00, 0x00, 0x00, 0x00, 0x00, /* Src mac */
-				0x08, 0x06, /* ARP */
-				0x00, 0x01, /* Ethernet */
-				0x08, 0x00, /* IP */
-				0x06, /* Hw Size */
-				0x04, /* Protocol Size */
-				0x00, 0x01, /* ARP Request */
```
Logan007, line 660: As far as I know, gratuitous ARP is a **reply** and not a request – changed to `0x02`.

src/edge.c
```diff
@@ -647,59 +647,6 @@ static int loadFromFile(const char *path, n2n_edge_conf_t *conf, n2n_tuntap_priv
...
-  if(buffer_len < sizeof(gratuitous_arp)) return(-1);
-
-  memcpy(buffer, gratuitous_arp, sizeof(gratuitous_arp));
-  memcpy(&buffer[6], device.mac_addr, 6);
-  memcpy(&buffer[22], device.mac_addr, 6);
-  memcpy(&buffer[28], &device.ip_addr, 4);
-
-  /* REVISIT: BbMaj7 - use a real netmask here. This is valid only by accident
-   * for /24 IPv4 networks. */
-  buffer[31] = 0xFF; /* Use a faked broadcast address */
```
Logan007, line 680: In gratuitous ARP, the IP address is just the host's address. Thus, the originally assumed issue _use a real netmask_ is not an issue and having the host bits all set does not seem to be accepted by other hosts – removed.

Logan007:
As mentioned above, one aim is to avoid unnecessary network load, i.e. only send an gartuitous ARP packet if required.

For now, I see two approaches:

1. Send a gratuitous ARP packet (or two) at every edge start-up (and remain silent afterwards).

2. Let the supernode evaluate, if the MAC has changed (that would happen by the code already changed in `sn_utils.c:333`) and let the edge know that it has to send out gratuitous ARP packet along with REGISTER_SUPER_ACK which needs one indicator byte then as I can't use the general flags (this is REGISTER_SUPER_ACK specific).

Every gratuitous ARP packet contains 42 bytes which will sum up to 92 bytes including encryption overhead (IV) and PACKET format header. Broadcasting it unconditionally at every edge start-up will result in _n² × 92_ bytes always sent-out network load overhead where _n_ denotes the number of edges.

On the other hand, one additional byte per REGISTER_SUPER_ACK being sent out every 20 seconds would lead to _n × t / 20 × 1_ bytes as permanent additional network load with _t_ being the up-time.

A comparison shows that in total, the second solution puts more load on the network as soon as (average) edge up-time exceeds roughly half an hour for every edge present in the network, that is _t > n × 1,840_ seconds. If we dare to assume an average edge up-time of at least a week in permanent setups, only networks containing more than _329_ edge nodes would start to benefit from solution number 2.

In most cases, I would exclude networks as big as this and therefore implement number 1.

This does not mean that bigger networks will break down under the pressure – remember that we are talking about 98 byte-sized packets.

Logan007:
I have just removed the draft marking and will implement the above outlined idea (send only once) in a separate pull request.

Issue #473, opened 2020-10-21
Edges Identification in One Community


wlt-cortex:
Hello everyone:

I am a new one to the n2n project just approaching yesterday. And I want to know that how can I identify the edge node in some specific community. Here is my case:

Suppose I have two machines named `dev-sever` and `office` in different places and networks. The two are connected with `n2n` in community name: `labs` using DHCP, and then I wonder to know if there are some parameters in the startup of edge node to register human-readable name into supernode so that I can find which name is the right one when listing the accessible peers in edge node or supernode.

> Another proposed problem is that the query command: `nc -u localhost 5644` is out-dated sometimes. I don't know this situation is normal or maybe refresh manually or just wait some time.

some notes:
    
+ I don't find early issues describing the same question.
+ I find some out-dated scripts and `n2n` v1 seems to have the `-d` parameter indicating that edge device name? And now the parameter has been removed yet.

```shell
# On supernode
dnf install dhcp-server dhcp-client n2n -y
systemctl disable dhcpd --now
supernode -f -l 9090 -v

# On supernode/one edge
edge -f -d edge0 -r -a 192.168.1.0 -c foo -k bar -l localhost:9090 -v -A
cat > /etc/dhcp/dhcpd.conf <<EOF
subnet 192.168.1.0 netmask 255.255.255.0 {
        range 192.168.1.10 192.168.1.100;
}
EOF
dhcpd -f edge0

# On all other edges
dnf install dhcp-client n2n -y
edge -f -d edge0 -r -a dhcp:0.0.0.0 -c foo -k bar -l 116.203.50.42:9090 -v -A -m $(echo $(hostname)|md5sum|sed 's/^\(..\)\(..\)\(..\)\(..\)\(..\).*$/02:\1:\2:\3:\4:\5/') # To get the same MAC address and thus the same IP address every time
pkill -9 dhclient; dhclient edge0
```
referenced by [here](https://gitlab.com/snippets/1895685).

+ I know the `static IP` can solve the problem but it's awkward I thought, I need to manually set the static IP and try to avoid the IP conflicting. I have set lots of duplicated IPs when testing early and seems that the supernode will accept this request!!! strange.

So could you tell me how can I do?

Logan007:
Would the MAC address (`-m`) be helpful? Hex-only but readable. However, you still have to avoid conflicts then.

An internal edge-name is not available (yet).

[...]

PR #475, opened 2020-10-22
Extra Device Description in P2P Connection and Register to Super Node

betabeauty:
According to the discussion in #473, I have added the extra description field (named `dev_desc` in source code) and tested some simple case by myself. There are three points worth noting generally:

### 1. Command Parser

I added the extra command line parser: `-I` in the edge option parser, which refers to the `community_name` field parsing and checking logic. The `dev_desc` is added into the `conf` struct like `community_name` and generated by the **machine hostname** by default, non-over the max-length 16 or will be truncated. Users can set another server name (aka description, hint) for whatever they want with the `-I {server name}` option.

### 2. Edge Node Register into Super Node

- Both the supernode source and edge files are changed via the struct: `n2n_REGISTER_SUPER`. Add an extra `dev_desc` field in the struct and the corresponding struct encoder and decoder in `src/wire/c`.
- Change the exposing field via super node's manage port, show the extra device name in the details table.

### 3. Edge Node Register into Edge Node

- As section 2 indicates, peer to peer connection is via the struct: `n2n_REGISTER`.
- Show the extra device name in the details table from the edge node's manage port.

Some questions:

the two functions: `register_with_new_peer` and `check_peer_registration_needed` in `edge_util.c` are passing a lot of edge node's parameters recursively. And why not pack it into the `n2n_REGISTER` struct to make it more elegant, currently, I have to add an extra parameter: `dev_desc` in these functions. I have watched the `n2n_REGISTER` has an **unused** filed: `sock` coincidently, to be honest, I think it's better to unify the two process logics.

```
typedef struct n2n_REGISTER
{
  n2n_cookie_t         cookie;         /**< Link REGISTER and REGISTER_ACK */
  n2n_mac_t            srcMac;         /**< MAC of registering party */
  n2n_mac_t            dstMac;         /**< MAC of target edge */
  n2n_sock_t           sock;           /**< REVISIT: unused? */
  n2n_ip_subnet_t      dev_addr;       /**< IP address of the tuntap adapter. */
  n2n_desc_t           dev_desc;       /**< Hint description correlated with the edge */
} n2n_REGISTER_t;
```

Welcome to review.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,8 @@ cmake_minimum_required(VERSION 2.6)
 include(CheckFunctionExists)
 SET(CMAKE_VERBOSE_MAKEFILE ON)
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
+
 # N2n release information
 set(N2N_VERSION "2.9.0")
 set(N2N_OSNAME ${CMAKE_SYSTEM})
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -197,6 +197,7 @@ typedef char n2n_sn_name_t[N2N_EDGE_SN_HOST_SIZE];
 struct peer_info {
   n2n_mac_t        mac_addr;
   n2n_ip_subnet_t  dev_addr;
+  n2n_desc_t       dev_desc;
   n2n_sock_t       sock;
   int              timeout;
   uint8_t          purgeable;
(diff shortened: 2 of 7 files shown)


fengdaolong:
Most people do not add a description field when running an edge, and if the number of online edges is large, the description field may be repeated a lot, and finally have to be distinguished by a unique MAC address. I think the description field cannot be reported by the peer edge actively, but is filled in by the users of the currently running edge according to their own preferences, and only the users of the current edge know it.
We need to fully discuss #473, @Logan007  what do you think?

Logan007:
As I said over at #473, I still wonder why anyone would specify a clear text description in favor of a MAC address, but that's maybe just me :wink: So, basically, I have no objections against a textual description. It could even serve as (an additional) source for automated MAC generation!

I still am bemoaning the frequently repeating transmission on a regular basis (also see our discussion at #367). Does an optional textual identifier which probably does not change once an edge is up belong to the registration process or wouldn't that be better part of the peer information to be queried?

lucaderi:
I agree that propbably the PR does not solve a super-popular isse, however I don't see anything wrong with it.

lucaderi:
Fixes #473

PR #478, opened 2020-10-26
Update REGISTER_SUPER_ACK handling on edge

### LINKED ISSUE #449: Handle REGISTER_SUPER_ACK payload
An edge has to analyze REGISTER_SUPER_ACK payload in order to find out which supernodes are available. After the edge decodes the message, it could use the existing code in `re_register_and_purge_supernodes()` (`src/sn_utils.c`) to check alive supernodes and add them in a specific data structure.

fcarli3:
This PR modifies and updates how edges handle supernodes. First I changed the array of supernodes into an hash list of supernodes, like we have in `n2n_sn_t` structure. That way, we can know other many informations about supernodes, such as MAC, socket and so on. I also added a `struct peer_info *curr_sn` to `n2n_edge_t` to mantain the current supernode which an edge is registered to. 

Regards REGISTER_SUPER_ACK payload handling, I used the same approach as in `src/sn_utils.c`, in fact I moved the function `add_sn_to_federation_by_mac_or_sock()` from this file to `src/n2n.c` and I renamed it in `add_sn_to_list_by_mac_or_sock()`. So when a packet arrives, we decode it and we iterate through its payload to update informations of supernodes in the list or add supernodes that aren't already in the list. 

I also added a feature that I have to complete yet because now it works only in the common scenario with one supernode. That feature implements a selection strategy from edges: they have the list of available supernode and with a strategy based on a `ping_time`, an edge will choose the best supernode in the ordered list, that is sorted calling `sort_supernodes()` in `run_edge_loop()`.

Anyway, @Logan007 and I tested many scenario with supernodes and edges:
- edges connected to a supernode
-  many supernodes connected to one using `-l` parameter (e.g. A, B->A, C->A, D->A)
- a "chained" scenario (e.g. A->B, B->C, C->D, D->A ...)
- a supernode suddenly is stopped and restarted after a while
- a supernode enters the federation later after the others
- scenario where we mixed cases above

All tests above work well, we checked if MACs of supernodes and edges are handled and shared properly through the federation using `netcat -u localhost <mgmt_port>`.

This PR fixes #449 .

### DIFF
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -198,11 +198,13 @@ struct peer_info {
   n2n_mac_t        mac_addr;
   n2n_ip_subnet_t  dev_addr;
   n2n_sock_t       sock;
+  n2n_cookie_t     last_cookie;
   int              timeout;
   uint8_t          purgeable;
   time_t           last_seen;
   time_t           last_p2p;
   time_t           last_sent_query;
+  time_t           ping_time;
   uint64_t         last_valid_time_stamp;
   char             *ip_addr;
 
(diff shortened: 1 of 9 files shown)


Logan007:
It is important to point out that (so far) **regular usage** (providing _one_ supernode via `-l` to edges and _no additional_ supernodes to the one supernode) **works fine**.

However, everything beyond that is still somewhat under development and at user's own risk.

PR #479, opened 2020-10-29
rpm: fix installation warning

gsanchietti:
Avoid cosmetic error when installing the package.

Extract from yum output:
```
Creating n2n group
/var/tmp/rpm-tmp.S67n4H: line 8: cho: command not found
  Installing : n2n-2.9.0-746.x86_64   
```


lucaderi:
Thank you

PR #484, opened 2020-11-03
refrained from calling gettimeofday() too often during benchmark

### LINKED ISSUE #483: n2n slower on AMD Ryzen than on Intel CPU?
I had a chance to test n2n on a computer with AMD's Ryzen Renoir 4750U CPU. I ran n2n's benchmark utility and was somewhat disappointed. Even my good old i7 2860QM outperformed it. I also saw null_transform and compression showing unusually low numbers: It all ranged between **145 MB/s** (twofish enc) and **450 MB/s** (null transform) if compiled with hardware support and optimization `CFLAGS="-O3 -march=native"`.

While other benchmarks (`openssl speed -evp aes128`, `cryptsetup benchmark`, `lzbench -ezstd`, …) showed more expected results on that AMD CPU, I am not sure what part of n2n's code makes it run so much slower on AMD than on Intel. I tried several times and once even used a binary from the Intel computer to rule out platform compiling issues.

Can anyone confirm? I tried on current Arch Linux using gcc.

Logan007:
Some computers seem to be very slow on calling `gettimeofday()`, see #483. Thus, this pull request changes benchmark behavior in a way that does not call `gettimeofday()` too often.

Fixes #483

### DIFF
--- a/tools/benchmark.c
+++ b/tools/benchmark.c
@@ -155,11 +155,11 @@ static void run_compression_benchmark() {
       printf("\n\t compression error\n");
       exit(1);
     }
+    num_packets++;
     if (!(num_packets & PACKETS_BEFORE_GETTIME)) {
       gettimeofday( &t2, NULL );
       tdiff = ((t2.tv_sec - t1.tv_sec) * 1000000) + (t2.tv_usec - t1.tv_usec);
     }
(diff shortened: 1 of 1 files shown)


skyformat99:
very good

PR #486, opened 2020-11-04
Packet forwarding between supernodes

### LINKED ISSUE #481: Cross-supernode information/packet forwarding
Supernode needs to broadcast/forward certain packets to other supernodes (those for which it does not find a locally registered edge). Also, supernode must accept such incoming packets and handle them properly (e.g. take action, do not forward to another supernode if already forwarded).

fcarli3:
Supernodes need to broadcast/forward certain packets (e.g. PACKET, REGISTER, QUERY_PEER...) in order to really fulfill the purpose for which the federation of supernodes was created: communication between supernodes and information sharing. The receiving supernodes must decide if they received from an edge or a supernode (checking `from_supernode` flag) and handle it accordingly in case they do not have any information about the requested/destination edge, that is drop or forward (including setting the `from_supernode` flag).

I added a parameter to `try_broadcast` and `try_forward` to check if the receiving packet is from an edge or another supernode. Moreover, I modified the body of `try_broadcast` because we have to make sure that a broadcast reaches the other supernodes and edges connected to them. As I said lines above, `try_broadcast` needs a `from_supernode` parameter: if set do forward to edges of community only. If unset forward to all locally known nodes and all supernodes.

Regarding QUERY_PEER packets, I used the following approach:
- packet from edge: check if peer is known (registered to that supernode). If it's registered I used the regular handling as before, otherwise set `from_supernode` and socket flags in packet common section, add socket to QUERY_PEER and forward it to other supernodes.
- packet from supernode: regular handling as before.

I also did a cleanup of the code, correcting wrong indentions and renaming some macro with more self-explicative names.

This PR fixes #481.

### DIFF
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -32,23 +32,25 @@
 /* Max available space to add supernodes' informations (sockets and MACs) in REGISTER_SUPER_ACK
  * Field sizes of REGISTER_SUPER_ACK as used in encode/decode fucntions in src/wire.c
  * REVISIT: replace 255 by DEFAULT_MTU as soon as header encryption allows for longer packets to be encrypted. */
-#define MAX_AVAILABLE_SPACE_FOR_ENTRIES \
+#define REG_SUPER_ACK_PAYLOAD_SPACE \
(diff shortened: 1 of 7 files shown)


Logan007:
This PR along with #480 allows for a working fail-over scenario with several supernodes.

lucaderi:
@fcarli3 Can you please fix conflicts so I can merge the code?

fcarli3:
Conflicts should be solved now.

Issue #495, opened 2020-11-11
compile error  in Visual Studio 2015


skyformat99:
1) git Clone the latest version of the code
2) Use Cmake to generate the VS2015 project
3)compile error  in Visual Studio 2015:

![image](https://user-images.githubusercontent.com/10938976/98751300-d65ac080-23fa-11eb-861d-ccb7a4e3b978.png)

skyformat99:
close it.

PR #496, opened 2020-11-11
Update sn_utils.c

skyformat99:
fix  compile error in Visual Studio 2015 
ref  #495

### DIFF
--- a/src/sn_utils.c
+++ b/src/sn_utils.c
@@ -1293,8 +1293,8 @@ static int process_udp(n2n_sn_t * sss,
       }
     }
 
-    tmp_sock = (void *)dec_tmpbuf;
-    tmp_mac = (void*)dec_tmpbuf + sizeof(n2n_sock_t);
+    tmp_sock = (n2n_sock_t *)dec_tmpbuf;
+    tmp_mac = (n2n_sock_t *)dec_tmpbuf + sizeof(n2n_sock_t);
 
     for(i=0; i<ack.num_sn; i++) {
       skip_add = SN_ADD;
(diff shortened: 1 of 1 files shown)


src/sn_utils.c
```diff
@@ -1293,8 +1293,8 @@ static int process_udp(n2n_sn_t * sss,
       }
     }
 
-    tmp_sock = (void *)dec_tmpbuf;
-    tmp_mac = (void*)dec_tmpbuf + sizeof(n2n_sock_t);
+    tmp_sock = (n2n_sock_t *)dec_tmpbuf;
+    tmp_mac = (n2n_sock_t *)dec_tmpbuf + sizeof(n2n_sock_t);
```
Logan007, line 1297: I think this change will lead to wrong addresses (pointer arithmetic).

src/sn_utils.c
```diff
@@ -1305,8 +1305,8 @@ static int process_udp(n2n_sn_t * sss,
       }
 
       /* REVISIT: find a more elegant expression to increase following pointers. */
-      tmp_sock = (void*)tmp_sock + REG_SUPER_ACK_PAYLOAD_ENTRY_SIZE;
-      tmp_mac = (void*)tmp_sock + sizeof(n2n_sock_t);
+      tmp_sock = tmp_sock + REG_SUPER_ACK_PAYLOAD_ENTRY_SIZE;
```
Logan007, line 1308: I think this change will lead to wrong addresses (pointer arithmetic).

src/sn_utils.c
```diff
@@ -1305,8 +1305,8 @@ static int process_udp(n2n_sn_t * sss,
...
+      tmp_mac = tmp_sock + sizeof(n2n_sock_t);
```
Logan007, line 1309: I think this change will lead to wrong addresses (pointer arithmetic).

Logan007 (changes requested):
Do you get an error or a warning in VS2015?

PR #497, opened 2020-11-11
Update edge_utils.c

skyformat99:
fix  compile error in Visual Studio 2015
ref  #495

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -2045,8 +2045,8 @@ void readFromIPSocket(n2n_edge_t * eee, int in_sock) {
                   }
 
                   /* REVISIT: find a more elegant expression to increase following pointers. */
-                  tmp_sock = (void*)tmp_sock + REG_SUPER_ACK_PAYLOAD_ENTRY_SIZE;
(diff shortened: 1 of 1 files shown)


Logan007:
Please see my remarks over at #496.

skyformat99:
> Please see my remarks over at #496.

ref #495,
If it is not fix, an error occurs in vs2015.
and  The fixed code  also points to the first address,  so it doesn't change the value of the expression.

lucaderi:
I personally hate using void* and moving pointers with ++. Types should be used instead. Please rework this code

fengdaolong:
> I personally hate using void* and moving pointers with ++. Types should be used instead. Please rework this code

I also think it should not be written like this.

Even if it is changed to `(char*)tmp_sock`, it is better than `(void*)tmp_sock`.

Logan007:
> Even if it is changed to (char*)tmp_sock, it is better than (void*)tmp_sock.

We probably need to do that, or we need to find fully typed expression which I fear might be a bit more unreadable. Let me think. I am not sure if the code as it has changed up to now will lead to the correct addresses.

PR #508, opened 2020-11-17
Deal with warnings at `network_traffic_filter.c` 

joshuafc:


src/network_traffic_filter.c
```diff
@@ -364,17 +402,23 @@ n2n_verdict filter_packet_from_tap(network_traffic_filter_impl_t *filter, n2n_ed
...
+network_traffic_filter_t *create_network_traffic_filter();
+
 network_traffic_filter_t *create_network_traffic_filter() {
   network_traffic_filter_impl_t *filter = malloc(sizeof(network_traffic_filter_impl_t));
 
   memset(filter, 0, sizeof(network_traffic_filter_impl_t));
-  filter->filter_packet_from_peer = filter_packet_from_peer;
-  filter->filter_packet_from_tap = filter_packet_from_tap;
-  return filter;
+  filter->filter_packet_from_peer = (filter_packet_from_peer_func_t)filter_packet_from_peer;
```
lucaderi: Why do you need a cast here? Can you please define filter_packet_from_peer of type filter_packet_from_peer_func_t instead of casting?
Same applies to filter_packet_from_tap

src/network_traffic_filter.c
```diff
@@ -364,17 +402,23 @@ n2n_verdict filter_packet_from_tap(network_traffic_filter_impl_t *filter, n2n_ed
...
+  filter->filter_packet_from_tap = (filter_packet_from_tap_func_t)filter_packet_from_tap;
+  return (network_traffic_filter_t*)filter;
 }
 
+/* for [-Wmissing-declarations] */
+void destroy_network_traffic_filter(network_traffic_filter_t *filter);
+
 void destroy_network_traffic_filter(network_traffic_filter_t *filter) {
-  network_traffic_filter_impl_t *_filter = filter;
+  network_traffic_filter_impl_t *_filter = (network_traffic_filter_impl_t*)filter;
```
lucaderi: Same here

src/network_traffic_filter.c
```diff
@@ -393,18 +437,24 @@ void destroy_network_traffic_filter(network_traffic_filter_t *filter) {
...
 
+/* for [-Wmissing-declarations] */
+void network_traffic_filter_add_rule(network_traffic_filter_t* filter, filter_rule_t* rules);
+
 void network_traffic_filter_add_rule(network_traffic_filter_t* filter, filter_rule_t* rules) {
   filter_rule_t *item=NULL, *tmp=NULL;
 
   HASH_ITER(hh, rules, item, tmp) {
-    network_traffic_filter_impl_t *_filter = filter;
+    network_traffic_filter_impl_t *_filter = (network_traffic_filter_impl_t *)filter;
```
lucaderi: And here

lucaderi (changes requested):
Please fix the issues without using casting

src/network_traffic_filter.c
joshuafc: I want to expose `network_traffic_filter_t` which contain interface  to outside, but hide implementation details inside `network_traffic_filter_impl_t`.  Codes outside `network_traffic_filter.c/h` no need to care about fields implementation details of `network_traffic_filter_impl_t`. for example fields  `filter_rule_t *rules`,`filter_rule_pair_cache_t *connections_rule_cache` inside `network_traffic_filter_impl_t` shouldn't accessible outside `network_traffic_filter.c/h`.

If this does not conform to the coding specification of the project, I will change it later.

joshuafc:
Now codes work without casting.

lucaderi:
Thank you

PR #509, opened 2020-11-17
Protection mechanism against duplicated MACs

### LINKED ISSUE #505: Prevention against duplicated MACs
A possible internal attack of n2n is MAC duplication. If two edges have the same (command-line provided) MAC, they won't be able to communicate. Moreover, looking at the output on the supernode's management port, they will overlap each other over time. A new UNREGISTER_SUPER message could solve this issue.

This new message contain an authentication token and the source MAC of an edge as fields.  It will be sent when an edge switches supernode or at the end of program. Supernode that receives this message will check if that MAC is present in the hash list and delete it.

To implement entirely this mechanism of protection and prevention against doublets of MACs and MAC spoofing, we need also to forward REGISTER_SUPER packets between supernodes in order to check if an edge can register or not.

fcarli3:
@Logan007  and I implemented a protection mechanism against duplicated MACs and MAC spoofing. To realize this feature, we used a new type of message (UNREGISTER_SUPER) and we used also the existing REGISTER_SUPER_NAK message and `update_edge` function.

When a new edge registers to a supernode, it sends a REGISTER_SUPER packet. Inside this packet, there is an `auth` field, so we added this field also in `n2n_sn_t` and `n2n_edge_t` structures in order to store that value. Auth fields on edge and supernode are randomly initialized inside `sn_init` and `edge_init` functions. With the auth field, we were able to implement an authentication mechanism.

When a supernode receives a REGISTER_SUPER packet, calls `update_edge` and in this function we added the comparison between auth tokens. If authentication fails, supernode will send a REGISTER_SUPER_NAK packet to edge. Otherwise, it will broadcast REGISTER_SUPER packet so other supernodes in the network will be able to authenticate that edge, too. 

In REGISTER_SUPER_NAK handling on supernode, we used HASH_FIND_PEER to find that spoofed edge and we removed it from the list. In REGISTER_SUPER_NAK handling on edge we compared the auth field received from the packet with the auth field in `n2n_edge_t` and if they are equal, edge is stopped with a log message. If auth tokens are different, we searched that edge inside `known_peers` and `pending_peers`.

On edge, we added a new `send_unregister_super` function that sends this new message. This function is called when an edge swtiches from a supernode to another one and when an edge ends.

This PR fixes #505 .

### DIFF
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -22,12 +22,13 @@
 #define MSG_TYPE_PACKET                 3
 #define MSG_TYPE_REGISTER_ACK           4
 #define MSG_TYPE_REGISTER_SUPER         5
-#define MSG_TYPE_REGISTER_SUPER_ACK     6
-#define MSG_TYPE_REGISTER_SUPER_NAK     7
-#define MSG_TYPE_FEDERATION             8
-#define MSG_TYPE_PEER_INFO              9
-#define MSG_TYPE_QUERY_PEER            10
-#define MSG_TYPE_MAX_TYPE	       10
+#define MSG_TYPE_UNREGISTER_SUPER       6
+#define MSG_TYPE_REGISTER_SUPER_ACK     7
+#define MSG_TYPE_REGISTER_SUPER_NAK     8
+#define MSG_TYPE_FEDERATION             9
+#define MSG_TYPE_PEER_INFO              10
+#define MSG_TYPE_QUERY_PEER             11
+#define MSG_TYPE_MAX_TYPE	        11
 
 /* Max available space to add supernodes' informations (sockets and MACs) in REGISTER_SUPER_ACK
  * Field sizes of REGISTER_SUPER_ACK as used in encode/decode fucntions in src/wire.c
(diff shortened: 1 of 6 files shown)


lucaderi:
@fcarli3 @Logan007 Not sure I understand the fix. You are relying on the auth field but in case you have an attacker from inside your network (so auth is correctly set) that creates troubles to the n2n, how can this mechanism help?

Logan007:
This basic authentication scheme relies on a secret shared between each edge and the supernode(s), something like a tax payer ID or social security number, created once at start-up, registered once with the authoritative supernodes at edge's birth, that does not change or does not get updated during lifetime (as usually also the MAC address would be expected) and needs to be presented to the supernode to be verified. The difference to the MAC address is that this ID only is presented to the supernodes and not to the other edges (as PACKETs or REGISTERs always contain the MAC address). Other edges would only be able to know about the auth string from capturing the traffic between supernodes and the other edge – some hurdling effort required compared to MAC addresses which come with free delivery.

So, this is to show how to generally counter the showcase of MAC address spoofing, namely by edge authentication. This specific implementation, due to current time constraints, is admittedly somewhat simple and just one step – a first – towards the implementation of maybe more sophisticated, crypto-based authentication mechanisms.

lucaderi:
Understood this is a basic mechanism. I would appreciate if you could write a file where you describe the mechanism as the code is poorly commented

PR #510, opened 2020-11-18
Update CMakeLists.txt

skyformat99:
1)  add openssl version check
2) add  build option for  pcap lib : N2N_OPTION_USE_PCAPLIB
3) add build option for  libzstd  : N2N_OPTION_USE_ZSTD

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,9 @@ OPTION(BUILD_SHARED_LIBS "BUILD Shared Library" OFF)
 
 # N2n specific params
 OPTION(N2N_OPTION_USE_OPENSSL "USE OPENSSL Library" OFF)
+OPTION(N2N_OPTION_USE_PCAPLIB "USE PCAP Library" OFF)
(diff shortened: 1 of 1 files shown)


fengdaolong:
I think we can just keep the CMakeLists and discard the Makefile (configure), keep only one set of build scripts to avoid maintaining multiple sets of build scripts. @Logan007 @lucaderi

Logan007:
With a view to the build system, I really have no clear-cut opinion as I do not know much about those. I guess a build system of choice should just be easy to use and maybe somewhat common.

I personally am extremely used to `make` and like the `--with-openssl` and `--with-zstd`-style parameters as well as the possibility to configure and pass the speed-wise essential `CFLAGS`. If you find a way to easily keep all that with CMake, make it work on all platforms and would document how to use it, it might be worth considering.

I have absolutely no objections neither way – as long as it does not get more complicated.

lucaderi:
I use makefile and would like to keep using it. So both make and cmake should work IMHO

fengdaolong:
This is just a proposal, I respect the majority vote.

lucaderi:
Thank you

PR #512, opened 2020-11-18
Protection mechanism against MAC spoofing

fcarli3:
This PR implements a protection mechanism for edges against MAC spoofing. In fact, if an edge suddenly goes crazy and start to send packets with random MAC (e.g. I tried to randomize REGISTER'S MAC), on the edge management port output we can see the list of peers exploding.

So I defined a new `find_peer_by_sock` function that search by socket inside `pending_peers` and `known_peers` before a peer is really added. That way, if a MAC change, we still can find that node and do not create a new one, thus, no exploding lists. I called this new function in parts of the code where it used HASH_FIND_PEER, right after it just like a second chance to find the peer.

Concerning supernodes, they are already able to cope with changing MACs because of searching by TAP IP implemented in `update_edge`. They are also supported by the auth mechanism implemented in #509.

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -1063,6 +1063,18 @@ void update_supernode_reg(n2n_edge_t * eee, time_t nowTime) {
 
 /* ************************************** */
 
+/** NOT IMPLEMENTED
+ *
+ *  This would send a DEREGISTER packet to a peer edge or supernode to indicate
+ *  the edge is going away.
+ */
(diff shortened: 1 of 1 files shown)


lucaderi:
This build is broken. Please fix the PR

fcarli3:
It should be fixed now.

lucaderi:
@fcarli3 Can you please update this PR as the are some merge errors?

Issue #344, opened 2020-08-01
How to know whether p2p punching success? Is it possible to disable supernode forward?


joshuafc:
Greate Project!

How to know whether p2p punching success? Is it possible to disable supernode forward? 

If p2p connection failed,  then keep node unreachable, to save my aws cloud server broadband bill.

Logan007:
> How to know whether p2p punching success?

Please see #208 on how to get further stats including the number of packets sent via p2p vs. through the supernode.

> Is it possible to disable supernode forward?

This is not implemented.

> If p2p connection failed, then keep node unreachable, to save my aws cloud server broadband bill.

I am not completely sure if this would be compatible with the current communication scheme because as far as I can see, first  PACKTETs to so far unknown nodes via a supernode additionally serve to introducing the sending nodes and their p2p sockets. Also, only the supernode will know where to forward the fist packet to (by the help of having the nodes REGISTERed to the supernode).

joshuafc:
thanks! I will close it.

Logan007:
Concerning you second question...

_Is it possible to disable supernode forward?_

... we should continue discussion at #784 to avoid duplicates. Closing it here.

PR #515, opened 2020-11-19
Add -P for supernode to disable forward peer to peer packet, make it possible to only use p2p mode.

joshuafc:
#344 How to know whether p2p punching success? Is it possible to disable supernode forward?

===================

Add -P for supernode. If configured, supernode will drop packets unless it is ARP packet, or it is broadcast packet.

As documented here : https://github.com/ntop/n2n/blob/dev/doc/Hacking.md#packet-message-format ,flags N2N_FLAGS_ARP_PACKET = 0x100 is add to mark ARP packet.

It is edge's responsibility to mark packet's `N2N_FLAGS_ARP_PACKET` flag.

For supernode: broadcast packet and ARP packet will always forward to help NAT punching. Other packet will be droped if `-P` is set.

A simple  interactive operation is add to supernode like edge. Forward switch can dynamic change by command `forward` as below:
```
root@ubuntu:~# netcat -u 127.0.0.1 5645

    id    tun_tap             MAC                edge                   hint             last_seen
-------------------------------------------------------------------------------------------------
community: *Federation
community: *****
*****
----------------------------------------------------------------------------------------------------
uptime 1549 | edges 6 | reg_sup 381 | reg_nak 0 | errors 0 
fwd 139 | broadcast 1351 | cur_cmnts 2
last_fwd  1 sec ago | last reg  9 sec ago

forward: disabled


Type "help" to see more commands.

forward
    id    tun_tap             MAC                edge                   hint             last_seen
-------------------------------------------------------------------------------------------------
community: *Federation
community: *****
*****
----------------------------------------------------------------------------------------------------
uptime 1552 | edges 6 | reg_sup 381 | reg_nak 0 | errors 0 
fwd 139 | broadcast 1351 | cur_cmnts 2
last_fwd  4 sec ago | last reg  12 sec ago

forward: enabled


Type "help" to see more commands.
```


include/n2n_typedefs.h
```diff
@@ -253,7 +253,15 @@ typedef enum n2n_pc
...
-#define N2N_FLAGS_OPTIONS               0x0080
+
+/* This value is used in sn_utils.c process_udp, flags large than this value will make supernode treat packet's header as encrypted.
+ * If add new flags, must change N2N_FLAGS_MAX_VALUE at the same time.*/
+#define N2N_FLAGS_MAX_VALUE            0x0100
+
+/* Set at edge if the packet paylod is ARP. Supernode always forward ARP packet.*/
+#define N2N_FLAGS_ARP_PACKET            0x0100
+
+#define N2N_FLAGS_OPTIONS               0x0080 // TODO: this flags not need any more, can be delete.
```
Logan007: As far as I see it, `N2N_FLAGS_OPTIONS` is meant to indicate the first available value for (new) user options, so `0x0080` could be used as `ARP` while `N2N_FLAG_OPTIONS` could be shifted to `0x0100` (and optionally renamed to `N2N_FLAG_MAX_VALUE` if more convenient)

src/edge_utils.c
```diff
@@ -1945,7 +1959,7 @@ void readFromIPSocket(n2n_edge_t * eee, int in_sock) {
 	}
 
 	/* Update the sender in peer table entry */
-	check_peer_registration_needed(eee, from_supernode, pkt.srcMac, NULL, NULL, orig_sender);
+        check_peer_registration_needed(eee, from_supernode, pkt.srcMac, NULL, NULL, orig_sender, 1);
```
Logan007: Is this `1` value the same as `is_packet_msg`?

src/edge_utils.c
```diff
@@ -2344,7 +2358,7 @@ int run_edge_loop(n2n_edge_t * eee, int *keep_running) {
     update_supernode_reg(eee, nowTime);
 
     numPurged =  purge_expired_registrations(&eee->known_peers, &last_purge_known, PURGE_REGISTRATION_FREQUENCY);
-    numPurged += purge_expired_registrations(&eee->pending_peers, &last_purge_pending, PURGE_REGISTRATION_FREQUENCY);
+    numPurged += purge_expired_registrations(&eee->pending_peers, &last_purge_pending, eee->conf.register_interval);
```
Logan007: I do not fully understand the reasoning behind this change. Could you please provide a short explanation?

src/sn_utils.c
```diff
@@ -748,6 +748,25 @@ static int process_mgmt(n2n_sn_t *sss,
...
+                            "\t<enter> | Display statistics\n\n");
+
+    sendto_mgmt(sss, sender_sock, (const uint8_t *) resbuf, ressize);
+
+    return 0;
+  }
+  if ((0 == memcmp(mgmt_buf, "forward", 7))) {
+    ressize = 0;
+
+    sss->enable_forward = sss->enable_forward == 0 ? 1 : 0;
```
Logan007, line 766: :+1:

I really like the idea of adding new commands to the management port!

Logan007 (commented):
Any PACKET type message (also the non-ARP) forwarded via the supernode are able to initiate a p2p-communication through the socket added to the PACKET at the supernode. I am curious to see if their omission has any impact on p2p-communication – it probably depends on purge timing vs. ARP table refresh timing.

src/edge_utils.c
joshuafc: I ran into a situation:  After run some time, `edge A` and `edge B` always communicate by supernode forward. But if I kill and restart any edge, `edge A` and `edge B` can punching success.

I find that `MSG_TYPE_PACKET` and `MSG_TYPE_REGISTER` will call `check_peer_registration_needed`, if punching not success, peer not exists in `known_peers`, `register_with_new_peer` will be called. Inside `register_with_new_peer`,  if peer already in `pending_peers`, nothing will do except update `last_seen`.
 
What if B is in A's `pending_peers`, same time A is in B's  `pending_peers`, but punching temporary failed? 

In summary, current code, punching will execute only once when add peer to `pending_peers`. If some reason this time punching failed, `last_seen` will be update every `MSG_TYPE_PACKET` and `MSG_TYPE_REGISTER` packet. So purge_expired_registrations cannot clear that peer from `pending_peers` because it use `last_seen`. So no punching retry will happen.


So I made a temporary imperfections fix: only update `last_seen` at `MSG_TYPE_PACKET`, at my code, this type packet is dropped by `-P` if not punching success.  So if punching not success for some time, peer will be delete from `pending_peers`. Then next time when received a packet from the peer forward by supernode, `check_peer_registration_needed` will tigger a new punching process while add peer to `pending_peers`.

`is_packet_msg` is used to pass info (packet type is `MSG_TYPE_PACKET`) to functions which be called.

This is not a perfect fix for that issue, maybe we need find an other way.

joshuafc: This situation is found before I introduce the `-P` function.  And it can be confirm at dev@ntop/n2n at this time.

joshuafc: I have made a new way to let punching retry:

When begin punching (add peer to pending_peers), record time into `last_try_punching_time`.

At `run_edge_loop`, use `purge_expired_pending_peers` instead of `purge_expired_registrations` to purge pending peers by last_try_punching_time.

All changes about `is_packet_msg` and `last_seen` is reverted.

==== working flow description ====

When punching started, record the time. There has 10s to wait punching success.
If success, peer will be delete from `pending_peers`, insert into `known_peers`. According to current upstream codes.
If not success, peer will be delete from `pending_peers` by `purge_expired_pending_peers` because of timeout(10s later).
Once peer be deleted from `pending_peers`, it will be treat as new peer, and a new punching process will be trigger when MSG_TYPE_PACKET or MSG_TYPE_REGISTER is received from that peer.

lucaderi: @Logan007 What is your opinion? Are you happy with this?

Logan007: Basically, I think you have provided n2n with an excellent idea @joshuafc and I absolutely support its implementation.

I still am struggling with some details which might be caused by a lack of knowledge about networks on my side, however, the following questions arise – maybe, answers are easy:

- Is it really likely that punching fails, i.e. that the fields `last_seen` and `last_try_punching_time` fall apart? Do we really need that other field?

- **If** this _only-ARP-through-the-supernode_ approach keeps the network working, I see no issue in not implementing it. Does the `known_peers` entry (at the edge) or the `peer_info` entry (at the supernode, eventually asked for socket by MAC through QUERY_PEER_INFO mechanism) time-wise survive the ARP entry? What happens if connection shall be re-built while those n2n-entries have gotten purged and ARP entry is still in table?

- What happens if  `-S` and `-P` are used parallel? Both options control the behavior of PACKET (and maybe REGISTER and REGISTER_ACK) sending, acceptance and forwarding. Shouldn't we help the user to avoid this situation (maybe check if both options are present)? A future extension (in a separate pull request, not now) could merge  these two options to one `-S_` with, for example, `-S0` indicating the (default) regular behavior, `-S1` the current `-S`, `-S2` the current `-P`.

- Does the supernode really requires the `-P` option? It affects all communities then. Maybe you had different use-cases in mind, but wouldn't it be sufficient to leave it to the edges only to use `-P` or not `-P`? If they do not send out anything, the supernode does not need to reject anything.

joshuafc: - `last_seen` will be update if received a supernode forward `PACKET` or `REGISTER`.

This implementation currently instability. As I Mentioned, codes at `ntop:dev` branch:

> After run some time, edge A and edge B always communicate by supernode forward(It cost me a lot of traffic bills). But if I kill and restart both edge, 
> edge A and edge B can punching success. 

Although I try add a re-punching by `last_try_punching_time`, this will still happen. With `-P`, edge A and edge B link broken forever.

Maybe we need to solve the re-punching problem first, It's a little hard for me to solve the punching issue. After that, I will make a new edge-only implementation.

PR #522, opened 2020-11-24
changed timer source

### LINKED ISSUE #463: Slow rate and packet loss in old version of openwrt

Logan007:
This pull request chooses another source for high precision timer values being used with timestamps (header encryption) and within the benchmark tool. It solves an issue that has been seen on lightweight computers using _uclibc_ instead of _glibc_, e.g. openWRT boxes, arosen from _uclibc_ not returning µsec values from calling `gettimeofday()`. Hence, it is changed to `clock_gettime()` which shall have an accuracy of at least (!) 20 msec on every computer.

**Is there anybody who could confirm that it also works on Windows, please?** And especially, if calling `GetSystemTimeAsFileTime()` at `n2n.c:530` requires inclusion of additional headers (`sysinfoapi.h` or `Windows.h`) for compilation? Thank you!

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -48,8 +48,8 @@ N2N_LIB=libn2n.a
 N2N_OBJS=$(patsubst src/%.c, src/%.o, $(wildcard src/*.c))
 N2N_DEPS=$(wildcard include/*.h) $(wildcard src/*.c) Makefile $(N2N_LIB)
 
-LIBS_EDGE+=$(LIBS_EDGE_OPT)
-LIBS_SN=
+LIBS_EDGE+=-lrt $(LIBS_EDGE_OPT)
+LIBS_SN=-lrt
 
 #For OpenSolaris (Solaris too?)
 ifeq ($(shell uname), SunOS)
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -54,7 +54,6 @@
 
 #define PACKAGE_BUILDDATE (__DATE__ " " __TIME__)
 
-#include <time.h>
 #include <ctype.h>
 #include <stdlib.h>
 
@@ -121,6 +120,7 @@
 #define closesocket(a) close(a)
 #endif /* #ifndef WIN32 */
 
+#include <time.h> /* should be included after eventually including unistd.h for _POSIX_TIMERS macro gets correctly defined */
 #include "minilzo.h"
 #include <signal.h>
 #include <string.h>
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -728,7 +728,14 @@ typedef struct n2n_sn
   n2n_auth_t auth;
 } n2n_sn_t;
 
-
+#if defined(WIN32)
+// used for n2n.c's clock_gettime() replacement and calls thereof
+// from time_stamp() and several functions of tools/benchmark.c
+struct timespec {
+  long tv_sec;
+  long tv_nsec;
+};
+#endif
 
 
 #endif /* _N2N_TYPEDEFS_H_ */
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -728,14 +728,7 @@ typedef struct n2n_sn
   n2n_auth_t auth;
 } n2n_sn_t;
 
-#if defined(WIN32)
-// used for n2n.c's clock_gettime() replacement and calls thereof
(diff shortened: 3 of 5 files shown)


skyformat99:
My personal advice is not particularly necessary and my advice may not be correct.

Using this function clock_gettime()  can cause some problems, In particular, when the Linux kernel is less than 2.6,
https://stackoverflow.com/questions/13069758/got-compile-error-when-use-clock-gettime-in-c99
http://www.tin.org/bin/man.cgi?section=3&topic=clock_gettime

In addition, the benchmark test does not need to be particularly precise,  just a reference point.

Logan007:
> Using this function clock_gettime() can cause some problems, In particular, when the Linux kernel is less than 2.6

The stackoverflow discussion proposes to use `-std=gnu99` in that case (instead of the `-std=c99` option which seems to cause the problem). Do you think it will solve it?

So far, I have not realized that `clock_gettime()` isn't supported by kernel before 2.6. Good catch!

Do you think that there are still a lot of pre-2.6 kernel? I have no idea.

> In addition, the benchmark test does not need to be particularly precise, just a reference point.

Switching to `clock_gettime()` is mainly because of some systems not offering any sub-second accuracy with `gettimeofday()` at all. This does not aim at the `n2n-benchmark` tool, it just uses the function that is available. I did not want to maintain two timing functions so I changed it for both, timestamp generation as well as `n2n-benchmark`.

But, perhaps, it will be required to maintain two code versions anyway: If on Linux and below 2.6, use `gettimeofday()`, `clock_gettime()` otherwise. Can we even find out? **What do you think?** Could they be distinguished by `#if defined(CLOCK_REALTIME)`?

I put this pull request "on hold" marking it a draft until a solution is found.

> My personal advice is not particularly necessary and my advice may not be correct.

You should know by now that every feedback is extremely welcome!

skyformat99:
thanks,  Here are the details:
https://github.com/exdev/exsdk/wiki/gettimeofday-vs-clock_gettime

It says something like this:
clock_gettime is present only on newest Linux. other system have only gettimeofday() – vitaly.v.ch Dec 18 '09 at 16:23

For some embedded systems, many still use older linux.
In order to make n2N more widely applicable, it is recommended not to modify for the time being.

Logan007:
Thank you!

I get from the discussion at that other page that older Linux does not have `clock_gettime()` (2009, below 2.6?). So, I think I will leave the current version (`gettimeofday()`) as **default** for time stamp generation and use the `clock_gettime()` as **additional option**, and only if it is available (`#ifdef _POSIX_TIMERS`).

Why do I still want to add that other option? Because otherwise, some edges or supernodes on openWrt using uclibc would not be able to use that feature at all.

Do you expect difficulties if we add this as additional optional code?

Logan007:
I have implemented the changes laid out in the post above and also reverted the time measurement in benchmark (which might be subject to some later pull request then). What do you think @skyformat99?

skyformat99:
> I have implemented the changes laid out in the post above and also reverted the time measurement in benchmark (which might be subject to some later pull request then). What do you think @skyformat99?


thanks.
if use clock_gettime(),  When compiling, you should add the -lrt  option

Logan007:
> you should add the -lrt option

I do not see any warning here on Linux using `gcc` without `-lrt`. Is that a Windows specific requirement?

**EDIT**:   Oh, now I see the man page stating _Most systems require the program be linked with the librt library to use these functions._ I will add it to the Makefile then.

skyformat99:
> > you should add the -lrt option
> 
> I do not see any warning here on Linux using `gcc` without `-lrt`. Is that a Windows specific requirement?
> 
> **EDIT**: Oh, now I see the man page stating _Most systems require the program be linked with the librt library to use these functions._ I will add it to the Makefile then.

thanks.

Issue #563, opened 2020-12-25
ARM PERFORMANCE:  Does NEON hardware support really help?


Logan007:
I found that SPECK cipher is slower with NEON hardware support than without on Raspberry Pi 3B+ (64-bit Arch Linux on ARM):

- Compiled with NEON support `./configure CFLAGS="-O3 -march=native"`, `tools/n2n-benchmark` shows around 81 MB/s for SPECK.

- Compiled **without** NEON support  `./configure CFLAGS="-O3 -march=armv8-a+nosimd"`, the benchmark goes up to **108** MB/s.

This is a plus of 33 percent and very surprising as I expected the NEON SIMD hardware support to accelerate this cipher's performance.

I would be very interested if your ARM platform shows similar tendencies. Perhaps, the issue is specific to Raspberry CPU because it may have a very optimized scalar core (don't know much about). But if this shows up on all platforms, I will remove the NEON code.

So, if you have an ARM platform, especially an unusual one, **please test and let me know your results** here. Thank you!

Issue #466, opened 2020-10-15
review header encryption


Logan007:
Discussing a larger time frame for allowed time stamp jitter (#463), I found that replay protection needs adjustments as well. That includes but is not limited to the length of the checksum as well extending header encryption to the full packet.

A larger frame for time jitter offers more opportunities for quick replay attacks. Especially, attaching old payloads to freshly catched headers could be a possibility especially while using the faster stream ciphers (ChaCha20, SPECK) requiring a number of replays (depending on checksum-length) to be performed during the jitter time frame. No need to panic, it is not as easy as it reads here because checksum inside the ethernet frame¹ and tcp/udp need to be tricked, too, but better be safe than sorry. 

So, here is my work plan:

- I want to extend header encryption to payload as well to close the crypto-gap between header and payload which would effectively prevent any _replay-with-exchange-of-payload_. That might go along with SSE and AVX2 hardware acceleration support for header encryption (so far not implemented because headers alone are relatively short). A possibly positive side-effect is that payload encryption can be omitted **if** you trust your supernode, e.g. if is under your full control (DMZ, home, …)

- To counter _trial-and-error-replays_ using bit-manipulations (especially with stream ciphers), I want to extend the checksum length.

- Finally, to prevent _same-packet-replays_ during that extended time-frame, I plan to test how a counter could come in handy instead of random bits used in the pre-IV of header encryption and how it affects packets loss with a view to overtaking packets. This might be the toughest part but maybe the least critical one as the probably most widely used encapsulated TCP ensures sequence.

This all will take a while to implement, discussion and suggestions are very welcome.

________
¹ CRC? linear, easy to modifiy if stream-ciphered

Logan007:
> I want to extend the checksum length

> I plan to test how a counter could come in handy

This affects header encryption IV format. So, here is what I have come up with for the plain text of the encrypted IV (96 bits):

- Extend time stamp to full 64 bit by using 32 bit second value and 32 bit sub-second value (`clock_gettime()`) – actually, there will be two bits remaining because nanosecond accuracy only eats up 30 bits.

- To counter varying accuracy (#463), increment the sub-second counter by one if equal to the one used before. This should even work with 20 millisecond accuracy.

- XOR a 64-bit full packet checksum into the timestamp. This serves two purposes: 
[...]

PR #565, opened 2020-12-28
hardware-accelerated header encryption

Logan007:
This PR consolidates the code of the two SPECK-128-CTR versions with 128 bit and 256 bit key-size as they share a lot of commonalities. As a positive side effect, also header encryption now benefits from supported hardware-acceleration (mainly SSE2, SSSE3, AVX2 – maybe NEON, see #563?) which then will especially help the upcoming changes to header encryption (#466).

This is still fully compatible to current _dev_.

### DIFF
--- a/include/speck.h
+++ b/include/speck.h
@@ -17,7 +17,7 @@
  */
 
 
-// cipher SPECK -- 128 bit block size -- 256 bit key size -- CTR mode
+// cipher SPECK -- 128 bit block size -- 128 and 256 bit key size -- CTR mode
 // taken from (and modified: removed pure crypto-stream generation and seperated key expansion)
 // https://github.com/nsacyber/simon-speck-supercop/blob/master/crypto_stream/speck128256ctr/
 
@@ -51,6 +51,7 @@
 typedef struct {
     u256 rk[34];
     u64 key[34];
+    u32 keysize;
 } speck_context_t;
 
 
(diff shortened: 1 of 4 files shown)


skyformat99:
For compatibility, or for embedded systems, it is recommended not to enable hardware encryption by default

Logan007:
The support for hardware acceleration is a compile-time decision based on the availability of gcc's platform specific macros.

If you cross-compile for a specific platform, e.g. some embedded system, the code will be compiled to run on exactly the platform specified. As you always need to specify the target the platform when cross-compiling, there is no danger of inadvertently enabling it. 

On the other hand, if you do not specify a target platform, it will be compiled to run on the local platform, as usual.

Therefore, I see no danger of wrongly compiled code. Did you experience differently?

skyformat99:
> The support for hardware acceleration is a compile-time decision based on the availability of gcc's platform specific macros.
> 
> If you cross-compile for a specific platform, e.g. some embedded system, the code will be compiled to run on exactly the platform specified. As you always need to specify the target the platform when cross-compiling, there is no danger of inadvertently enabling it.
> 
> On the other hand, if you do not specify a target platform, it will be compiled to run on the local platform, as usual.
> 
> Therefore, I see no danger of wrongly compiled code. Did you experience differently?

thnaks, me too.
If you are not using assembly code, it no effect is expected.

PR #570, opened 2021-01-04
switched to Pearson B. Hashing

### LINKED ISSUE #537: hash speed
I am not too happy with the speed of the hashing function on non-accelerated Intel paltforms (AES-NI) which mainly affects header encryption, all other calls are more or less singular.

I think that byte-wise digestion is slowing it down a bit. So, I might **either** change Pearson hash's word size to 32 bit including a 32-but permutation **or** come up with something completely different.

Logan007:
This pull request changes the current hashing function from pure Pearson Hashing to  [Pearson Block Hashing](https://github.com/Logan007/pearsonB). Speed-ups of half a magnitude or so are observed, you can use the `tools/benchmark` utility to determine your own mileage. On i7 5775C CPU, I see **1,953.6 MB/s**.

Especially header encryption should benefit.

This is compatibility-breaking as hash values are formed differently now, so all edges and supernodes require re-compilation.

Fixes #537.

### DIFF
--- a/include/pearson.h
+++ b/include/pearson.h
@@ -1,5 +1,5 @@
 /**
- * (C) 2007-20 - ntop.org and contributors
+ * (C) 2007-21 - ntop.org and contributors
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


lucaderi:
Great work!

Issue #567, opened 2020-12-30
Can one of supernode in federation stays behind NAT ?


timfaner:
My network topology is like this :

<img width="601" alt="image" src="https://user-images.githubusercontent.com/15355789/103334676-3e4d8f00-4aad-11eb-8573-ebbd76e39fdd.png">

Some of edge nodes are behinds a __symmetrical__ NAT , and some are out of NAT scoop. Can I put a supernode behind NAT, to make edge nodes in NAT communicate without relay ?

Logan007:
Well, it depends…

If the internal edges are on the same LAN, they might find each other via multicast without the help of a supernode. Note that this feature currently is not available in Windows version (see #65) of edges.

An internal supernode which can provide the internal IP addresses/sockets would help to connect the internal edges to each other.

In that given scenario, I think it could work only if the internal edges are not able to connect to the external supernode directly (because no ping, no registration then, only internal sockets distributed for p2p). Unfortunately we have no possibility for edges to prefer certain supernodes.

If the internal edges are non-Windows computers, or you  experimentally re-enable multicast (reportedly working for some on Windows), or following a solution @fengdaolong proposed in #65 (no idea yet if that could work), you do not need the internal supernode.

Please be aware that to work properly, the federation feature requires a fix which is not included in the _dev_ branch yet (see #568).

fengdaolong:
> or following a solution @fengdaolong proposed in #65 (no idea yet if that could work),

This solution is feasible, and I have implemented the code. Although the multicast function can be used under windows, other error codes are occasionally prompted. I have not studied the reasons for these error codes. I can submit a PR in time for everyone to optimize it.

timfaner:
Thanks for reply. Now I think maybe the best way to solve this problem is run Linux on every machine 😂

Logan007:
This would be a general recommendation anyway… :wink:

Logan007:
@timfaner, does #576 work for you on Windows?

Logan007:
I assume #576 to be working on Windows. Please open a new issue if multicast peer detection causes problems.

Issue #452, opened 2020-10-04
LAN discovery


aman448:
It's seem that I/we can't get working services LAN discovery,
There may be a configuration problem? or is it currently not possible?

Logan007:
LAN discovery via multicast is disabled on Windows because there were too many issues, see #65.

Concerning other operating systems, we have not heard of any problems recently.

Please note that in case you use `-S` on edge's command line, all connections will be made through the supernode – also those to locally neighboring peers. Other than that, the feature is not configurable.

aman448:
Thanks for the quick reply, yes it happened to my friend which is on Windows 10, while I myself on Linux.

Logan007:
If you are able to compile yourself for Windows, you might want to re-enable it by commenting line 31 of the `include/edge_utils_win32.h` file and experiment with the findings of #65: some blame a driver version.

aman448:
I can't figure out it, there may problem on **my end**,
I tried torrenting and got nothing via "Local Peer discovery" with a friend who is using Windows 10.
later I tested some video games that were hosted on VPS (Quake 3, Jedi Knight), searching LAN network for games
and got nothing again.
Here the configuration:
Me (Linux):
``./edge -S -A3 -r -E -D -z2 -c *** -k *** -a 192.168.254.1 -p 50001 -l ***:***``
Edge runs on the same machine that runs supernode (VPS) (Linux):
``./edge -S -A3 -r -E -D -z2 -c *** -k *** -a 192.168.254.2 -p 50001 -l ***:***``

Logan007:
> Please note that in case you use `-S` on edge's command line, all connections will be made through the supernode – **also those to locally neighboring peers.** 

That is, there will be no multicast peer registration (because everything runs through the supernode).

Please try to **omit** `-S`.

Logan007:
_P. S._ Are we talking about the same thing? Is your friend physically in the same LAN than you? Or are you talking about trasmitting general broad/multocasts through n2n?

aman448:
> _P. S._ Are we talking about the same thing? Is your friend physically in the same LAN as you are? Or are you talking about transmitting general broad/multicasts through n2n?

Not physically, and about generally transmitting multicasts. In text above I mention that I tried running some video games servres on VPS machine for testing purpose.

I removed all options beside `-c`, `-k`, `-a`, `-p`, `-l` i.e
`sudo ./edge -c *** -k *** -a 192.168.254.* -p 50001 -l ***` for both machines
[...]

Issue #133, opened 2019-06-11
P2P can not be formed in the LAN


lucktu:
Some people say there is no need to waste too much time building P2P in the LAN (It's actually very complicated), I agree.

If it is convenient, you can optimize it (It is best not to use parameters to complete, for example -L, it better be automatic).

Otherwise, ignore the proposal please.

emanuele-f:
Lan p2p already occurs by using multicast

lucktu:
Today, other friends also said, no problem.
That's great!  
Thank you!

lucktu:
It's not really working

Logan007:
Note that multicast has been disabled in Windows version (#65). A fix is under [discussion](https://github.com/ntop/n2n/issues/567#issuecomment-753758330).

Logan007:
#576 seems to solve it. If #576 does not work for you, please open a new issue.

Issue #65, opened 2019-02-09
ERROR: Failed to bind to local multicast group 224.0.0.68:1968


Menci:
On Windows 10, with patch below so I compiled it successfully.

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 3e351df..cacbb77 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,6 @@ endif(NOT DEFINED CMAKE_BUILD_TYPE)
 #set(CMAKE_BUILD_TYPE Debug)
 #set(CMAKE_BUILD_TYPE Release)

-if (DEFINED UNIX)
 # None
 set(CMAKE_C_FLAGS "-Wall -Wshadow -Wpointer-arith -Wmissing-declarations -Wnested-externs")
 set(CMAKE_CXX_FLAGS "-Wall -Wshadow -Wpointer-arith -Wmissing-declarations -Wnested-externs")
@@ -39,7 +38,8 @@ set(CMAKE_CXX_FLAGS_DEBUG "-g")
 # Release
 set(CMAKE_C_FLAGS_RELEASE "-O2 -DNDEBUG")
 set(CMAKE_CXX_FLAGS_RELEASE "-O2 -DNDEBUG")
-endif()
+
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -I/usr/include -Wl,-LC:\\\\MinGW\\\\lib -Wl,-LC:\\\\MinGW\\\\msys\\\\1.0\\\\lib")

 ## DEBUG FOR CMAKE
 #message(${N2N_VERSION})
diff --git a/edge.c b/edge.c
index d92472c..3028722 100644
--- a/edge.c
+++ b/edge.c
@@ -730,7 +730,15 @@ int main(int argc, char* argv[]) {

     /* allow multiple sockets to use the same PORT number */
     setsockopt(eee.udp_multicast_sock, SOL_SOCKET, SO_REUSEADDR, &enable_reuse, sizeof(enable_reuse));
+#ifndef WIN32
+    /*
+     * From https://stackoverflow.com/questions/17212789/multiple-processes-listening-on-the-same-port
+     * Windows only knows the SO_REUSEADDR option, there is no SO_REUSEPORT.
+     * Setting SO_REUSEADDR on a socket in Windows behaves like setting
+     * SO_REUSEPORT and SO_REUSEADDR on a socket in BSD...
+     */
[...]

PR #576, opened 2021-01-08
Solve the problem that Windows cannot multicast.

fengdaolong:
#65
#452
#567 
#133


Logan007:
Thank you! That will help a lot of issues. Let's merge it for more people will be able to test it.

What is the exact idea of the adapter selection strategy inside the loop of `get_best_interface_ip()`?

fengdaolong:
The purpose of `get_best_interface_ip()` is to obtain the IP address of the physical network adapter.

PR #606, opened 2021-01-27
Added network interface metric option for windows

anzz1:
Idk if did the config part right, PR reviewer please refactor if needed.

### DIFF
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -599,6 +599,7 @@ typedef struct n2n_edge_conf {
     int                mgmt_port;
     n2n_auth_t         auth;
     filter_rule_t      *network_traffic_filter_rules;
+    int                metric;                /**< Network interface metric (Windows only). */
 } n2n_edge_conf_t;
 
 
--- a/src/edge.c
+++ b/src/edge.c
@@ -280,7 +280,7 @@ static void help (int level) {
                "                   |    | line  |  ...TCP+/-,UDP+/-,ICMP+/-'\n");
 #ifdef WIN32
         printf(" -x <metric>       | set TAP interface metric, defaults to 0 (auto),\n"
-               "                   | disables multicast local peer-discovery if set\n");
+               "                   | e.g. set to 1 for better multiplayer game detection\n");
 #endif
         printf ("\n");
         printf (" LOCAL OPTIONS\n");
--- a/src/edge.c
+++ b/src/edge.c
@@ -627,6 +627,7 @@ static int setOption (int optkey, char *optargument, n2n_tuntap_priv_config_t *e
 #ifdef WIN32
         case 'x': {
             conf->metric = atoi(optargument);
+            ec->metric = atoi(optargument);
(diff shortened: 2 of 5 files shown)


Logan007:
Thank you so much!

Unfortunately, it seems that we both have been working on the same code at the same time. I will resolve the resulting conflicts and make required changes.

Also, given the fact that `-x` could conflict with the multicast local edges-peer detection, shouldn't we disable the local peer detection if `-x` is given? What do you think?

anzz1:
I'm not really very well versed with multicast but shouldn't it be a different concept than broadcast, so the interface metric would not matter in this case? Also in PR #576 where fengdaolong solved the windows multicast issue by choosing the right (WAN) interface to multicast to? @fengdaolong  should weigh in on this as he probably knows better. 

In any case we would want both broadcast and local peer detection working.

Logan007:
> I'm not really very well versed with multicast...

Nor am I. I was just assuming from [earlier discussions](https://github.com/ntop/n2n/issues/452#issuecomment-749367396) that multicast on Windows will happen on the interface with the lowest metrics, too. But, maybe, that has been a broadcast and not a multicast behind the scenes...

Also, I am not sure if `get_best_interface_ip()` (called for multicast interface detection on Windows) will be distracted towards the tap device then resulting in trying to multicast-detect peers through the tap. Indeed, @fengdaolong, your counsel would be highly appreciated.

anzz1:
The way I understood it, the `eee->curr_sn->sock.addr.v4` contains the public internet ipv4 address of the supernode, and not the private VPN one, am I right? So the function `get_best_interface_ip` which uses winapi GetBestInterface to find the best (and in most cases, only) route to that address, and the interface for it, would return the WAN (not TAP) interface and add it to the multicast group so regardless of the metric it would use that WAN (home router) interface for local peer discovery, as it should. So changing the TAP interface metric shouldn't affect the multicast in this case. But we'll need to wait for confirmation from fengdaolong

e: to clarify, the line `mreq.imr_interface.s_addr = inet_addr(ip_addr);` instead of `mreq.imr_interface.s_addr = htonl(INADDR_ANY);` in edge_utils.c is where the magic happens would force to use the route of that certain interface, and not "any" interface which would make broadcast/multicast use the interface with the lowest metric in Windows, and all interfaces in other OSes. Because of that, the metric should not matter at all

Logan007:
I think you are right!

I have not realized that `get_best_interface_ip()` looks for the best interface with a route *to the supernode* which can't be the TAP then – at least not this edge's one; I had edges connect to a supernode via another edge running... crazy stuff :wink:

So, I assume we can safely remove the multicast-guarding `if`. I will do that right away to trigger travis CI build, I hope you don't mind, and merge at some later point in time if we do not hear back from @fengdaolong with any objection in due time.

Unfortunately, I am not able to run tests on Windows. Can you confirm that the device metric actually changes?

Logan007:
Merged to have this feature easily accessible for user testing.

Issue #603, opened 2021-01-27
questions when I read the source code..


randomize00:
Hello, I've been reading the source code for a while.
I have two little questions.
First, in src/sn_util.c:1488 :

`if((from_supernode == 0) != (comm->is_federation == IS_NO_FEDERATION)) {`

this checking seems odd, Can REGISTER_SUPER_ACK be sent by an edge ?

Second.
I find a macro, N2N_FLAGS_SOCKET, since there is no comment, I don't know what does it mean. it looks like it always appears with the macro N2N_FLAGS_FROM_SUPERNODE.

any  helpful reply will be appreciated, thank you.

Logan007:
> Can REGISTER_SUPER_ACK be sent by an edge ?

No, it can't. 

`if((from_supernode == 0) != (comm->is_federation == IS_NO_FEDERATION))` makes sure that packets that claim to be from a supernode (`from_supernode` flag in the header) really are from (super)nodes of the corresponding supernode community. In other words, flag and community must correspond. If not, there is an error or some manipulation we want to detect and avoid. Thus, we perform the check with the following error message.

> I find a macro, N2N_FLAGS_SOCKET ... it always appears with the macro N2N_FLAGS_FROM_SUPERNODE.

Those are flag masks defined in `include/n2n_typedefs.h:259`. They define the (currently two) flags used in the common section of each packet. FROM_SUPERNODE indicates that this is a packet sent from or through a supernode which is important to know in case we need to decide if to forward again or not, etc. The SOCKET flag indicates that some previous (super-)node added the socket of the original sender to the packet, so the the final recipient can use that one to directly send data back. The flags are not necessarily equal.

randomize00:
Thank you very much.
About the checking,  I have a new question - Do you mean, if an edge (not in federation commutity), or a supernode ( in federation community ) manipulate a REGISTER_SUPER_ACK and send to us, then it can pass this checking ? How to guarantee that those nodes won't try to manipulate a packet to attack us ?

Logan007:
> Do you mean, if an edge (not in federation commutity), or a supernode ( in federation community ) manipulate a REGISTER_SUPER_ACK and send to us, then it can pass this checking ?

No. What I mean is that those manipulations are detected by performing this check.

> How to guarantee that those nodes won't try to manipulate a packet to attack us ?

[...]

PR #611, opened 2021-01-29
Safer checking for REGISTER_SUPER_ACK and UNREGISTER_SUPER

randomize00:
safer checking for REGISTER_SUPER_ACK and REGISTER_SUPER_NAK ( only those packet from supernode in federation are legal )

For more details please see #603

### DIFF
--- a/src/sn_utils.c
+++ b/src/sn_utils.c
@@ -1429,7 +1429,7 @@ static int process_udp (n2n_sn_t * sss,
                 return -1;
             }
 
-            if((from_supernode == 0) != (comm->is_federation == IS_NO_FEDERATION)) {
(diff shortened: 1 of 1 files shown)


Logan007:
While the REGISTER_SUPER_ACK is not sent out by edges, the UNREGISTER_SUPER, however, is supposed to be sent only by edges. So, we cannot apply the same condition to UNREGISTER_SUPER if I am not mistaken here.

Would you please correct that in the code – and do a`add`, `commit` and `push` (to the same branch, it automatically gets into this pull request)?

randomize00:
OK. Fixed.

Logan007:
Thank you!

PR #623, opened 2021-02-02
reworked edge management port output

Logan007:
I have started to slightly rework edge's management port output. What do you think?

Any suggestions for how to neatly format the additional stats?


Logan007:
supernode management port output rework to follow sometime later

PR #627, opened 2021-02-05
added TCP support

### LINKED ISSUE #131: Camo Style for Privacy
- Startled by [current](https://gizmodo.com/russia-plans-to-block-major-vpns-as-next-step-in-closin-1835326332) news of which similar ones seem to lately pop up more and more often,
- valuing privacy,
- being somewhat impressed by [professional](https://wikileaks.org/vault8/#Hive) ideas of effective cover, and
- remembering the former `-t` option of n2n v1, ...

...I was wondering if there is need for (optionally) putting some camouflage on n2n's traffic between edges but also _between edges and supernode(s)_; of which the latter does not seem to be encrypted at all. Also, I was thinking about how to provide supernode functionality only to authorized edges while offering a different service to non-authorized users.

Is there any need for such a feature set? Would https-tunneling be a useful option to consider?

#30 might be related.

### LINKED ISSUE #144: Add support for TCP encapsulation
Hi, [on your website you say](https://www.ntop.org/products/n2n/):

> Q. n2n is based on UDP. Do you have plans for supporting HTTP encapsulation?
A. Currently n2n supports an initial HTTP tunneling. In order to use it you need to start the edge node with -t (the supernode already listens on both TCP and UDP).

But in reality `CMD ["supernode", "-l", "8443", "-vv"]` listens only on UDP sockets:

    (n2n)[root@4e1ae97bdf7d /]$ lsof -i -P -n                                                                                                                                    
    COMMAND   PID USER   FD   TYPE DEVICE SIZE/OFF NODE NAME                                                                                                                     
    supernode   1 root    3u  IPv4  68720      0t0  UDP *:8443                                                                                                                   
    supernode   1 root    4u  IPv4  68721      0t0  UDP 127.0.0.1:5645                                                                                                           
    (n2n)[root@4e1ae97bdf7d /]$

There is no option to enable TCP mentioned in the `supernode --help`, I am using latest CentOS 7 packages from your repository:


    (n2n)[root@4e1ae97bdf7d /]$ supernode --help                                                                                                                                 
    Welcome to n2n v.2.4 for x86_64                                                                                                                                              
    Built on Jun 20 2019 18:50:21                                                                                                                                                
    Copyright 2007-18 - ntop.org and contributors                                                                                                                                                                                                                                                                                                               
    supernode  (see supernode.conf)                                                                                                                                 
    or                                                                                                                                                                           
    supernode -l  [-f] [-v]                                                                                                                                                                                                                                                                                                                            
    -l       Set UDP main listen port to                                                                                                                           
    -v              Increase verbosity. Can be used multiple times.                                                                                                              
    -h              This help message.                                                                                                                                                                                                                                                                                                                        
    (n2n)[root@4e1ae97bdf7d /]$
--

Logan007:
Last night, I read about network programming in C. The next thing I remember is having created this pull request... :woozy_face: 

It adds the ability to have the edge connect to the supernode through TCP. This behavior can be triggered by `-S2` at the edge. Of course, this does not allow for peer-to-peer connections anymore, just like with the regular `-S` – or the new synonymous `-S1`. Furthermore, supernode and edge need to be replaced in order to run it. The supernode accepts both UDP and TCP connections from edges.

It should run only somewhat stable and maybe even in a federation environment. But, as it still lacks a few known things as annotated in the code and some more unknown, I mark it as _draft_. It definitely is _work in progress_.

However, I would appreciate any testing and especially your feedback from testing. Just write it down in the comments below or leave a code review with hints. Any contribution for further refinement is very welcome!

Closes #131
Closes #144

### DIFF
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -219,11 +219,14 @@ int time_stamp_verify_and_update (uint64_t stamp, uint64_t * previous_stamp, int
 /* Operations on peer_info lists. */
 size_t purge_peer_list (struct peer_info ** peer_list,
                         SOCKET socket_not_to_close,
+                        n2n_tcp_connection_t *tcp_connections,
                         time_t purge_before);
+
 size_t clear_peer_list (struct peer_info ** peer_list);
 
 size_t purge_expired_nodes (struct peer_info **peer_list,
                             SOCKET socket_not_to_close,
+                            n2n_tcp_connection_t *tcp_connections,
                             time_t *p_last_purge,
                             int frequency, int timeout);
 
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -106,6 +106,7 @@
 #include <netinet/in.h>
 #include <netinet/ip.h>
 #include <netinet/udp.h>
+#include <netinet/tcp.h>
 #include <arpa/inet.h>
 #include <sys/types.h>
 #include <sys/time.h>
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -207,7 +207,7 @@ void print_edge_stats (const n2n_edge_t *eee);
 char* sock_to_cstr (n2n_sock_str_t out,
                     const n2n_sock_t * sock);
 char * ip_subnet_to_str (dec_ip_bit_str_t buf, const n2n_ip_subnet_t *ipaddr);
-SOCKET open_socket (int local_port, int bind_any);
+SOCKET open_socket (int local_port, int bind_any, int type);
 int sock_equal (const n2n_sock_t * a,
                 const n2n_sock_t * b);
 
@@ -218,9 +218,14 @@ int time_stamp_verify_and_update (uint64_t stamp, uint64_t * previous_stamp, int
 
 /* Operations on peer_info lists. */
 size_t purge_peer_list (struct peer_info ** peer_list,
+                        SOCKET socket_not_to_close,
                         time_t purge_before);
 size_t clear_peer_list (struct peer_info ** peer_list);
-size_t purge_expired_nodes (struct peer_info **peer_list, time_t *p_last_purge, int frequency, int timeout);
+
+size_t purge_expired_nodes (struct peer_info **peer_list,
+                            SOCKET socket_not_to_close,
+                            time_t *p_last_purge,
+                            int frequency, int timeout);
 
 /* Edge conf */
 void edge_init_conf_defaults (n2n_edge_conf_t *conf);
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -133,6 +133,8 @@
 #include "n2n_typedefs.h"
 
 #ifdef WIN32
+#include <winsock2.h>           /* for tcp */
+#define SHUT_RDWR      SD_BOTH  /* for tcp */
 #include "win32/wintap.h"
 #include <sys/stat.h>
 #else
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -45,7 +45,7 @@
 #endif
 #define N2N_CAN_NAME_IFACE 1
 #undef N2N_HAVE_DAEMON
-#undef N2N_HAVE_TCP
+#undef N2N_HAVE_TCP           /* as explained on https://github.com/ntop/n2n/pull/627#issuecomment-782093706 */
 #undef N2N_HAVE_SETUID
 #else
 #ifndef CMAKE_BUILD
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -32,6 +32,7 @@
 */
 
 #define N2N_HAVE_DAEMON /* needs to be defined before it gets undefined */
+#define N2N_HAVE_TCP    /* needs to be defined before it gets undefined */
 
 /* #define N2N_CAN_NAME_IFACE */
 
@@ -44,6 +45,7 @@
 #endif
 #define N2N_CAN_NAME_IFACE 1
 #undef N2N_HAVE_DAEMON
+#undef N2N_HAVE_TCP
 #undef N2N_HAVE_SETUID
 #else
 #ifndef CMAKE_BUILD
@@ -134,7 +136,8 @@
 
 #ifdef WIN32
 #include <winsock2.h>           /* for tcp */
-#define SHUT_RDWR      SD_BOTH  /* for tcp */
+#define SHUT_RDWR   SD_BOTH     /* for tcp */
+#define SOL_TCP     IPPROTO_TCP /* for tcp */
 #include "win32/wintap.h"
 #include <sys/stat.h>
 #else
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -108,11 +108,13 @@ enum sn_purge{SN_PURGEABLE = 0, SN_UNPURGEABLE = 1};
 #define HASH_FIND_PEER(head,mac,out) \
     HASH_FIND(hh,head,mac,sizeof(n2n_mac_t),out)
 #define N2N_EDGE_SN_HOST_SIZE     48
-#define N2N_EDGE_NUM_SUPERNODES   2
 #define N2N_EDGE_SUP_ATTEMPTS     3             /* Number of failed attmpts before moving on to next supernode. */
 #define N2N_PATHNAME_MAXLEN       256
 #define N2N_EDGE_MGMT_PORT        5644
 #define N2N_SN_MGMT_PORT          5645
+
+#define N2N_TCP_BACKLOG_QUEUE_SIZE   3         /* number of concurrently pending connections to be accepted */
+                                               /* NOT the number of max. TCP connections                    */
 
 /* flag used in add_sn_to_list_by_mac_or_sock */
 enum skip_add{SN_ADD = 0, SN_ADD_SKIP = 1, SN_ADD_ADDED = 2};
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -698,6 +698,10 @@ typedef struct n2n_tcp_connection {
     SOCKET socket_fd;       /* file descriptor for tcp socket */
     struct sockaddr sock;   /* network order socket */
 
+    uint16_t expected;                                    /* number of bytes expected to be read */
+    uint16_t position;                                    /* current position in the buffer*/
+    uint8_t  buffer[N2N_PKT_BUF_SIZE + sizeof(uint16_t)]; /* buffer for data collected from tcp socket incl. prepended length */
+
     UT_hash_handle hh; /* makes this structure hashable */
 } n2n_tcp_connection_t;
 
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -415,6 +415,7 @@ struct peer_info {
     n2n_ip_subnet_t                  dev_addr;
     n2n_desc_t                       dev_desc;
     n2n_sock_t                       sock;
+    SOCKET                           socket_fd;
     n2n_cookie_t                     last_cookie;
     n2n_auth_t                       auth;
     int                              timeout;
@@ -598,6 +599,7 @@ typedef struct n2n_edge_conf {
     int                register_ttl;           /**< TTL for registration packet when UDP NAT hole punching through supernode. */
     int                local_port;
     int                mgmt_port;
+    uint8_t            connect_tcp;            /** connection to supernode 0 = UDP; 1 = TCP */
     n2n_auth_t         auth;
     filter_rule_t      *network_traffic_filter_rules;
     int                metric;                /**< Network interface metric (Windows only). */
@@ -692,6 +694,12 @@ struct sn_community_regular_expression {
     UT_hash_handle hh; /* makes this structure hashable */
 };
 
+typedef struct n2n_tcp_connection {
+    SOCKET socket_fd;  /* file descriptor for tcp socket */
+
+    UT_hash_handle hh; /* makes this structure hashable */
+} n2n_tcp_connection_t;
+
 typedef struct n2n_sn {
     time_t                                 start_time;      /* Used to measure uptime. */
     sn_stats_t                             stats;
@@ -700,7 +708,9 @@ typedef struct n2n_sn {
     uint16_t                               lport;           /* Local UDP port to bind to. */
     uint16_t                               mport;           /* Management UDP port to bind to. */
     int                                    sock;            /* Main socket for UDP traffic with edges. */
-    int                                    mgmt_sock;         /* management socket. */
+    int                                    tcp_sock;        /* auxiliary socket for optional TCP connections */
+    n2n_tcp_connection_t                   *tcp_connections;/* list of established TCP connections */
+    int                                    mgmt_sock;       /* management socket. */
     n2n_ip_subnet_t                        min_auto_ip_net; /* Address range of auto_ip service. */
     n2n_ip_subnet_t                        max_auto_ip_net; /* Address range of auto_ip service. */
 #ifndef WIN32
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -695,7 +695,8 @@ struct sn_community_regular_expression {
 };
 
 typedef struct n2n_tcp_connection {
-    SOCKET socket_fd;  /* file descriptor for tcp socket */
+    SOCKET socket_fd;       /* file descriptor for tcp socket */
+    struct sockaddr sock;   /* network order socket */
 
     UT_hash_handle hh; /* makes this structure hashable */
 } n2n_tcp_connection_t;
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -694,6 +694,7 @@ struct sn_community_regular_expression {
     UT_hash_handle hh; /* makes this structure hashable */
 };
 
+
 typedef struct n2n_tcp_connection {
     SOCKET socket_fd;       /* file descriptor for tcp socket */
     struct sockaddr sock;   /* network order socket */
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -633,7 +633,7 @@ struct n2n_edge {
 
     /* Sockets */
     /* supernode socket is in        eee->curr_sn->sock (of type n2n_sock_t) */
-    int                              udp_sock;
+    int                              sock;
     int                              udp_mgmt_sock;                      /**< socket for status info. */
 
 #ifndef SKIP_MULTICAST_PEERS_DISCOVERY
--- a/include/sn_selection.h
+++ b/include/sn_selection.h
@@ -26,6 +26,7 @@ typedef char selection_criterion_str_t[SN_SELECTION_CRITERION_BUF_SIZE];
 /* selection criterion's functions */
 int sn_selection_criterion_init (peer_info_t *peer);
 int sn_selection_criterion_default (SN_SELECTION_CRITERION_DATA_TYPE *selection_criterion);
+int sn_selection_criterion_bad (SN_SELECTION_CRITERION_DATA_TYPE *selection_criterion);
 int sn_selection_criterion_calculate (n2n_edge_t *eee, peer_info_t *peer, SN_SELECTION_CRITERION_DATA_TYPE *data);
 
 /* common data's functions */
--- a/src/edge.c
+++ b/src/edge.c
@@ -47,7 +47,9 @@ void send_register_super (n2n_edge_t *eee);
 void send_query_peer (n2n_edge_t *eee, const n2n_mac_t dst_mac);
 int supernode_connect (n2n_edge_t *eee);
 int supernode_disconnect (n2n_edge_t *eee);
-
+int fetch_and_eventually_process_data (n2n_edge_t *eee, SOCKET sock,
+                                       uint8_t *pktbuf, uint16_t *expected, uint16_t *position,
+                                       time_t now);
 /* ***************************************************** */
 
 /** Find the address and IP mode for the tuntap device.
@@ -841,10 +843,16 @@ int main (int argc, char* argv[]) {
     n2n_tuntap_priv_config_t ec;  /* config used for standalone program execution */
     uint8_t runlevel = 0;         /* bootstrap: runlevel */
     uint8_t seek_answer = 1;      /*            expecting answer from supernode */
-    time_t now_time, last_action; /*            timeout */
+    time_t now, last_action;      /*            timeout */
     macstr_t mac_buf;             /*            output mac address */
     fd_set socket_mask;           /*            for supernode answer */
     struct timeval wait_time;     /*            timeout for sn answer */
+
+    size_t   bread = 0;
+    uint16_t expected = sizeof(uint16_t);
+    uint16_t position = 0;
+    uint8_t  pktbuf[N2N_SN_PKTBUF_SIZE + sizeof(uint16_t)]; /* buffer + prepended buffer length in case of tcp */
+
 
 #ifndef WIN32
     struct passwd *pw = NULL;
@@ -964,12 +972,12 @@ int main (int argc, char* argv[]) {
 
     while(runlevel < 5) {
 
-        now_time = time(NULL);
+        now = time(NULL);
 
         // we do not use switch-case because we also check for 'greater than'
 
         if(runlevel == 0) { /* PING to all known supernodes */
-            last_action = now_time;
+            last_action = now;
             eee->sn_pong = 0;
             // (re-)initialize the number of max concurrent pings (decreases by calling send_query_peer)
             eee->conf.number_max_sn_pings = NUMBER_SN_PINGS_INITIAL;
@@ -987,7 +995,7 @@ int main (int argc, char* argv[]) {
                 supernode_connect(eee);
                 traceEvent(TRACE_NORMAL, "Received first PONG from supernode [%s].", eee->curr_sn->ip_addr);
                 runlevel++;
-            } else if(last_action <= (now_time - BOOTSTRAP_TIMEOUT)) {
+            } else if(last_action <= (now - BOOTSTRAP_TIMEOUT)) {
                 // timeout
                 runlevel--;
                 // skip waiting for answer to direcly go to send PING again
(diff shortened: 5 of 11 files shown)


skyformat99:
thanks, great job!
Can the two modes be automatically switched?
If the connection cannot be established using UDP mode, it will automatically switch to TCP mode to establish the connection.

Logan007:
> Can the two modes be automatically switched?

That is a very interesting idea!

The current design requires a decision at the beginning, that is even for bootstrap, and then needs to stick with it. I am not sure if the socket type can be switched (close and re-open of the other type) back and forth between UDP and TCP while program is running – but I will try it once TCP is reliably working.

> thanks, great job!

Is it working for you?

tlsalex:
This is my first time to touch v2.9.xx, and I want to do a test for TCP support, I try to get help from supernode --help, it does not show me the usage about TCP things, could you please help ?

```
xx@xx:~/temp/n2n# ./supernode -h

Welcome to n2n v.2.9.0.r873.c73e593 for Debian buster/sid
Built on Feb  9 2021 08:50:13
Copyright 2007-2021 - ntop.org and contributors

 general usage:  supernode <config file> (see supernode.conf)

            or   supernode [-p <local port>]
                           [-F <federation name>]
 options for under-        [-l <supernode host:port>]
 lying connection

 overlay network           [-c <community list file>]
 configuration             [-a <net ip>-<net ip>/<cidr suffix>]

 local options             [-f] [-t <management port>] [-v]
                           [-u <numerical user id>][-g <numerical group id>]

 meaning of the            [-f]  do not fork but run in foreground
 flag options              [-v]  make more verbose, repeat as required

 technically, all parameters are optional, but the supernode executable
 requires at least one parameter to run, .e.g. -v or -f, as otherwise a
 short help text is displayed

  -h    shows this quick reference including all available options
 --help gives a detailed parameter description
   man  files for n2n, edge, and superndode contain in-depth information
```

tlsalex:
BTW, when using TCP , does n2n include encryption for communication ?

Logan007:
> supernode --help

It's not a supernode's decision to use TCP, is's made at the **edge** by using `-S2` or not.

`./edge -h` should show

```
...
                      [-S<level of solitude>]
...
```

in the first category of options. And `./edge --help` should give more details

```
...
 OPTIONS FOR THE UNDERLYING NETWORK CONNECTION
 ---------------------------------------------
...
 -S1 ... -S2 or -S | -S1 or -S do not connect p2p, always use the supernode
                   | -S2 connects through TCP and only to the supernode
...
```

The **supernode** just accepts TCP in addition to UDP (same port). The most recent push includes a `TCP`  indicator on the supernode's management port output.

> BTW, when using TCP , does n2n include encryption for communication ?

Yes, only difference is the use of TCP instead of UDP. Encryption schemes including header encryption remain the same.

A side effect is that TCP can not ping the other supernodes because that would require to open a new / additional connection for each ping. TCP connection remains restricted to one supernode at a time, i.e. all communication needs to run through this supernode (no p2p). But the edge can switch to other supernodes of the federation in case of connection failure. Being limited to one connection is a property inherent with TCP being a connection-based protocol.

tlsalex:
Hello @Logan007 ,
I can not see SUPERNODE listen on TCP port (from ss -tln), only see UDP traffic even I use -S2 on both two edge side.
this is my conf for one of my edge:
```
-c=xxxxnetwork
-k=xxxxxxxxxxx
-m=DE:AD:BE:EF:99:11
-l=xx.xx.xx.xx:8080
-a=static:192.168.99.11
-p=30011
-S2
-r
-f
```
And I run SUPERNODE like this: `./supernode -p 8080 -f -v` 


And my `./edge --help` is like below, not like yours,
```
 -D                | enable PMTU discovery, it can reduce fragmentation but
                   | causes connections to stall if not properly supported
 -S                | do not connect p2p, always use the supernode
 -i <reg_interval> | registration interval, for NAT hole punching (default
                   | 20 seconds)

```
it's because the latest ce61945 or  2513c7a have not been pushed (in fact I don't know what does it mean, just guess)?  then cause my build does not have TCP support ?

Logan007:
> And my ./edge --help is like below, not like yours,

Then, you need to re-build supernode and edge from this commit 2513c7a (not from _dev_, as this pull request has not been merged into _dev_ branch yet).

> in fact I don't know what does it mean

If the pull request is not accepted / merged into _dev_ branch yet, the _dev_ branch does not contain the code -- for TCP support, in this case. So, you need to use the code from this pull request. To download a complete `.zip` containing the code, try [this link](https://github.com/ntop/n2n/archive/2513c7a.zip) (better use _Right click_ + _Save as..._) with 2513c7a being the short SHA value of the commit you want to test.

@tlsalex

tlsalex:
I got issue if I test in this way: 
Run edge on my homepc , run supernode on server, also run an edge on server, all edges with -S2

Homepc_Edge    --------------->   Server_Suppernode  <------------- Sever_Edge

edge: 192.168.9.11:3011, and -l=x.x.x.x:8080 ------>   Supernode: 0.0.0.0:8080  <--------192.168.9.22:3022, but -l=127.0.0.1:8080

If I login to Server from homepc, then ping 192.168.9.11, in the first time it may work ,seond time also may work , but 3rd, 4rd,xxxx will fail.

If I ping 192.168.9.22 from homepc , seems it always work, but another issue comes:  delay is quite different,maybe it's performance issue of my server, but maybe not.
not go through n2n `rtt min/avg/max/mdev = 183.675/228.768/283.883/29.413 ms` (ping server IP rather than 192.168.9.22)
when go through n2n `rtt min/avg/max/mdev = 197.727/340.494/1238.464/246.082 ms, pipe 2` (ping 192.168.9.22)

supernode
```
10/Feb/2021 16:23:26 [sn_utils.c:1998] ERROR: recvfrom() failed 0 errno 107 (Transport endpoint is not connected)
10/Feb/2021 16:23:33 [sn_utils.c:2040] run_sn_loop accepted incoming TCP connection from 127.0.0.1:18037
10/Feb/2021 16:23:33 [sn_utils.c:1998] ERROR: recvfrom() failed 0 errno 107 (Transport endpoint is not connected)
10/Feb/2021 16:23:35 [sn_utils.c:1049] Processing incoming UDP packet [len: 91][sender: xxx.140.31.24:20995]
10/Feb/2021 16:23:35 [sn_utils.c:1465] Rx REGISTER_SUPER for DE:AD:BE:EF:99:11 [xxx.140.31.24:20995]
10/Feb/2021 16:23:35 [sn_utils.c:494] update_edge for DE:AD:BE:EF:99:11 [xxx.140.31.24:20995]
10/Feb/2021 16:23:35 [sn_utils.c:554] update_edge unchanged DE:AD:BE:EF:99:11 ==> xxx.140.31.24:20995
10/Feb/2021 16:23:35 [sn_utils.c:233] try_broadcast
10/Feb/2021 16:23:35 [sn_utils.c:176] sendto_sock sent=50 to 
10/Feb/2021 16:23:35 [sn_utils.c:1538] Tx REGISTER_SUPER_ACK for DE:AD:BE:EF:99:11 [xxx.140.31.24:20995]
10/Feb/2021 16:23:39 [sn_utils.c:2040] run_sn_loop accepted incoming TCP connection from 127.0.0.1:18037
10/Feb/2021 16:23:39 [n2n.c:455] Purging old registrations
10/Feb/2021 16:23:39 [n2n.c:460] Remove 0 registrations
10/Feb/2021 16:23:39 [sn_utils.c:1998] ERROR: recvfrom() failed 0 errno 107 (Transport endpoint is not connected)
10/Feb/2021 16:23:45 [sn_utils.c:2040] run_sn_loop accepted incoming TCP connection from 127.0.0.1:18037
10/Feb/2021 16:23:45 [sn_utils.c:1998] ERROR: recvfrom() failed 0 errno 107 (Transport endpoint is not connected)
10/Feb/2021 16:23:45 [sn_utils.c:1049] Processing incoming UDP packet [len: 36][sender: xxx.140.31.24:20995]
10/Feb/2021 16:23:45 [sn_utils.c:1790] Rx PING from DE:AD:BE:EF:99:11.
10/Feb/2021 16:23:45 [sn_utils.c:176] sendto_sock sent=50 to 
10/Feb/2021 16:23:45 [sn_utils.c:1818] Tx PONG to DE:AD:BE:EF:99:11
```
edge on server
```
10/Feb/2021 16:23:22 [edge_utils.c:1840] Tx PACKET to 127.0.0.1:8080 (dest=DE:AD:BE:EF:99:11) [96 B]
10/Feb/2021 16:23:23 [edge_utils.c:1840] Tx PACKET to 127.0.0.1:8080 (dest=DE:AD:BE:EF:99:11) [96 B]
10/Feb/2021 16:23:23 [edge_utils.c:1289] Registering with supernode [127.0.0.1:8080][number of supernodes 1][attempts left 0]
10/Feb/2021 16:23:24 [edge_utils.c:1840] Tx PACKET to 127.0.0.1:8080 (dest=DE:AD:BE:EF:99:11) [96 B]
10/Feb/2021 16:23:26 [edge_utils.c:1266] WARNING: Supernode not responding, now trying 127.0.0.1:8080
10/Feb/2021 16:23:26 [edge_utils.c:202] Binding to local port 30022
10/Feb/2021 16:23:26 [edge_utils.c:202] Binding to local port 30022
10/Feb/2021 16:23:26 [edge_utils.c:1289] Registering with supernode [127.0.0.1:8080][number of supernodes 1][attempts left 1]
10/Feb/2021 16:23:30 [edge_utils.c:1289] Registering with supernode [127.0.0.1:8080][number of supernodes 1][attempts left 0]
10/Feb/2021 16:23:32 [edge_utils.c:1840] Tx PACKET to 127.0.0.1:8080 (dest=FF:FF:FF:FF:FF:FF) [96 B]
10/Feb/2021 16:23:33 [edge_utils.c:1840] Tx PACKET to 127.0.0.1:8080 (dest=FF:FF:FF:FF:FF:FF) [96 B]
10/Feb/2021 16:23:33 [edge_utils.c:1266] WARNING: Supernode not responding, now trying 127.0.0.1:8080
```

Logan007:
@tlsalex, thank you for testing and especially your feedback!

From the log it seems that the TCP connection gets lost in between (107) quite often and needs to be rebuilt (binding, registering,...). I will try to find reasons for connection loss in the code. So far, I have not been aware of it as not observed it in my test-setup too often yet. I might add some more log messages to support debugging.

Logan007:
I should have read Part III as well... :books:

For TCP, we need to prepend packet size, introduce packet delimiters, other kind of markers, make packets same size, or use additional GCR-style coding for self-synchronization on gaps between packets... Even `TCP_NODELAY` option does not guarantee that chunks arriving  either at the supernode or at the edge correspond to the packet sizes. I will think of something, it might solve observed throughput issues.

Logan007:
The latest commit 75e89da (code [here](https://github.com/ntop/n2n/archive/75e89da.zip)) looks much more stable and very promising.

Please test and let me know about any issue you might encounter.

tlsalex:
Hello @Logan007 

The first issue which is "connection lost"  has gone!
But second issue still exist and even worse, I am not sure whether it's n2n cause a big delay.

The same setup as my last one,
Homepc_Edge ---------------> Server_Suppernode <------------- Sever_Edge

ping edge (IP of edge0)on the server from homepc ,network delay is around 622ms,
`go through n2n: rtt min/avg/max/mdev = 551.784/622.309/1104.406/161.779 ms, pipe 2`

ping the server(public IP) directly from homepc, network delay is around 167ms,
`not go through n2n:  rtt min/avg/max/mdev = 167.011/167.533/168.955/0.545 ms`

The delay is quite different. 167ms much much smaller than  622ms.

PS: I also tried to disable encryption with -A1 on both edges . but the issue still exist , it should not be a performance issue of both edges.

Logan007:
@tlsalex, thank you very, very much for testing!! :muscle: 

I have it run for a while now and I have not been able to trigger any unstable behavior so far.

Concerning the ping **delay**, I can confirm your observation and I think there are reasons for it:

1. Both `-S_` options forward through the supernode. So, part of the delay is caused by the supernode which performs some other tasks in between and needs to be woken up by the OS or ... This additional instance on packet's way causes some delay. You can even see it if you compare pure p2p and "regular" UDP-forwarding using `-S` or `-S1` or just no p2p possible.
2. TCP has some overhead and also allows some more OS control over how data is sent out. So, a packet might not immediately be sent out but maybe at a more suitable moment when sufficient data (more packets) are available.

In a local setup, I get the following numbers:
```
     min  avg  max
p2p  2.2  2.8  6.7 ms 
udp  2.1   15  190 ms   -S1 on both edges
tcp  5.3   30  133 ms   -S2 on both edges
```

The longtime average shows these effects very well and I think it is something we should have expected.

For further testing, I checked on the data flow for continuous data transmission by `scp`ing a larger file over the line (`scp` outputs the achieved average data rate when finished). I saw the following already averaged (from several rounds) **throughput** values:

```
     avg
p2p  11.69 MB/s 
udp  11.63 MB/s   -S1 on both edges
tcp  10.20 MB/s   -S2 on both edges
```

I think that what we see here is the more complicated TCP handling plus the TCP inherent protocol overhead – a total of 12 % lower performance in terms of throughput observed. Your mileage may vary.

And finally, another effect kicks in. n2n works by passing around Ethernet frames. Those have a certain size and are self-contained complete. So, UDP packets are the most natural match to encapsulate the Ethernet frames. As opposed to TCP which is better suited for transmission of continuous streams such as files. A UDP packet is directly sent out just like the Ethernet frame would. With TCP you never can be sure when and how the data gets sent, if in one piece or several, ...

For the n2n use case, my conclusion would be to use TCP as fall-back in case UDP is not possible.

I will go on and finish the still open action items as per annotations in the code. I would be glad to hear about any issue you encounter so I can try to solve it along the way while being at it.

tlsalex:
>  So, a packet might not immediately be sent out but maybe at a more suitable moment when sufficient data (more packets) are available.

I am not quite agree on this one ,  when I do a normal ping to my server (public IP) , seems it doesn't need to wait for more package to make its queue has sufficient data , and send them out, looks like once server receive icmp request ,then server response immediately. maybe the design of ICMP and TCP for package delivery is quite different. 

I wonder is there any flag in TCP header can optimize the delivery.

Logan007:
> when I do a normal ping to my server (public IP) , seems it doesn't need to wait for more package to make its queue has sufficient data

ICMP sends out directly, TCP [not necessarily](https://en.wikipedia.org/wiki/Nagle%27s_algorithm).

> I wonder is there any flag in TCP header can optimize the delivery

Not in the header flags but the socket options might support the [earlier](https://github.com/ntop/n2n/pull/627#issuecomment-777312885) mentioned `TCP_NODELAY` as well as the antagonistic `TCP_CORC`. I actually tried them and they show some significant better results as they imitate the UDP behaviour as far as the OS allows which does not always seems to be the case. As their use is [discouraged](https://linux.die.net/man/7/tcp) for portability, I am not sure if and how we can safely use them. However, I have added them to b8d3087 for test. Let me know what you find.

Logan007:
– **HELP WANTED** – 

This whole pull request does **not** compile on Windows following our [building instructions](https://github.com/ntop/n2n/blob/dev/doc/Building.md#build-on-windows).

Is there anyone able to advise? :construction_worker: :firefighter: :mage: 

– **HELP WANTED** –

tlsalex:
I still test the commit of 75e89da

I still want to confirm whether the big ping delay is cause by TCP mechanism.
I tried to ping with 65507 byte package in size , which is a big package (63KB), which should be big enough to fill full the tcp queue , should no more waiting to send out the packages. but the delay is still very big.

So I wonder , how big it is for the TCP queue , which size it need to reach before it send packages out?

Of course I will test the latest commit soon.

```
[root@n2n01 ~]# ping -c2 -s 90000 192.168.99.22
ping: packet size 90000 is too large. Maximum is 65507

[root@n2n01 ~]# ping -c2 -s 65507 192.168.99.22
PING 192.168.99.22 (192.168.99.22) 65507(65535) bytes of data.
65515 bytes from 192.168.99.22: icmp_seq=1 ttl=64 time=719 ms
65515 bytes from 192.168.99.22: icmp_seq=2 ttl=64 time=527 ms

--- 192.168.99.22 ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1000ms
rtt min/avg/max/mdev = 527.196/622.992/718.789/95.796 ms
```

tlsalex:
Could any one tell me how to git clone a specific commit in order to compile it. I check from google ,different people have different way, not sure which one is correct.
```
git clone https://github.com/ntop/n2n.git
cd n2n
git reset --hard 69021895e8599e5c3e627901e7bba93592fc8e04  
fatal: Could not parse object '69021895e8599e5c3e627901e7bba93592fc8e04'.--doesn't work for me
git checkout 69021895e8599e5c3e627901e7bba93592fc8e04
fatal: reference is not a tree: 69021895e8599e5c3e627901e7bba93592fc8e04 --doesn't work for me

```
https://www.freecodecamp.org/news/git-checkout-explained/
https://www.teosoft.it/note/2018-07-27-git-clone-a-specific-commit/

skyformat99:
> Could any one tell me how to git clone a specific commit in order to compile it. I check from google ,different people have different way, not sure which one is correct.
> 
> ```
> git clone https://github.com/ntop/n2n.git
> cd n2n
> git reset --hard 69021895e8599e5c3e627901e7bba93592fc8e04  
> fatal: Could not parse object '69021895e8599e5c3e627901e7bba93592fc8e04'.--doesn't work for me
> git checkout 69021895e8599e5c3e627901e7bba93592fc8e04
> fatal: reference is not a tree: 69021895e8599e5c3e627901e7bba93592fc8e04 --doesn't work for me
> ```
> 
> https://www.freecodecamp.org/news/git-checkout-explained/
> https://www.teosoft.it/note/2018-07-27-git-clone-a-specific-commit/


https://github.com/Logan007/n2n/tree/tcp
or
https://github.com/Logan007/n2n/archive/tcp.zip

Logan007:
> So I wonder , how big it is for the TCP queue , which size it need to reach before it send packages out?

I would say that TCP-specific [Nagle's algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) mainly depends on the network line properties as well as local and also remote computer's network stack – which is a good thing because it is made to help transferring streams (not packets) as seamlessly and continuously as possible.

> I still want to confirm whether the big ping delay is cause by TCP mechanism.

It is. Please compare fb62eb5 to b8d3087 of which the latter disables Nagle's algorithm (that is what `TCP_NODELAY` does) and additionally uses a corc, I see an astonishing difference.

>  how to git clone a specific commit in order to compile it

If you only want the code, you could try `https://github.com/ntop/n2n/archive/<7 digit commit hash>.zip`, e.g. https://github.com/ntop/n2n/archive/6902189.zip.

Other than that, `git` command line using `git checkout <7 digit hash value>` works well. Note that for checking out commits made to this pull request, you should do it from a clone of **Logan007**'s repository as this pull request is not merged into n2n's repository  yet. I just tested a fresh clone

```
git clone https://github.com/Logan007/n2n.git n2nTest
```

with `n2nTest` being the directory I want to have it in – without it, it would be cloned into `n2n` by default.

And then,

```
cd n2nTest
git checkout 86b6e73
```

works well.

tlsalex:
thanks @Logan007 ,  at this moment, in my test that build on the commit of 6902189,   the connection issue is gone, the connectivity is good , and the network delay is within the normal range, I don't see other issue about TCP connection, I will do more test later.

However, I met a new issue. when I add  -n option (in my case -n=0.0.0.0/0:192.168.99.22) in the edge.conf. It will complain bellow info:

```
18/Feb/2021 22:46:59 [edge_utils.c:3049] ERROR: [err=101] route: Add 0.0.0.0/1 via 192.168.99.22
18/Feb/2021 22:46:59 [edge_utils.c:367] ERROR: routes setup failed
18/Feb/2021 22:46:59 [edge.c:941] ERROR: Failed in edge_init

```
I also run SUPERNODE  in daemon mode on my server , I want to know how to log everything generated by supernode to a file. could we add an option to the configure file ,to specify the location of that log file and instruct SUPERNODE log to that file.

Logan007:
Please file new Issues; one for the routing, and one for the log – if still required: I think that the log issue could easily be solved at the command line level without adding a whole new option.

Have you tried `supernode .... > log.file`?

tlsalex:
I already raise two new issues. and I have another question:

In TCP mode, does n2n will generate a session key(key will be changed from time to time) derived from the key specified in the -k option to encrypt payload?  or it will always use the same key specified in the -k option to encrypt data ,never change . will it possible encrypt payload with a certification ,just like TLS , maybe this is a hard work , maybe this can implement in n2n V4.0.

Logan007:
> In TCP mode, does n2n will generate a session key(key will be changed from time to time) derived from the key specified in the -k option to encrypt payload? or it will always use the same key specified in the -k option to encrypt data ,never change . 

Encryption of TCP's packet-sized chunks happens the same way as for UDP packets using a pre-shared key. There are no session keys negotiated. The actual key going into the bare metal cipher is a hash of the user-provided key. All ciphers use random IVs. Details can be found in our [crypto doc](https://github.com/ntop/n2n/blob/dev/doc/Crypto.md).

> will it possible encrypt payload with a certification ,just like TLS , maybe this is a hard work , maybe this can implement in n2n V4.0.

Some might find it advantageous **not** needing to deal with certificates but handle a pre-shared key only. It is just simpler.

4.0 has not even been drafted yet. But as it might go even more towards pure p2p in terms of all nodes provide same function 
(supernodes too), edge authentication could become a  more important aspect. Current 2.9 / upcoming 3.0 has some space left for implementation of edge authentication mechanisms that do not require certificates.

However, even for public-key cryptography, you need to keep every participant's private key secret. And for certificates, you need infrastructure to easily have their validity checked and keep track of a possible revocation. So, you are right, it would require some hard work to holistically switch to public-key cryptography and, at the same time, to keep it user-friendly.

Logan007:
Still trying to compile binaries for Windows. I found what it takes to successfully compile which unfortunately does not mean that the resulting executable runs as expected...

- I included Winsock2 and set missing defines in `n2n.h` inside some `#ifdef WIN32` section:
```
#include <winsock2.h>
#  define SHUT_RDWR      SD_BOTH
#ifndef SOL_TCP
    #define SOL_TCP      SOL_SOCKET             /* also tried IPPROTO_TCP */
#endif
```
- Windows does not seem to support the `TCP_CORK` socket option, so I defined out the corresponding `setsockopt(..., TCP_CORK, ...)` lines using `#ifndef WIN32` in `sn_utils.c` and `edge_utils.c`.

- Setting the socket to NONBLOCKing in Windows works differently as well:
 ```
#ifdef WIN32
            u_long value = 1;
            ioctlsocket(eee->sock, FIONBIO, &value);
#else
            fcntl(eee->sock, F_SETFL, O_NONBLOCK);
#endif
```

With these changes applied, it compiles well on Windows but `-S2` (TCP) does not let the edge connect as expected (while `-S1` for UDP does).

These outlined changes are not committed yet as I do not want to seduce anyone into thinking it would be working on Windows. Apart from this Windows-related issue, for now, I do not have any more changes to make to this feature as it is running very stable over here, even in a federated scenario – umm, well, on Linux. This is where my expertise ends.

**– HELP WANTED –**

P. S. _Alternatively, we could add this feature in a way it compiles on Windows but `#ifdef` out access to it on that platform until someone gets it working._

**Comment by tlsalex** [comment]
New issue:  
a running edge can not rebuild TCP connection with suprenode after a restart of  supernode, it happens very randomly: 

sometimes it will rebuilt TCP connection very soon ,
sometimes it took a long time to rebuild TCP connection, 
sometimes it never can connect to supernode. I have to stop both edge and supernode , then wait for a few minutes , then start supernode , and then start edge . the order is important , otherwise it may fail.

I thinks it's something about Three-Way Handshake and Four-Way Wavehand in TCP , or the network resource issue(like socket, I guess) , when restart supernode, it release some resources , but edge still connect the same resource, until timeout.

I just run one edge to test at this time,  edge and supernode on different host, if they are on the same host ,the issue seems is more obvious. I need to do more test to confirm.

some error info:
supernode:
```
Feb 20 16:44:48 xxx supernode[28953]: 20/Feb/2021 16:44:48 [sn_utils.c:2113] ERROR: recvfrom() failed 0 errno 0 (Success)
Feb 20 17:24:15 tus supernode[29290]: 20/Feb/2021 17:24:15 [sn_utils.c:2127] ERROR: too many bytes in tcp packet expected
```


edge:
```
Feb 20 16:44:16 xxx edge[28916]: 20/Feb/2021 16:44:16 [edge_utils.c:1300] WARNING: Supernode not responding, now trying 127.0.0.1:8080
Feb 20 16:44:16 xxx edge[28916]: 20/Feb/2021 16:44:16 [edge_utils.c:200] Binding to local port 30022
```

Logan007:
Do you use `-p <port>` at the edges? If so, please try without. I found a similar behavior only if I use `-p <port>` at the edges in conjucntion with TCP. I will commit a change which makes sure that port gets ignored for TCP.

If it still persists, how do your exact command lines look like?

tlsalex:
so far so good, when without -p option,
and there is a small thing confuses me when I test supernode.

when I telnet localhost 8080 , then send Ctrl + ] and then send "quit" to quit , it will show me: 
`Feb 20 20:28:55 xxx supernode[30176]: 20/Feb/2021 20:28:55 [sn_utils.c:2113] ERROR: recvfrom() failed 0 errno 107 (Transport endpoint is not connected)`
when I telnet localhost 8080 , then send Ctrl + "other key" except  " ] " , it will show me:
`Feb 20 20:28:27 xxx supernode[30176]: 20/Feb/2021 20:28:27 [sn_utils.c:2127] ERROR: too many bytes in tcp packet expected`

Despite this may not be an issue, these logs contains the keyword "Error" , which will confuse user.

Logan007:
> when I telnet localhost 8080 , then send Ctrl + "other key" except " ] " , it will show me:
> ERROR: too many bytes in tcp packet expected

I assume, you telnet to the regular supernode port, not the management port. To handle ethernet frames on "streamy" TCP, this implementation works as follows: It prepends two bytes indicating the length of the following data. And then, the packet data follows (of that same length) of which exactly the indicated length is read. As this might be subject to attacks or just hick-ups while reading, there of course are sanity checks of the length value so our buffer does not overflow. If those checks do not pass, the "too many bytes" message is output. By sending keystrokes, their ASCII values might indicate an out-of-bound length.

> Despite this may not be an issue, these logs contains the keyword "Error" , which will confuse user.

Yes, the log messages including their categories will be subject to revision one day. They also still contain "udp" only and do not fully reflect tcp yet. But that is something that should be done after merging because I am under the impression that all messages could afford revision. Better do that in one piece then.

Logan007:
Disabled TCP support on Windows supernode as well as edge for now. Just can't get it running.

Will merge to _dev_ for broader access to testing.

PR #640, opened 2021-02-19
added more targeted routing for packet forwarding in federation

Logan007:
This pull request adds a more targeted packet forwarding between supernodes of a federation. It uses "associations" to remember the supernodes associated to each remotely registered edge. To keep memory requirements low, the list does not contain full peer information but only the socket and a time stamp which allows regular purging – also with a view to memory usage.

The associations are generated and renewed with each REGISTER_SUPER which is forwarded into the federation. As this only happens from time to time (every other 20 secondes), the supernode snoops on the PEER_INFOs being forwarded to faster build up associations.

Once an association is present, the supernode does **not** forward the PACKETs to **all** other federated supernodes anymore but only to the one associated to the target edge. Of course, forwarding happens only if no p2p is possible / allowed.

Broadcasts however are still forwarded to all supernodes. Experiments have shown that it a same-style automated "community-supernode association" intially even prevents communication as it starts with empty tables and broadcasted ARPs do not get through. One idea is that some meta communication between the supernodes could be required to even avoid this, some kind of FEDERATION_TALK message format exchanging current communities and other stats.

This code is based on the TCP code #627, so this code should be merged **after** the TCP code will have been merged.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -6,6 +6,7 @@ GIT_COMMITS=@GIT_COMMITS@
 ########
 
 CC=@CC@
+AR=@AR@
 
 #Ultrasparc64 users experiencing SIGBUS should try the following gcc options
 #(thanks to Robert Gibbon)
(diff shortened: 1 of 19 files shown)


Logan007:
Can't get it straightly rebased back onto current dev. Git'n'me... Will open a new pull request.

PR #672, opened 2021-03-20
Fix the port config in supernode.conf is incorrect in README.md

maomiaomi:


Logan007:
@maomiaomi, thank you very much for your contribution. Actually, your remark is correct but it only is valid for version 2.8. 

Version 2.8 uses `-l` for the port parameter. The README.md that comes with version 2.8 in the _2.8 stable_ branch therefore shows `-l` for the port.

Version 2.9 ( = current _dev_ branch) uses `-p` for port and so will all future versions. The _dev_ branch (and its README) is the one shown by default.

We have changed the port parameter from `-l` to `-p` during 2.9 development to better match supernode and edge parameters. I am sorry for the confusion it might cause.

maomiaomi:
sorry，I forgot to check the commit history

PR #673, opened 2021-03-24
added user-password edge authentication

### LINKED ISSUE #670: Advanced Edge Authentication (Username & Password)
So far, edges authenticate by knowing the pre-shared key and the right community name. They prove their knowledge by correctly applying it to payload encryption and / or header encryption respectively.

But what happens if an edge gets stolen, e.g. in case of mobile devices? To get the thief out of the network the key on every single edge would need to be changed.

We were thinking of some more advanced edge authentication scheme. I do not want to go fully "public key" yet as it brings up a lot of PKI related questions which I do not want to load on n2n for now.

So, we internally discussed a scheme which I will put into the following post – very lengthy! **TLDR**: You can remove an edge from the network by deleting one line from a text file residing at the supernode instead of manually re-keying all your other edges. It is based on user-password / public key authentication.

I would be most interested in your opinion, especially if we shall integrate it into upcoming 3.0.

Logan007:
This pull request shall add experimental user-password edge authentication to n2n. It will be developed in increments and is considered work-in-progress and therefore marked as draft. New increments might come along with new comments explaining the changes. 

If you want to try, current code base should be available from [github's auto archive](https://github.com/Logan007/n2n/archive/edgeAuth.zip).

Comments and reviews are very welcome!

The first commit adds `curve25519()` elliptic curve cryptography to the code and, to get an impression of its speed, to the benchmark utility `tools/n2n-benchmark`. Intel Celeron 3865U delivers around 400 operations per second, i7 5775C makes around 900 operations per second.

Fixes #670.


tlsalex:
I want to give it a try by compiling the latest code which is based on the commit of 05c4249 ,
But I can't find a -J parameter in ./edge --help or ./supernode --help,
And I don't know how to use it, (PS: #670 is a long post on how it work , from a user perspective, like me , it's a little hard to understand how it is implemented).Could you help?

1. How to configure authentication at edge, just need to add -I and -J in the edge.conf ?
```
-c=xxnetwork
-k=XXXXX
...
-p=30033
-S2
-I alextls ----just like a username that choose by me
-J alextlspassword  -- just like a password that choose by me
```
2. How to configure authentication at supernode? Let's say my community is "xxnetwork", from the explanation in #670, the code in community.list should like bellow, right ?

```
-- A sane regular expression cannot start neither with an asterisk * nor with a closing bracket ] 
-- I don't understand, why you still use a * in the explanation , makes me confuse.
<community name>
* <identifier 1> <j auth value 1>
* <identifier 2> <j auth value 2>
* <identifier 3> <j auth value 3>

```
```
xxnetwork
@ alextls alextlspassword
@ user1 user1password
@ userN userNpassword
```

There is a space character to separate these three items,
which is "special character"+"space character"+"user name"+"space character"+"user password" , right?

tlsalex:
I also see something about timestamp in authentication in #670 , what will happen if the timestamp is not the same between supernode and edge ?

Logan007:
> I want to give it a try by compiling

@tlsalex, thank you for your enthusiasm! :+1:

> I can't find a -J parameter

not yet. as mentioned above, the implementations happens in _increments_ or _step by step_. First step only has the curve algorithm and the the benchmark. So, the only thing you can do with it for now is to check algorithm speed using `tools/n2n-benchmark`. Be prepared, it is slow.

This is still the beginning. Further steps will follow soon. I will be explaining each new commit as this progresses.

> How to configure authentication at edge, just need to add -I and -J in the edge.conf ?

Yes, that is exactly the plan. `-I` for _username_ and `-J` for _password_ at the edge.

> How to configure authentication at supernode?

You  got the right idea. The community file would look like
```
xxnetwork
* tlsalex =hkjdh32jhjhlj2khfdkjehjhkjhfkjhewkjfh=
* logan =43zfhu4huzbp2332pup23d923dup32d=
```

> There is a space character to separate these three items.

Yes.

> I don't understand, why you still use a * in the explanation , makes me confuse.

Because I think of special character as `*`. When reading the file, the supernode needs to decide if the line is a new _community name_ **or** a _user / password_ pair which still belongs to the previous community. So, we use a special character. We can use `*` because no real _regular expression_ can **start** with `*`.

> which is "special character"+"space character"+"user name"+"space character"+"user password" , right?

Basically yes. The special character will probably be `*` as mentioned above. And we will not write the password itself. It will be special string generated from the password. It will be created by a password generator, a simple command line for example `tools/n2n-keygen <username> <password>` which will create the public key from the password provided.

> from a user perspective, like me , it's a little hard to understand how it is implemented

You are right. It aims at explaining technical background. There will be more user-friendly documentation.

> timestamp in authentication ... if the timestamp is not the same between supernode and edge

The timestamps will be the same as used for header encryption. Header encryption will be mandatory to be enabled for this scheme to work because this scheme builds up on header encryption – just using a dynamic key instead of static key.

So, conditions are the same as for header encryption: +/- 16 seconds are allowed, otherwise, they will not be able to communicate.

Logan007:
9f8dc4e adds a a first version of the above announced key generating tool, a simple command line tool at `tools/n2n-keygen` to generate a public key from a username-password combination. Usage:

```
n2n-keygen tool

  usage:  n2n-keygen <username> <password>

          outputs a line to insert at supernode's community file for user-and-
          password authentication, please see doc/Authentication.md or the man
          pages for details
```

Please see the following example featuring the username`sisterLogan` and her secret password `007`:
```
$ tools/n2n-keygen sisterLogan 007
* sisterLogan xIly3qOTcMBEZDMMJ37csljjFMKIPIH8R6QSduH0Pn4
```

It is intended that the output line can directly go into the supernode's community list file. Therefore, it has the leading special character `*`, the username as well as the calculated public key, all neatly separated by white spaces.

Logan007:
3e884e2 makes the supernode read user data from community list file (`-c <file>`). User data can be added under some community name, e.g.

```
myCommunity
* logan nHWum<r42k1qDXdIeH>WFKeylK5UyLStRzxofRNAgpG
```

where the user line is as generated by the `tools/n2n-keygen` with username `logan` and password `007`. This will only be working with fixed name communities. First verbosity level (`-v`) will make the supernode hopefully show the successfully added username and public key at its log output.

Logan007:
e6cecec lets the `tools/n2n-keygen` utility also generate the public federation key from the federation name (`-F`), such as 

```
$ tools/n2n-keygen -F myFed
-P f5p95e9wgNx2>YwUMV5hLohJxXLWNlDQsnXAa5AUvb8
```

The output is intended to be directly used at the `edge` command line or in the config file. So, `edge` will soon get the new command line parmeter `-P` to provide the public federation key. This parameter must be same on all edges connecting to supernodes of this federation. The corresponding private key is generated from the federation name which needs to remain undisclosed to the edges (only known and shared among the supernodes).

Logan007:
dc33da7 make the supernode calculate the shared secrets (elliptic curve Diffie-Hellman) at program start for all the provided user-password pairs. Doing it once upfront keeps the costly (see `tools/n2n-benchmark`, curve25519) calculation out of the REGISTER_SUPER handling and saves us from always needing to check for (not) already calculated / present secrets.

These secrets will be used to encrypt the current dynamic key for header-encryption sent through the auth token.

Logan007:
09a8bac adds the `-J <password>` option to the edge  for use in conjunction with hostname / username `-I <username>`.

Also, the public federation key as generated by the `tools/n2n-keygen -F myFedName` utility can now be passed to the edge with `-P <public federation key>`.

A slight change to the public key's ascii encoding scheme was applied. 

Internally, edge as well as supernode calculate the shared ECDH secret(s) required to encrypt the auth tokens (yielding the same value on both sides as verified through several hexdumps).

Logan007:
9d162f3 lays ground for handling REGISTER_SUPER (plus its ACK and  NAK) with different header encryption keys than all the other packets. The `static` key is calculated as before and derived from the community name. It only gets used for REGISTER_SUPER and REGISTER_SUPER_ACK / _NAK and allows edges to start making contact to the supernode. Like entering a _Lobby_ ...

The authentication token submitted inside REGISTER_SUPER_ACK will provide the `dynamic`  header encryption key (can be calculated by supernodes only) to use with all the other kind of packets. If the edge is legit, it can decipher this key from the REGISTER_SUPER_ACK and use it to access the rest of the community network. It has entered the _Lounge_.

In the regular header encryption case without user-password authentication, `static` and `dynamic` keys are identical.

Logan007:
949f3e0 adds roughly drafted token handling as well as `dynamic` IV encryption key. The token handling needs more refinement, especially token size will require 16 more bytes [public key: 32 bytes; encrypted (using shared secret) challenge: 16 bytes]. Part 2 to follow.

Logan007:
a3ee9a8 adds the missing pieces to challenge/token/dynamic key-handling such that user/password-authentication should be working fine with one supernode now.

**Please test and let me know the results and any ideas you might have!**

By the way, user/password authentication allows the same user being "logged in" from several edges. Let me know if you think we should change it to one edge per user only.

From here, federation support is the next big step, probably cut in smaller intermediate steps again.

Logan007:
I run some final tests on this and plan to merge soon as it becomes more and more difficult to keep it on par with _dev_.

The multi-supernode support for user-password authentication would follow later with a different pull request.

Any comments?

PR #677, opened 2021-03-28
Update packaging for debian

SvenRoederer:
This improves the packaging for Debian.

* remove auto-generated files
* add the community.list file as an example
* update deprecated deb-helper "dh_clean -k"

tested on Debian 10 and 9


lucaderi:
Thank you

PR #679, opened 2021-04-02
CI: build on Linux, Windows and MacOS via GitHubAction

SvenRoederer:
This adds a simple GitHubWorkflow to build via CMake for the following platform:

* Linux (x86 and arm)
* Windows (GitHub provides VisualStudio compiler)
* MacOS

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -1,0 +1,46 @@
+name: CMake
+
+on: [push]
+
+env:
+  # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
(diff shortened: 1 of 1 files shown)


SvenRoederer:
an example can be seen here: https://github.com/SvenRoederer/n2n-vpn/actions/runs/712783936

lucaderi:
@Logan007 I have no objection to this interesting PR. For the sake of simplicity if we believe cmake is the way to go, I would move to that one and remove Makefile to avoid code duplication

Logan007:
@lucaderi As for `cmake`, I certainly will be able to get used to it. :wink:

PR #680, opened 2021-04-03
drop non portable SOL_TCP / fix MacOS v2

SvenRoederer:
Building for MacOS was broken for some time and fixed in 53e930db105df0e1103d8b6de9d5cb223c744d49. In parallel I worked on a similar fix, triggered by failing CI-builds.
This new approach seems more clean and portable than the previous as it drops using "SOL_TCP" completely. This way we can stop redefining it for specific platforms.

2nd issue is making use on TCP_CORK only when building on Linux, as it seems a Linux-specific and non-portable feature (https://linux.die.net/man/7/tcp). Defining it "use it when on Linux" seems more easy to read than "don't use it when not on Linux"
> TCP_CORK (since Linux 2.2)
    If set, don't send out partial frames. All queued partial frames are sent when the option is cleared again. This is useful for prepending headers before calling sendfile(2), or for throughput optimization. As currently implemented, there is a 200 millisecond ceiling on the time for which output is corked by TCP_CORK. If this ceiling is reached, then queued data is automatically transmitted. This option can be combined with TCP_NODELAY only since Linux 2.5.71. This option should not be used in code intended to be portable.

### DIFF
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -155,16 +155,10 @@
 #include "sn_selection.h"
 #include "network_traffic_filter.h"
 
+/* ************************************** */
+
 #include "header_encryption.h"
 #include "tf.h"
-
-/* ************************************** */
-
-#if !defined(SOL_TCP) && defined(IPPROTO_TCP)
-#define SOL_TCP IPPROTO_TCP
-#endif
-
-/* ************************************** */
 
 #ifndef TRACE_ERROR
 #define TRACE_ERROR       0, __FILE__, __LINE__
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -137,7 +137,6 @@
 #ifdef WIN32
 #include <winsock2.h>           /* for tcp */
 #define SHUT_RDWR   SD_BOTH     /* for tcp */
-#define SOL_TCP     IPPROTO_TCP /* for tcp */
 #include "win32/wintap.h"
 #include <sys/stat.h>
 #else
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -888,10 +888,10 @@ static ssize_t sendto_sock (n2n_edge_t *eee, const void * buf,
     // if the connection is tcp, i.e. not the regular sock...
     if(eee->conf.connect_tcp) {
 
-        setsockopt(eee->sock, SOL_TCP, TCP_NODELAY, &value, sizeof(value));
+        setsockopt(eee->sock, IPPROTO_TCP, TCP_NODELAY, &value, sizeof(value));
         value = 1;
 #ifndef WIN32
-        setsockopt(eee->sock, SOL_TCP, TCP_CORK, &value, sizeof(value));
+        setsockopt(eee->sock, IPPROTO_TCP, TCP_CORK, &value, sizeof(value));
 #endif
 
         // prepend packet length...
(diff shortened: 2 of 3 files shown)


SvenRoederer:
@cardigliano what your oppinion?

Logan007:
@SvenRoederer This looks good to me. Thank you!

I however requested @cardigliano to review as he brought forward the Mac OS fix.

@SvenRoederer Do you have an opportunity to check if `-S2` (TCP) actually works and lets the edge communicate well on MacOS? It appears not to work on Windows and therefore is disabled for now (`include/n2n.h:48`). I wonder if we also need to disable it on `__APPLE__`.

tlsalex:
I hope -S2 on MacOS will give some

tlsalex:
I hope -S2 on MacOS will give some inspiration for Windows

Logan007:
@tlsalex Yeah... I really do not know why it is not working on Windows. You can easily try by commenting `include/n2n.h:48` before compilation, that is *where* I turned it off. Maybe it is just my VM, but I just did not get it running (as opposed to UDP) and that is *why* I turned it off.

Hints are appreciated.

SvenRoederer:
> @SvenRoederer Do you have an opportunity to check if `-S2` (TCP) actually works and lets the edge communicate well on MacOS? It appears not to work on Windows and therefore is disabled for now (`include/n2n.h:48`). I wonder if we also need to disable it on `__APPLE__`.

Sorry, but I don't have access to any MacOS. I just found and fixed this based on GitHubActions build (https://github.com/ntop/n2n/pull/679).
But probably it'S worth thinking of adding a test that is building up a small VPN between the GHA-runners

Logan007:
> Sorry, but I don't have access to any MacOS. I just found and fixed this based on GitHubActions build (#679).

Thank you!!

> But probably it'S worth thinking of adding a test that is building up a small VPN between the GHA-runners

Wow, that would actually hit the ball out of the park! Can you do that?

Logan007:
And one more question as you seem very comfortable with Git Hub Actions: Could the binaries generated somehow be exported just like a "nightly build"?

However, I am not sure about github's policy on distribution of binaries.

lucaderi:
> And one more question as you seem very comfortable with Git Hub Actions: Could the binaries generated somehow be exported just like a "nightly build"?
> 
> However, I am not sure about github's policy on distribution of binaries.

we already make nightly build available a https://packages.ntop.org

lucaderi:
> Building for MacOS was broken for some time and fixed in [53e930d](https://github.com/ntop/n2n/commit/53e930db105df0e1103d8b6de9d5cb223c744d49). In parallel I worked on a similar fix, triggered by failing CI-builds.
> This new approach seems more clean and portable than the previous as it drops using "SOL_TCP" completely. This way we can stop redefining it for specific platforms.
> 
> 2nd issue is making use on TCP_CORK only when building on Linux, as it seems a Linux-specific and non-portable feature (https://linux.die.net/man/7/tcp). Defining it "use it when on Linux" seems more easy to read than "don't use it when not on Linux"
> 
> > TCP_CORK (since Linux 2.2)
> > If set, don't send out partial frames. All queued partial frames are sent when the option is cleared again. This is useful for prepending headers before calling sendfile(2), or for throughput optimization. As currently implemented, there is a 200 millisecond ceiling on the time for which output is corked by TCP_CORK. If this ceiling is reached, then queued data is automatically transmitted. This option can be combined with TCP_NODELAY only since Linux 2.5.71. This option should not be used in code intended to be portable.

I am building n2n on MacOS without any problem, however your patch is a good contribution I believe as it addresses some issues you reported.

PR #682, opened 2021-04-07
fix potential memory leak

randomize00:
Hi guys, I think I may found a memory leak when I read the code of edge_utils.c.


Logan007:
Good catch! Thank you!

PR #684, opened 2021-04-14
add callback function for hin2n to protect socket

randomize00:
Hello guys.
I'm trying to develop hin2n (an android APP which based on n2n source code plays the role as **edge** ) to adapt to the latest n2n source code.
But I met a problem - when edge create a socket to supernode, the socket must be protected at once, or the APP will crash.
I've tried this code and it works with hin2n well.
I will also submit a pull request to hin2n when I complete the adaptation.


Logan007:
Thank you!

Issue #710, opened 2021-06-11
Cannot compile the latest branch on Openwrt X86 toolchain!


galaxyskyknight:
when configure, it prompts the gcc cannot create excutable error, please see the log below, the stable release (v2.8) can compile well,  I don't know what happened , may related the 2.9.0 code change some parameter usage for gcc? To compile this code on different openwrt platform is very important to me, please kindly help. thanks @ntop @Logan007 

This file contains any messages produced by compilers while
running configure, to aid debugging if configure makes a mistake.

It was created by edge configure 2.9.0, which was
generated by GNU Autoconf 2.69.  Invocation command line was

  $ ./configure --target=x86_64-openwrt-linux --host=x86_64-openwrt-linux --build=x86_64-pc-linux-gnu --program-prefix= --program-suffix= --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --libexecdir=/usr/lib --sysconfdir=/etc --datadir=/usr/share --localstatedir=/var --mandir=/usr/man --infodir=/usr/info --disable-nls

## --------- ##

hostname = 
uname -m = x86_64
uname -r = 4.15.0-142-generic
uname -s = Linux
uname -v = #146-Ubuntu SMP Tue Apr 13 01:11:19 UTC 2021

/usr/bin/uname -p = unknown
/bin/uname -X     = unknown

/bin/arch              = unknown
/usr/bin/arch -k       = unknown
/usr/convex/getsysinfo = unknown
/usr/bin/hostinfo      = unknown
/bin/machine           = unknown
/usr/bin/oslevel       = unknown
/bin/universe          = unknown

PATH: /root/OpenWRT/openwrt-sdk-21.02.0-rc2-x86-64_gcc-8.4.0_musl.Linux-x86_64/staging_dir/target-x86_64_musl/host/bin
PATH: /root/OpenWRT/openwrt-sdk-21.02.0-rc2-x86-64_gcc-8.4.0_musl.Linux-x86_64/staging_dir/hostpkg/bin
PATH: /root/OpenWRT/openwrt-sdk-21.02.0-rc2-x86-64_gcc-8.4.0_musl.Linux-x86_64/staging_dir/toolchain-x86_64_gcc-8.4.0_musl/bin
[...]

PR #712, opened 2021-06-14
Update openwrt makefile and add patch

small-5:
Fix https://github.com/ntop/n2n/issues/710
Remove unnecessary dependes


Logan007:
Thank you!!

PR #716, opened 2021-06-18
added manual MAC/IP address spoofing protection override at supernode (-M)

### LINKED ISSUE #714: option to supernode to turn off the function of mac authentication
In addition,could you please add an option to supernode to turn off the function of mac authentication?Because in some cases, it is not necessary to start MAC anti spoofing.Thank you!

from  DDNS Bug #703   (This issue also discusses this point)

In fact, in embedded devices, it operates unattended.Most of the time, often power off to restart, or network instability and other reasons disconnected.
It would be a lot easier if you added an option.

Logan007:
This pull requests adds `-M` cli option to the supernode. It overrides the MAC/IP address spoofing protection. Although it could be useful in scenarios comprising often unexpectedly happening edge reboots, user/password-based authentication is recommended instead. An corresponding WARNING is eventually output.

Fixes #714.


small-5:
Test OK,Thank you!

Issue #698, opened 2021-05-28
communication via spuernode<->supernode fail


fghlw:
edge A connected to node A; edge B connected to node B, node B set -l xxxx:xxxx connect to node A; 
node A with internet IP, node B be in internal network but can access internet communicate to node A.
But edge A can't communicate to edge B

Logan007:
Do you want to check out current _dev_? I have a strong feeling that #707 might also have fixed this issue.

fghlw:
thank you. I will try it.

fghlw:
a question, if need rebuild "edge"? I remember I replaced some function(about get/set IP?) to make it work under "xp", but I forgot which one...

Logan007:
I am not all too deep into Windows, but you are maybe talking about `inet_ntop()`?

fghlw:
Yes, looks like this function not available in xp?

Logan007:
I can't say for sure, never tried XP. But I saw @etherealvisage having implemented a [corresponding wrapper function](https://github.com/elphasecure/n2n/commit/7cc7624a48798550ad6f4b5f26df4b3518125327).

If this turns out to be the missing piece to get the software running on Windows XP, hope is that someone back-ports it to our _dev_ in a way which does not affect compatibility with more recent Windows versions.

etherealvisage:
The backport should be pretty simple. Add the relevant source files and conditionally compile them in if on XP/if asked to during configuration. Not sure if that's all that's needed for XP support, but certainly this piece isn't hard :-)

Logan007:
@etherealvisage Do you want to provide a pull request?

fghlw:
the newest "autogen.sh" seems miss "./configure"? it can't work

Logan007:
As far as I remember, `./configure` always needs to be manually invoked after `./autogen.sh`. I have never seen anything else in this project.

fghlw:
> As far as I remember, `./configure` always needs to be manually invoked after `./autogen.sh`. I have never seen anything else in this project.

the code I downloaded on March include this line, so I thought it included

Logan007:
I am not sure what has happened there. Especially as we always [advertise](https://github.com/ntop/n2n#manual-compilation) to use the `./autogen.sh`-`./configure`-`make`trinity.

Apart from this, does it work for you with a manual `./configure`?

fghlw:
it works, but if included in "./autogen.sh" more "auto"..   :)

[...]

PR #717, opened 2021-06-20
modified route setup

### LINKED ISSUE #528: Windows -n routing option, wrong gateway
Hi, I have a problem with the **-n** option in Windows, this is the situation:
- I'm IP **192.168.1.34** with gateway **192.168.1.1**
- I'm connecting with ip **10.254.254.11**, to an edge with ip **10.254.254.10** and two subnet: **10.10.10.0** and **10.123.123.0**
- I'm using options **-n10.10.10.0/24:10.254.254.10** and **-n10.123.123.0/24:10.254.254.10**
The fact is that, I can ping 10.254.254.10 successfully, but I can't ping anything in the subnet 10.10.10.0 or 10.123.123.0

To solve I have just to run this command:
`route change 10.10.10.0 mask 255.255.255.0 10.254.254.10`, I can successfully go in subnet 10.10.10.0

So, why under the interface column, when I use the -n option, there isn't the n2n network adapter but the pc one?

**Before:** (not working)
ip2
**After:** (working 10.10.10.0)
ip3

### LINKED ISSUE #709: how to configure a specified IF for -n <cidr:gateway) parameter via my perfered Iface but not default gw?
I found if I configure the -n <cidr:gateway> with edge spaw up, it will automatically attached to a local IF which have default route(0.0.0.0) but not what I wanted IF( e.g n2n) this will cause the cidr not route to proper 'gateway' directly but route to default gw attached IF. so is there any tricky that I can change this or likely I could use '-n <cidr:gateway:IFname(or IFidx)> to specify the subnet route to my perferred network adapter?

Logan007:
This pull request modifies route setup. It is moved to after tap device setup happened (bootstrap area in `edge.c`'s `main()`) to make sure that all the required tap device properties already are available.

Also, for the sake of proper route setup, tap interface index is being caught during tap interface ouverture on Windows.

Fixes #709.
Fixes #528.

### DIFF
--- a/src/edge.c
+++ b/src/edge.c
@@ -52,6 +52,7 @@ int fetch_and_eventually_process_data (n2n_edge_t *eee, SOCKET sock,
                                        time_t now);
 int resolve_create_thread (n2n_resolve_parameter_t **param, struct peer_info *sn_list);
 int resolve_check (n2n_resolve_parameter_t *param, uint8_t resolution_request, time_t now);
+int edge_init_routes (n2n_edge_t *eee, n2n_route_t *routes, uint16_t num_routes);
 
 /* ***************************************************** */
 
@@ -1137,7 +1138,7 @@ int main (int argc, char* argv[]) {
             }
         }
 
-        if(runlevel == 4) { /* configure the TUNTAP device */
+        if(runlevel == 4) { /* configure the TUNTAP device, including routes */
             if(tuntap_open(&tuntap, eee->tuntap_priv_conf.tuntap_dev_name, eee->tuntap_priv_conf.ip_mode,
                            eee->tuntap_priv_conf.ip_addr, eee->tuntap_priv_conf.netmask,
                            eee->tuntap_priv_conf.device_mac, eee->tuntap_priv_conf.mtu
@@ -1151,6 +1152,11 @@ int main (int argc, char* argv[]) {
                                      eee->tuntap_priv_conf.ip_addr,
                                      eee->tuntap_priv_conf.netmask,
                                      macaddr_str(mac_buf, eee->device.mac_addr));
+            // routes
+            if(edge_init_routes(eee, eee->conf.routes, eee->conf.num_routes) < 0) {
+                traceEvent(TRACE_ERROR, "routes setup failed");
+                exit(1);
+            }
             runlevel = 5;
             // no more answers required
             seek_answer = 0;
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -41,7 +41,7 @@ static void check_peer_registration_needed (n2n_edge_t *eee,
                                             const n2n_sock_t *peer);
 
 static int edge_init_sockets (n2n_edge_t *eee);
-static int edge_init_routes (n2n_edge_t *eee, n2n_route_t *routes, uint16_t num_routes);
+int edge_init_routes (n2n_edge_t *eee, n2n_route_t *routes, uint16_t num_routes);
 static void edge_cleanup_routes (n2n_edge_t *eee);
 
 static void check_known_peer_sock_change (n2n_edge_t *eee,
@@ -414,11 +414,6 @@ n2n_edge_t* edge_init (const n2n_edge_conf_t *conf, int *rv) {
 #endif
     if(edge_init_sockets(eee) < 0) {
         traceEvent(TRACE_ERROR, "socket setup failed");
-        goto edge_init_error;
-    }
-
-    if(edge_init_routes(eee, eee->conf.routes, eee->conf.num_routes) < 0) {
-        traceEvent(TRACE_ERROR, "routes setup failed");
         goto edge_init_error;
     }
 
@@ -3391,6 +3386,7 @@ static int edge_init_routes_win (n2n_edge_t *eee, n2n_route_t *routes, uint16_t 
     struct in_addr net_addr, gateway;
     char c_net_addr[32];
     char c_gateway[32];
+    char c_interface[32];
     char cmd[256];
 
     for(i = 0; i < num_routes; i++) {
@@ -3404,12 +3400,12 @@ static int edge_init_routes_win (n2n_edge_t *eee, n2n_route_t *routes, uint16_t 
             memcpy(&gateway, &(route->gateway), sizeof(gateway));
             _snprintf(c_net_addr, sizeof(c_net_addr), inet_ntoa(net_addr));
             _snprintf(c_gateway, sizeof(c_gateway), inet_ntoa(gateway));
-            _snprintf(cmd, sizeof(cmd), "route add %s/%d %s > nul", c_net_addr, route->net_bitlen, c_gateway);
+            _snprintf(c_interface, sizeof(c_interface), "if %u", eee->device.if_idx);
+            _snprintf(cmd, sizeof(cmd), "route add %s/%d %s %s > nul", c_net_addr, route->net_bitlen, c_gateway, c_interface);
             traceEvent(TRACE_NORMAL, "ROUTE CMD = '%s'\n", cmd);
             system(cmd);
         }
     }
-
 #endif // WIN32
 
     return (0);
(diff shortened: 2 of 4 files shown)


small-5:
Test OK,Thank you!

Logan007:
@small-5 Thank you for testing and feedback!!

I will merge soon. I still am waiting for feedback related to compilation, especially on XP, and see if that is an issue or even still a requirement.

galaxyskyknight:
haven't find a way to pull your code since it is not commited, but I read the core part for  'route CMD', just a question: it seems implemented like this way:

`_snprintf(cmd, sizeof(cmd), "route %s %s/%d %s %s > nul", c_verb, c_net_addr, route->net_bitlen, c_gateway, c_interface);`

first of all, is the 'add' action missing in the latest change?

image

second  ,the windows console command from the code change looks like could be like the following even we assuming 'add'  fixed, here is the format for examle (assume cidr is 192.168.5.0/24 gateway is 10.0.0.1 corresponding ifidx is 5)
`route add 192.168.5.0/24 10.0.0.1 5`
it will be unrecognized by windows console , it says there is a wrong parameter hence you directly append the ifindex without specifiy the 'IF' prefix option
it should follow the following format, the MASK 255.255.255.0 part could be merged with /24 and metric  no need to be explicitely specified.
image
so the right command format would be
`route add 192.168.5.0/24 10.0.0.1 IF 5`

skyformat99:
> so the right command format would be
> `route add 192.168.5.0/24 10.0.0.1 IF 5`


The fixed code looks right.
line 3404  3405   with add and if.

galaxyskyknight:
> > so the right command format would be
> > `route add 192.168.5.0/24 10.0.0.1 IF 5`
> 
> The fixed code looks right.
> line 3404 3405 with add and if.

Oh yeah, sorry, I just skipped two lines above , my fault:)  so looks good. thanks.

Logan007:
> haven't find a way to pull your code since it is not commited

You could use github's [auto-archive](https://github.com/Logan007/n2n/archive/winRt.zip).

> first of all, is the 'add' action missing in the latest change?

I found that the VPN driver (still running after edge stops) on Windows keeps up the routes even if edge is restarted. So, I tried to find an easy way to delete the routes at edge's end again. That's why we can have `add` or `delete` action / verb. This is special to Windows. On Linux, all device-related routes seem to be gone as soon as the device is closed / destroyed.

> it should follow the following format, the MASK 255.255.255.0 part 

The actually assembled and issued route command is output to the log. Please check out the code linked above and let me know if the format still needs a change. 

> I still am waiting for feedback related to compilation, especially on XP,

I am not sure if this compiles on XP and / or if we might need to find another, i.e. XP-compatible, way to retrieve the interface index. From #698, I got that @fghlw compiles on Windows XP. Could you confirm?

galaxyskyknight:
on windows 10, when the edge teardown, the TAP IF related routes will be automatically deleted as the TAP interface itself down. that's what I observed. I used the win OpenVpn TUN/TAP drivers. but any way if there is a double check with add/delete action could be safer. For XP, is there anyway to use a macro when compile to distiguish the platform?  if it is XP, just keep the old way.

galaxyskyknight:
have tested the change on win10 build,  work perfect. thanks.

galaxyskyknight:
I tried the winRt branch code change build on win 10,  the basic function works fine, but looks one little bug like this:
for normal edge startup and exit, the logic will cover route add and delete branch, and setup/flush the routes, however, if there is a mac spoofing excetion (means the edge re-spawn shortly before its keepalive aged out on supernode)which caused the edge process exit, seems in this circumstance it will not go into route delete code routine, please check if this is the case.

Logan007:
> double check with add/delete action could be safer

It definitely is. Before I added the `delete` verb, I had witnessed that a route (-n ...) from a previous run disappeared when the edge quit but showed up again when I re-started the the edge (without any `-n` then).

> the edge process exit

I agree, this is worth to be fixed. It is just not route-specific because there are several places in the code which `exit()` the edge without further clean-up. We would need a kind of more general `clean_up_and_exit_edge()` to be called. Probably something to be coded in a separate pull request.

Do you want to open a corresponding issue?

fghlw:
> > haven't find a way to pull your code since it is not commited
> 
> You could use github's [auto-archive](https://github.com/Logan007/n2n/archive/winRt.zip).
> 
> > first of all, is the 'add' action missing in the latest change?
> 
> I found that the VPN driver (still running after edge stops) on Windows keeps up the routes even if edge is restarted. So, I tried to find an easy way to delete the routes at edge's end again. That's why we can have `add` or `delete` action / verb. This is special to Windows. On Linux, all device-related routes seem to be gone as soon as the device is closed / destroyed.
> 
> > it should follow the following format, the MASK 255.255.255.0 part
> 
> The actually assembled and issued route command is output to the log. Please check out the code linked above and let me know if the format still needs a change.
> 
> > I still am waiting for feedback related to compilation, especially on XP,
> 
> I am not sure if this compiles on XP and / or if we might need to find another, i.e. XP-compatible, way to retrieve the interface index. From #698, I got that @fghlw compiles on Windows XP. Could you confirm?

I compiled it via vs2019 on win10.  vs2019 with "VS 2017 (v141) C++ Windows XP support"

Logan007:
> I compiled it via vs2019 on win10. vs2019 with "VS 2017 (v141) C++ Windows XP support"

I might add this hint to the Building.md document, thank you!

Does this branch (see link to auto-archive above) compile well this way? Does the result (edge with some `-n ...` option) run on Windows XP?

galaxyskyknight:
> > double check with add/delete action could be safer
> 
> It definitely is. Before I added ther `delete` verb, I had witnessed that a route (-n ...) from a previous run disappeared when the edge quit but showed up again when I re-started the the edge (without any `-n` then).
> 
> > the edge process exit
> 
> I agree, this is worthy to be fixed. It is just not route-specific because there are several places in the code which `exit()` the edge without further clean-up. We would need a kind of more general `clean_up_and_exit_edge()` to be called. Probably something to be coded in a separate pull request.
> 
> Do you want to open a corresponding issue?

Yes, issue #718 opened, thanks

galaxyskyknight:
> > I compiled it via vs2019 on win10. vs2019 with "VS 2017 (v141) C++ Windows XP support"
> 
> I might add this hint to the Building.md document, thank you!
> 
> Does this branch (see link to auto-archive above) compile well this way? Does the result (edge with some `-n ...` option) run on Windows XP?

FYI: I run vs2019 build your code above and run on win10, I configured multiple routes with -n in edge.conf, work well as expected(besides the bug mentioned).  no enviroment for XP for testing.

fghlw:
> > I compiled it via vs2019 on win10. vs2019 with "VS 2017 (v141) C++ Windows XP support"
> 
> I might add this hint to the Building.md document, thank you!
> 
> Does this branch (see link to auto-archive above) compile well this way? Does the result (edge with some `-n ...` option) run on Windows XP?

I can test it, but last version code can't compile, by some code in "n2n.c" about thread(yesterday did it, remember line:397?).

Logan007:
The thread things are resolver related, I think. They compile well on Windows 10. I am sorry to hear that they might not compile on XP. With a view to the other network rrelated XP compatibility issue, we soon might need to make some decision on how to support XP builds.

I think it would be too early to retire XP support. What do you think?

On the other hand, I am a bit clueless on how to achieve compatibility -- there is a solution for inet_ntop, but none for the interface index yet, neither for the recently discovered thread issue (might be handeled the old way, undeffing the N2N_HAVE_PTHREAD macro for XP).

My availability will be limited during the upcoming weeks. Please feel free to discuss and test different solutions for we can implement one of them some time soon.

galaxyskyknight:
I am a little bit curious why we cannot retire xp support? :)  Since Microsoft itself have retired XP  7 years ago  and they worried about nothing:)  what we worried about for such a networking software retiring its support? and there is how many people are still using win XP indeed?  if they really need this software, they must upgrade to upper OS version as there is no security patch for XP and we actively introduce the support of N2N for XP is equal to expose their OS in more risky enviroment for this software's inter-connection feature and I believe it is not a wise option, we should advise them must make a choice either upgrade OS for better support if they really need this software in security consideration or they could keep stay there but no more support.

skyformat99:
> I am a little bit curious why we cannot retire xp support? :) Since Microsoft itself have retired XP 7 years ago and they worried about nothing:) what we worried about for such a networking software retiring its support? and there is how many people are still using win XP indeed? if they really need this software, they must upgrade to upper OS version as there is no security patch for XP and we actively introduce the support of N2N for XP is equal to expose their OS in more risky enviroment for this software's inter-connection feature and I believe it is not a wise option, we should advise them must make a choice either upgrade OS for better support if they really need this software in security consideration or they could keep stay there but no more support.


If XP can be supported, that would be great.
If it is not supported, it is understandable.

Logan007:
So we shall skip XP support for now.

Let's revisit some time soon to maybe find compatible implementations for the respective parts (`inet_ntop()`, interface index, `-x` cli option) or just def them out.

Issue #719, opened 2021-06-27
segfault


small-5:
Use the latest source code to cross compile, segfault appears on OpenWRT
Some core files are generated in the /tmp/
Please Fix it,Thanks!
[log.txt](https://github.com/ntop/n2n/files/6721486/log.txt)

small-5:
This problem arises when the supernode cannot connect.
This segfault will not appear if i don't use boot auto start and wait for the network to connect and start manually.

skyformat99:
> This problem arises when the supernode cannot connect.
> This segfault will not appear if i don't use boot auto start and wait for the network to connect and start manually.

![image](https://user-images.githubusercontent.com/10938976/123588361-37b09200-d81a-11eb-953e-f4a60afdbac8.png)

The problem is this function.
Actually, this shouldn't be a bug.

skyformat99:
Is it possible to consider adding an automatic reconnect option, when there is a problem, automatic cycle reconnect.

small-5:
I feel that this problem appeared after 1586ae62a77542becd6eb846060e9de799d4e272, but there was no before. I will use the version before 1586ae62a77542becd6eb846060e9de799d4e272 to compile and test.

small-5:
Yes, the problem is indeed caused by 3c6ab56d3fdbc6464803cca61fe814f8b8fbbd6f, I used c55e3e32463d9821b7459c9c86108e6f364a1529 to compile it is normal.

small-5:
The problem is indeed caused by 3c6ab56, I used c55e3e3 to compile it is normal.

PR #720, opened 2021-06-28
Fixed segfault

small-5:
Fix https://github.com/ntop/n2n/issues/719

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -3504,6 +3504,10 @@ int edge_conf_add_supernode (n2n_edge_conf_t *conf, const char *ip_and_port) {
     sock = (n2n_sock_t*)calloc(1,sizeof(n2n_sock_t));
     rv = supernode2sock(sock, ip_and_port);
 
+    while(rv == -2) {
+        rv = supernode2sock(sock, ip_and_port);
+    }
+
     if(rv < -2) { /* we accept resolver failure as it might resolve later */
         traceEvent(TRACE_WARNING, "Invalid supernode parameter.");
         free(sock);
--- a/src/n2n.c
+++ b/src/n2n.c
@@ -292,6 +292,7 @@ int supernode2sock (n2n_sock_t *sn, const n2n_sn_name_t addrIn) {
                     traceEvent(TRACE_WARNING, "supernode2sock fails to resolve supernode IPv4 address for %s", supernode_host);
                     rv = -1;
                 }
+                freeaddrinfo(ainfo); /* free everything allocated by getaddrinfo(). */
             } else {
                 traceEvent(TRACE_WARNING, "supernode2sock fails to resolve supernode host %s, %d: %s", supernode_host, nameerr, gai_strerror(nameerr));
                 rv = -2;
@@ -306,7 +307,6 @@ int supernode2sock (n2n_sock_t *sn, const n2n_sn_name_t addrIn) {
         rv = -4;
     }
 
-    freeaddrinfo(ainfo); /* free everything allocated by getaddrinfo(). */
     ainfo = NULL;
 
     return rv;


skyformat99:
> Fix #719

good

src/edge_utils.c
```diff
@@ -3504,6 +3504,10 @@ int edge_conf_add_supernode (n2n_edge_conf_t *conf, const char *ip_and_port) {
     sock = (n2n_sock_t*)calloc(1,sizeof(n2n_sock_t));
     rv = supernode2sock(sock, ip_and_port);
 
+    while(rv == -2) {
```
Logan007, line 3507: This piece of code
- definitely needs some sleep, e.g. `sleep(2);` for not maxing out the CPU
- should go to another place (please see conversation comment)

src/n2n.c
```diff
@@ -292,6 +292,7 @@ int supernode2sock (n2n_sock_t *sn, const n2n_sn_name_t addrIn) {
                     traceEvent(TRACE_WARNING, "supernode2sock fails to resolve supernode IPv4 address for %s", supernode_host);
                     rv = -1;
                 }
+                freeaddrinfo(ainfo); /* free everything allocated by getaddrinfo(). */
```
Logan007, line 295: Does this alone fix the segfault? Or does it only work when all supernodes are resolved?

Logan007 (commented):
Thank you for your contribution!! :orange_heart: 

I am not too deep into the route setup code yet but I fully agree that errors should be reported, not by a segfault of course.

I think it generally is a good idea to have n2n output an error message by default if something fails, e.g. name resolution or route setup. Those errors may not be that severe to also require the edge to stop. Furthermore, I do not think that fully completed name resolution should be a common requirement to start-up the edge (some names might resolve way later, e.g. strategic reserve supernodes).

If your intent is to make the edge wait until **all supernode names resolve well**, we should do it somewhere around the edge's bootstrap using a new command line option (as this is optional behavior) to trigger the wait loop.

If however your intention is to make the edge wait until **the network is up**, I am not sure if this is the way to go. It might only work if names need to be resolved and might not work if plain IP addresses are given.

src/edge_utils.c
small-5: I originally wanted to join sleep 2 or even 5, but the test found that it was actually 5 seconds before the next parsing.
The purpose of adding this section is to loop the resolution when the domain name is set to be non IP and the resolution fails. If the program continues to run when the RV value is - 2, an error will be reported (sorry, I forgot the error prompt). After adding this section, this error can be avoided.

small-5: I originally wanted to add sleep 2 or even 5, but the test found that there was already a 5 second interval for the next domain name resolution.
The purpose of adding this section is to loop the resolution when the domain name is set to be non IP and the resolution fails. If the program continues to run when the RV value is - 2, an error will be reported (sorry, I forgot the error prompt). After adding this section, this error can be avoided.

src/n2n.c
small-5: Yes, this paragraph solves the segfault bug, and the previous paragraph solves a error output.

src/edge_utils.c
Logan007: The aim is to have the edge start-up even without the fully resolved names. It has already been working in my test scenarios. So, we need to find out more about this error, its triggers and maybe a solution that does not require to resolve all names during start-up.

Could we handle this error message in a separate issue / pull request, please?

Logan007 (changes requested):
Please see my [comment](https://github.com/ntop/n2n/pull/720/files/c8ab0df834cd572b0d481fcb8e017f46387a825b#r667369771).

PR #724, opened 2021-07-13
fixed error message

small-5:
There is no need to add a sleep, because it already has a sleep of 5 seconds

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -3507,6 +3507,10 @@ int edge_conf_add_supernode (n2n_edge_conf_t *conf, const char *ip_and_port) {
     sock = (n2n_sock_t*)calloc(1,sizeof(n2n_sock_t));
     rv = supernode2sock(sock, ip_and_port);
 
+    while(rv == -2) {
+        rv = supernode2sock(sock, ip_and_port);
+    }
+
     if(rv < -2) { /* we accept resolver failure as it might resolve later */
         traceEvent(TRACE_WARNING, "Invalid supernode parameter.");
         free(sock);


Logan007:
Now I see what error messages you were referring to.

Actually, the WARNING is intentional and to inform the user that a supernode cannot be resolved. That might not only happen during startup (DNS not ready yet, wrong DNS setting, DNS fail, ...) but also anytime during edge's lifetime (DNS fail, ...). I think this is a condition the user needs to be informed about. Don't you think that WARNING is the right category?

The ERROR message is from trying to send data using an AF_INVALID socket, i.e. to a non-resolved supernode. This happens at a well defined place (`edge_utils.c:932`) where we should catch that error and reduce log output level to DEBUG (only shown if some `-v` given) in case of this error:

```C

/* edge_utils.c:916 */
    int trace_level = TRACE_ERROR;

    ...

/* edge_utils.c:935 */
        if((sent <= 0) && (errno)) {
            // downgrade to TRACE_DEBUG in case of custom AF_INVALID, i.e. supernode not resolved yet
            if(errno == EAFNOSUPPORT /* 93 */)
                trace_level = TRACE_DEBUG;
            char * c = strerror(errno);    
            traceEvent(trace_level, "sendto_fd sendto failed (%d) %s", errno, c);
#ifdef WIN32
            traceEvent(trace_level, "sendto_fd WSAGetLastError(): %u", WSAGetLastError());
#endif

   ...
```

This way, we fix the message at its source and do not interfere with the intended behavior that the edge starts up even if not all supernodes can be resolved yet.

Unconditionally waiting for all supernodes to be resolved could be added as an extra command line option instead – if still required then.

Do you want to apply to required changes to the pull request?

small-5:
So please create a PR, I will close this PR, because I am afraid that I will not be able to deal with it.Thanks!

Logan007:
Please see #725 and let me know.

PR #725, opened 2021-07-13
conditionally downgraded one ERROR msg to DEBUG msg

Logan007:
This addresses #724.

I opted against kick-out-checking for AF_INVALID in `dest`'s socket at the beginning of `sendto_fd()` because I do not want it to miss the following parts such as the call to `supernode_disconnect()` in case of TCP.


small-5:
I will give you feedback after the test tomorrow, thank you!

small-5:
The problem has been solved, please merge.Thanks!

Issue #562, opened 2020-12-24
TECHNICAL ROAD TO 3.0


Logan007:
- [x] perform DNS resolution in its own thread, please test #645 (assigned: `Logan`)
- [x] optimize packet forwarding in federations (assigned: `Logan & fcarli3`)
- [x] log message revision (assigned: `Logan`)
- [x] update man page (assigned: `Hamish Coleman`)
- [x] change the way help text is displayed (assigned: `Logan`)
- [x] add platform specifics to documentation (assigned: `Logan`)
- [x] review bootstrap and main loop (assigned: `Logan`)
- [x] review header encryption (assigned: `Logan`)
- [x] review hashing function (assigned: `Logan`)
- [x] fix configuration file parsing #561 (assigned: `Logan`)
- [x] propagate local socket information (assigned: `Logan`)

Optional:
- [x] TCP support #627 (assigned: `Logan`)
- [x] user-password authentication scheme for edges #670 (assigned: `Logan`)
- [ ] multi-link support (assigned: `-`)
- [ ] upnp support (assigned: `anzz1`)
- [x] TAP interface metric option (Windows) (assigned: `anzz1`)
- [x] find MinGW build incompatibilities #617 #631 #642 (assigned: `Hamish Coleman`, `Logan`)
- [x] supernode management port command to reload communities from file (assigned: `Logan`)
- [ ] prepare inter-supernode communication (assigned: `-`)
- [ ] support for `--pre-up`, `--post-up`, `--pre-down`, `--post-down` CLI options to run scripts #694 #743 (assigned: `-`)
- [x] optionally bind edge to a specified IP address (assigned: `Logan`)

This is about what technical features to include / change for an upcoming **n2n 3.0** release. This post will regularly be updated to reflect current changes.

The list above is absolutely not carved in stone. Please share your thoughts for discussion and do not hesitate to propose any of your ideas! 

If you want to contribute to some of the listed features or to a not-yet-listed feature, such as upnp support, or… let us know!

fengdaolong:
For users, a relatively stable mainline version is needed, because progressive versions change all the time, users will not be able to accept such incompatible changes.

As far as I know, the "master" branch seems to be unmanaged for a long time. We might as well use "master" to replace the current "dev" branch. The current "dev" branch becomes a bold and innovative development branch . In the entire code repository, set "master" as the default.

the "master" branches carry out conservative iterations, and the "dev" branches carry out bold functional innovations.

[...]

PR #729, opened 2021-07-18
Update man pages with current options

hamishcoleman:
A simple update of the man pages from the documentation embedded within the source code.

Attempting to start addressing the "update man page" task from the #562 ticket


Logan007:
> Attempting to start addressing the "update man page" task from the #562 ticket

Well noted (at #562), thank you for taking initiative!! :+1:

PR #730, opened 2021-07-20
2.8 stable

Nidhogg-Wyrmborn:
Would Like a Pull request to be able to use it in its stable form


Logan007:
I am not aware of any commit to _2.8_ branch which would not be part of the _dev_ branch as well; they all should be made to _dev_, too; see #390 and #391 for example.

Also, I am confused as the pull request shows `2.7.0` in _dev_ branch's `CMakeLists.txt` file whereas it definitely was raised to `2.9.0` as you can see [here](https://github.com/ntop/n2n/blob/289276467086236da498b8af59735b6cdabb47a8/CMakeLists.txt#L9) in _dev_'s code online.

I am not quite sure if I correctly get all this. Could you please provide some more explanation to give me a better understanding of  this pull request's exact purpose? Thank you!

skyformat99:
Maybe the person is mistaken.

Logan007:
@0034120, if I am not mistaken, the [2.8 stable branch](https://github.com/ntop/n2n/tree/2.8-stable) already contains the commits you wanted to have merged.

Please re-open with more detailed explanation if required.

PR #738, opened 2021-07-23
Update wintap.c

skyformat99:
fix  compile error  on MSVC2015:

The compiler failed to compile on Windows. The error is as follows:
file n2n.c line 397
// create thread
ret = pthread_create(&((*param)->id), NULL, resolve_thread, (void *)*param);
if(ret) {
traceEvent(TRACE_WARNING, "resolve_create_thread failed to create resolver thread with error number %d", ret);
return -1;
}

build error info:
Severity	Code	Description	File	Project	Line
Error	C2440	'function': cannot convert from 'void *(__cdecl *)(void *)' to 'LPTHREAD_START_ROUTINE'
\n2n\src\n2n.c	n2n	397


Logan007 (changes requested):
We try to keep the "external" includes all gathered in `n2n.h`.

win32/wintap.c
```diff
@@ -4,7 +4,10 @@
 
 #include "n2n.h"
 #include "n2n_win32.h"
-
+#if defined(_MSC_VER)
```
Logan007: Please move the whole block to `include/n2n.h`, e.g. around line 138.

Another option weould be the `win32/n2n_win32.h` file which already contains some other windows-specific network related  includes.

skyformat99:
It has been modified according to the new instructions.
Sorry, I'm not familiar with the PR  git command

Logan007:
Thank you very much!!

I have restored the one line feed in `wintap.c` to make it look original again (no changes compared to dev).

PR #740, opened 2021-07-23
added 'reload_communities' command to supernode management port

Logan007:
This pull request adds the `reload_communities` command to the supernode's management port (as well as a `help` command).

The `reload_communities` command will disconnect from all edges, free the related memory areas and reload the `-c`-provided community file. This is useful in case it has changed on disk, e.g. by adding, changing or deleting users or communities.

In case of user-password authenticated communities, the connected edges will be sent a new `RE_REGISTER_SUPER` order to re-register as soon as possible to fetch a new dynamic key and keep traffic going. ~~Actually, the key is not re-generated yet, that will be part of the sometime soon upcoming federation support for user-password authentication – along with a slightly changed key-generation scheme.~~ _A new dynamic key is generated when issuing `reload_communities`. It will prevent edges deleted from the `community.list` file to re-register. However, the key-generation scheme for the dynamic key will change with upcoming federation support._

Regular communities, also those with header encryption enabled, will "automatically" be updated because the supernode does not answer their REGISTER_SUPER anymore (if community was deleted from list).

### DIFF
--- a/include/n2n_define.h
+++ b/include/n2n_define.h
@@ -29,6 +29,7 @@
 #define MSG_TYPE_PEER_INFO                  10
 #define MSG_TYPE_QUERY_PEER                 11
 #define MSG_TYPE_MAX_TYPE                   11
+#define MSG_TYPE_RE_REGISTER_SUPER          12
 
 /* Max available space to add supernodes' informations (sockets and MACs) in REGISTER_SUPER_ACK
  * Field sizes of REGISTER_SUPER_ACK as used in encode/decode fucntions in src/wire.c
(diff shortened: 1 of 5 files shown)


galaxyskyknight:
how to use this feature? @Logan007

Logan007:
Please connect to the supernode management port as described [here](https://github.com/ntop/n2n/blob/dev/doc/Faq.md#can-i-get-a-list-of-connected-edge-nodes-and-their-community-and-source-ip-address-from-the-supernode).

Then, just send the command `help` or `reload_communities`  followed by ENTER (instead of a plain ENTER as described in the linked document).

PR #741, opened 2021-07-24
added federation support to user/password auth scheme

### LINKED ISSUE #726: Question: Is there any plan for the federation feature to support supernode sync up while using -I -J encryption?
currently for the federation feature, if between supernodes only specify the same federation, the edge info could be sync up and all edge nodes cound be inter-communicated, however, if using -I -J for supernodes, those edges registered on different supernodes could not be inter-communicated any more but limited to registered supernode ONLY. That seems if using federation encryption mode ,the edges info will not sync up if they attache to different supernode, there are different behavior for encryption/non-encryption in federation feature, I am more preffered that the edges should be synced up for those registered in the same federation zone among supernodes.

I mentioned above before in another issue, and seems is it WIP?  any plans?:)

Logan007:
This pull request adds federation support to the user/password auth scheme.

By ~~this first~~ _the second_ commit, it should already reliably be working. If you have a chance, **please test the [code created by github's auto-archiver](https://github.com/Logan007/n2n/archive/fedUserPw.zip) and let me know whatever you find**.

:warning: Just make sure that all edges and all supernodes are built from that code, it will not work otherwise.

However, it still is cosidered work in progress as one important part is still missing and thus this is marked as _Draft_:

For every REGISTER_SUPER from an edge, the receiving supernode needs to wait a while before sending out the REGISTER_SUPER_ACK just to be sure that no other supernode (to which the REGISTER_SUPER is forwarded) objects by a REGISTER_SUPER_NAK. Coming sometime soon...

Resolves #726.

### DIFF
--- a/doc/Authentication.md
+++ b/doc/Authentication.md
@@ -55,11 +55,14 @@ ntop[0-1][0-9]
 
 This example also lists another user `sister` (with the same secret password `007`). The users become part of the preceding community name, `netleo` in this case. The public keys are cryptographically tied only to the user name, not to the community name. That way, to switch a user from one community to another, her line can easily be copied from one community section to another. By the way, do not forget to provide the `community.list` file to the supernode through the `-c community.list` parameter.
 
-Current supernode behavior does not limit the simultanious usage of usernames, i.e. one username can be used from several edges at the same time. However, it is recommended to use a distinct username and password for each edge or computer. Your management port output will be much more meaningful then with the HINT column showing the related username. Also, the auto IP address feature, i.e. not using `-a <IP address>` at the edge, will more likely assign unique addresses as those depend on the username.
+Current supernode behavior does not limit the simultanious usage of usernames, i.e. one username can be used from several edges at the same time. However, it is recommended to use a distinct username and password for each edge or computer. Your management port output will be much more meaningful then with the `HINT` column showing the related username. Also, the auto IP address feature, i.e. not using `-a <IP address>` at the edge, will more likely assign unique addresses as those depend on the username.
 
-If a user chooses a new password or needs to be excluded from accessing the community (stolen edge scenario), the corresponding line of the `community.list` file can be replaced with a newly generated one or be deleted respectively. Restarting the supernode is required after performing changes because the supernode reads in this data only once at start-up.
+If a user chooses a new password or needs to be excluded from accessing the community (stolen edge scenario), the corresponding line of the `community.list` file can be replaced with a newly generated one or be deleted respectively. Restarting the supernode or issuing the `reload_communities` command to the management port is required after performing changes to make the supernode(s) read in this data again.
 
-With a view to the detailed explanations below, your supernode should have a non-default federation name given by the `-F <federation name>` command line parameter, e.g. `-F secretFed`. It is used to derive a priavte key on the supernode side and is only to be shared among supernodes. However, a federation-wide support for this feature, i.e. across several supernodes, is possible but not implemented yet.
+When using this feature federation-wide, i.e. across several supernodes, please make sure to keep all supernodes' `community.list` files in sync. So, if you delete or change a user one supernode (or add it), you need to do it at all supernodes. There is no built-in sync for the `community.list` files across the federation. External tools such as _Syncthing_ or your very own script-driven scp-based-file-distribution might be of assistance. Also, with every change, you need to restart the supernode or issue the `reload_communites` command to the management port as outlined above.
+
+With a view to the detailed explanations below, your supernode(s) should have a non-default federation name given by the `-F <federation name>` command line parameter, e.g. `-F secretFed`. It is used to derive a priavte key at the supernode side and is only to be shared among supernodes.
+
 
 #### Edge
 
@@ -84,18 +87,19 @@ Considering all this, our example expands to
 [user@host n2n]$ sudo ./edge -l <supernode:port> -c netleo -I logan -J 007 -A5 -k mySecretKey -P opIyaWhWjKLJSNOHNpKnGmelhHWRqkmY5pAx7lbDHp4
 ```
 
+You might want to consider the use of [`.conf` files](https://github.com/ntop/n2n/blob/dev/doc/ConfigurationFiles.md) to accomodate all the command line paramters more easily.
+
+
 #### How Does It Work?
 
 In order to make this authentication scheme work, the existing header encryption scheme is split into using two keys: a _static_ and a _dynamic_ one. The static key remains unchanged and is the [classic header encryption key](https://github.com/ntop/n2n/blob/dev/doc/Crypto.md#header) derived from the community name. It only is applied to the very basic registration traffic between edge and supernode (REGISTER_SUPER, REGISTER_SUPER_ACK, REGISTER_SUPER_NAK). The dynamic key is derived (among others) from the federation name – keep it secret! – and appplied to all the other packets, especially the data packets (PAKET) and peer-to-peer building packets (REGISTER), but also the ping and peer information (QUERY_PEER, PEER_INFO). An edge not provided with a valid dynamic key is not able to participate in the further communication.
 
 In regular header encryption mode, static key and dynamic key are equal. With activated user-password scheme, the supernode generates and transmits a dynamic key with the REGISTER_SUPER for further use. This happens in a secure way based on public key cryptography. A non-autheticated edge, i.e. without corresponding entry at the supernode or valid credentials, will not receive a valid dynmic key for communication beyond registration.
 
-In user-password scheme, the packets encrypted with the static key (REGISER_SUPER, REGISTER_SUPER_ACK, useless for REGISTER_SUPER_NAK) are "signed" with an encrypted outer hash using the shared secret which is only known to the federated supernodes and that specific edge.
+In user-password scheme, the packets encrypted with the static key (REGISTER_SUPER, REGISTER_SUPER_ACK, useless for REGISTER_SUPER_NAK) are "signed" with an encrypted outer hash using the shared secret which is only known to the federated supernodes and that specific edge.
 
 #### Possible Extensions
 
 Tools for automizing [`.conf` file](https://github.com/ntop/n2n/blob/dev/doc/ConfigurationFiles.md) generation for deployment ot delivery to freshly registered and approved users could greatly enhance this ecosystem; a user would not have to mess around with command line parameters but just copy a `.conf` file into a specified directory.
-
-Implementation of a federation-wide support for user-password authentication would allow much bigger networks using this feature. Still required steps are outlined [in some earlier post](https://github.com/ntop/n2n/issues/670#issuecomment-802119568).
 
 Let us know if you are intrerested in implementing or furthering these ideas.
--- a/include/n2n_typedefs.h
+++ b/include/n2n_typedefs.h
@@ -356,6 +356,7 @@ typedef struct n2n_REGISTER_SUPER {
     n2n_ip_subnet_t    dev_addr;    /**< IP address of the tuntap adapter. */
     n2n_desc_t         dev_desc;    /**< Hint description correlated with the edge */
     n2n_auth_t         auth;        /**< Authentication scheme and tokens */
+    uint32_t           key_time;    /**< key time for dynamic key, used between federatred supernodes only */
 } n2n_REGISTER_SUPER_t;
 
 
@@ -374,6 +375,8 @@ typedef struct n2n_REGISTER_SUPER_ACK {
      */
     uint8_t            num_sn;      /**< Number of supernodes that were send
                                        * even if we cannot store them all. */
+
+    uint32_t           key_time;    /**< key time for dynamic key, used between federatred supernodes only */
 } n2n_REGISTER_SUPER_ACK_t;
 
 
@@ -748,7 +751,6 @@ struct sn_community {
     he_context_t          *header_encryption_ctx_dynamic; /* Header encryption cipher context. */
     he_context_t                  *header_iv_ctx_static;  /* Header IV encryption cipher context, REMOVE as soon as separate fields for checksum and replay protection available */
     he_context_t                  *header_iv_ctx_dynamic; /* Header IV encryption cipher context, REMOVE as soon as separate fields for checksum and replay protection available */
-    uint32_t                      last_dynamic_key_time;  /* UTC time of last dynamic key generation (second accuracy) */
     uint8_t                       dynamic_key[N2N_AUTH_CHALLENGE_SIZE]; /* dynamic key */
     struct                        peer_info *edges;       /* Link list of registered edges. */
     node_supernode_association_t  *assoc;                 /* list of other edges from this community and their supernodes */
@@ -806,6 +808,7 @@ typedef struct n2n_sn {
     struct sn_community                    *federation;
     n2n_private_public_key_t               private_key;       /* private federation key derived from federation name */
     n2n_auth_t                             auth;
+    uint32_t                               dynamic_key_time;  /* UTC time of last dynamic key generation (second accuracy) */
     uint8_t                                override_spoofing_protection; /* set if overriding MAC/IP spoofing protection (cli option '-M') */
     n2n_resolve_parameter_t                *resolve_parameter;/*Pointer to name resolver's parameter block */
 } n2n_sn_t;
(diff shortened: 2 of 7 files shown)


galaxyskyknight:
Tested, looks work good, I have configured two supernodes in the same federation 'MyHomeFed' and 4 edges in the same community 'MyHomeN2N' under this federation domain. the community configure files are manually configured to same('sync') include all edges username/password in lines. As well particular username/password for the edge ( the username(-I) and password(-J) plus public federation name associated key) are configured to edge configure file. All edges could be inter-connected finally and the HINT is exactly matched to the configured username(see snap). 

Not have chance to do the performance test with iperf yet(I have two openwrt device in the same LAN but not running up in this time as I cannot compile the code yet, I am wondering if using this -I -J crypto could it decline the performance of connection speed? would like to test it in th LAN on this two openwrt devices). 

Btw: please kindly check in code ASAP as I could not compile under openwrt with the code not in git dev branch. thanks for this hard working effort.
supernode1
supernode2

galaxyskyknight:
so a futher question here, if I using -I -J for the edge auth, is there still any necessary to use -c -k for the communicty verification if I have only one community always under this federation for the home use purpose?

Logan007:
> Tested

Thank you!! :orange_heart:

> looks work good

I am glad to hear.

> if using this -I -J crypto could it decline the performance of connection speed?

It should be about the same speed as with header encryption enabled only (`-H`, without `-J`). Anything significantly beyond that, be it slower or faster, would be unexpected to me.

> please kindly check in code ASAP

Will do sometime during the next 24 hours if I do not hear back from anyone objecting. The remaining part could follow later.

> if I using -I -J for the edge auth, is there still any **necessary** to use **`-c` `-k`** for the communicty verification if I have only one community always under this federation for the home use purpose?

Yes as for the community parameter **`-c`** because there is no such thing like a default community. With `-J`, the community name still serves as key for the header encryption of REGISTER_SUPER, REGISTER_SUPER_ACK and REGISTER_SUPER_NAK packets, just as with regular header encryption `-H`. The other packet type's **headers** get encrypted with the new dynamic key, we have written down some [details](https://github.com/ntop/n2n/blob/dev/doc/Authentication.md#how-does-it-work).

Also, please consider that you might want to add more communities at a later point in time... for example, if your neighbors get interested in n2n'ing with you and one of your computers, you do not want them in the same community as your other computers. But you still could share the same supernodes without interference between the communities.

And yes for the **`-k`** key which takes care of the **payload** encryption of PACKET type messages, the actually transmitted data between the edges in ethernet frames. This key is only known to the user-operated edges and not known to the supernodes because they might be out of user's control. Without `-k`, the data would be transmitted in plain text despite the `-J`. But if I remember correctly, the edge will start with `-J` only if `-A4` or `-A5`, and those require a key `-k`, are provided along with it.

There would be a way to change it (no dedicated payload encryption) if we allowed the dynamic key (only exists with `-J`) to encrypt the whole packet. On the downside, also the supernode could decrypt the data then. As we do not trust the supernodes mostly residing on some hosted VPS, I did not follow this road.

Going full public key as I plan for the **4.x** series will allow to have not necessarily use a dedicated payload key for direct communication between peers, broadcast however will still require it. But ideas about all this are not laid out very well yet and thus still very "drafty" ...

galaxyskyknight:
Got, it is clear, thank you. and later on I managed to compile the openwrt version and I did the performance test for this code change on the two openwrt devices through Lan2Lan within 10 pairs(threads) via iperf3 tool. No performance decline at all. Just FYI.

galaxyskyknight:
BTW: it is occasionally printed out the following logs, what happened? it is a little bit noisy, can you make it as a higer level than WARNING for logs?

27/Jul/2021 14:09:57 [edge_utils.c:2355] WARNING: process_udp dropped packet encrypted with static key where dynamic key expected.

galaxyskyknight:
look like a bug or not @Logan007 when I netcat to port management port 5644 on each edge , there are two issues 
1) some other peer edge cannot display the ip address and user name in HINT which it is fine in edge side
2) not all  peer edges are present in a particular edge even I freshed by press enter multiple times

image

Logan007:
> 27/Jul/2021 14:09:57 [edge_utils.c:2355] WARNING: process_udp dropped packet encrypted with static key where dynamic key expected.

This happens when packets arrive and which are expected (by their type) be protected by the dynamic key, and in fact they are not.

I have not tested those cases extensively but this message presumably might be caused misconfiguration of one of the other 
edges (plain `-H` instead of `-J`) or other (local) edges with the same community name but different header encryption settings (either none, `-H` or `-J`).

In regular environments, this should not show up. Could you please check the other edges? What other messages do you see before this message when started with `-vvvvv`?

> not all peer edges are present in a particular edge even I freshed by press enter multiple times

Edge nodes maintain a list of only those edges with whom they currently exchange data. If there is no data-exchange (anymore), they forget about their partners after 30 seconds or. Connections are re-established via the supernode then. It helps scalability if not every edge has to maintain a list of all other edges at every point in time.

> some other peer edge cannot display the ip address and user name in HINT which it is fine in edge side

This happens whenever the direct connection was not (re-)established by a REGISTER message but a direct PACKET message. The PACKET messages do not carry the user identifier because that would add 16 or 20 bytes to every data packet sent out, decreasing available data space, or bandwidth respectively, by roughly 1.5 percent.

galaxyskyknight:
> > 27/Jul/2021 14:09:57 [edge_utils.c:2355] WARNING: process_udp dropped packet encrypted with static key where dynamic key expected.
> 
> This happens when packets arrive and which are expected (by their type) be protected by the dynamic key, and in fact they are not.
> 
> I have not tested those cases extensively but this message presumably might be caused misconfiguration of one of the other
> edges (plain `-H` instead of `-J`) or other (local) edges with the same community name but different header encryption settings (either none, `-H` or `-J`).
> 
all edges are uniformed to configure with -J but no -H, weired

> In regular environments, this should not show up. Could you please check the other edges? What other messages do you see before this message when started with `-vvvvv`?
> 
> > not all peer edges are present in a particular edge even I freshed by press enter multiple times
> 
> Edge nodes maintain a list of only those edges with whom they currently exchange data. If there is no data-exchange (anymore), they forget about their partners after 30 seconds or. Connections are re-established via the supernode then. It helps scalability if not every edge has to maintain a list of all other edges at every point in time.
> 
> > some other peer edge cannot display the ip address and user name in HINT which it is fine in edge side
> 
> This happens whenever the direct connection was not (re-)established by a REGISTER message but a direct PACKET message. The PACKET messages do not carry the user identifier because that would add 16 or 20 bytes to every data packet sent out, decreasing available data space, or bandwidth respectively, by roughly 1.5 percent.

So is it possible that  it only update this Table via REGISTER message but not via PACKET message? OR reuse the IP/USER last seen from REGISTER message when the PACKET message arrival ? w/o the IP and HINT user I cannot identify who is the currently peer connected at all.

Logan007:
> all edges are uniformed to configure with -J but no -H, weired

Seems to need further investigation then...

 > So is it possible that it only update this Table via REGISTER message but not via PACKET message? 

That's what actually happens.

> OR reuse the IP/USER last seen from REGISTER message when the PACKET message arrival ?

But if that peer was _purged_ (deleted) off the list because of no-traffic from remote peer, and the remote peer retries to send a PACKET which could be working without a new REGISTER if the port is still open in NAT or hasn't changed yet, the local peer sees the PACKET which comes without ID data (only MAC address).

> w/o the IP and HINT user I cannot identify who is the currently peer connected at all.

Only solution to current n2n would be to add `-I` data and IP data to each PACKET (just like with the REGISTER packet) at the cost of 1.5 percent of the bandwidth. Is it worth it?

By the way, **4.x** might have a different approach based on the hash of the peer's public key.

It might be of help to use a certain MAC address scheme and assign fixed MAC addresses according to your scheme (`-m` at the edge), e.g. make the last four of the MAC identical to the IP address.

Logan007:
By the way, how did you get the management port output colored?

galaxyskyknight:
> > all edges are uniformed to configure with -J but no -H, weired
> 
> Seems to need further investigation then...
> 
> > So is it possible that it only update this Table via REGISTER message but not via PACKET message?
> 
> That's what actually happens.
> 
> > OR reuse the IP/USER last seen from REGISTER message when the PACKET message arrival ?
> 
> But if that peer was _purged_ (deleted) off the list because of no-traffic from remote peer, and the remote peer retries to send a PACKET which could be working without a new REGISTER if the port is still open in NAT or hasn't changed yet, the local peer sees the PACKET which comes without ID data (only MAC address).
> 
> > w/o the IP and HINT user I cannot identify who is the currently peer connected at all.
> 
> Only solution to current n2n would be to add `-I` data and IP data to each PACKET (just like with the REGISTER packet) at the cost of 1.5 percent of the bandwidth. Is it worth it?
> 
> By the way, **4.x** might have a different approach based on the hash of the peer's public key.
> 
> It might be of help to use a certain MAC address scheme and assign fixed MAC addresses according to your scheme (`-m` at the edge), e.g. make the last four of the MAC identical to the IP address.

OK, I see, I believe 1.5 pt BW loss is not worth to do it. Yes, I assign the last section of MAC address align to the last section of IP is just  for making it easier to distinguish the edges is the right one from maintenance perspective .

galaxyskyknight:
> By the way, how did you get the management port output colored?

Have no idea at all....  I use the MobaXterm ssh client and there are may schemes for the terminal display and I used one of them, I guess this is where it comes from for those getting coloered display with dir/files/scripts/softlinks/mac/ip etc....

Logan007:
Last commit 12143f4 fixed a bug which interrupted connection between edges when a new supernode joinded the federation.

:warning: The latest commit also changed the way the dynamic key is calculated. So, if you want to test the code (linked in first post, it will always be pointing to current state of this PR branch) and already have tested it, you need to re-compile all your supernodes (maybe not edges, but you would want to do so to circumvent the bug mentioned) again.

Other than that, I decided to **not** implement the delayed REGISTER_SUPER handling at the supernode in case of user/password authentication scheme. One has just to keep the `community.list` (or whatever the `-c` provided file is called in the specific scenario) files of all supernodes in sync – and do not forget to `reload_communities`.

So, final bug-hunting for this PR! Let me know what you find. Meanwhile, I will have a deeper look into that `WARNING: process_udp dropped packet encrypted with static key where dynamic key expected.` message.

galaxyskyknight:
> Last commit [12143f4](https://github.com/ntop/n2n/commit/12143f43036c23a8f602268edab0c3f32f76e88a) fixed a bug which interrupted connection between edges when a new supernode joinded the federation.
> 
> ⚠️ The latest commit also changed the way the dynamic key is calculated. So, if you want to test the code (linked in first post, it will always be pointing to current state of this PR branch) and already have tested it, you need to re-compile all your supernodes (maybe not edges, but you would want to do so to circumvent the bug mentioned) again.
> 
> Other than that, I decided to **not** implement the delayed REGISTER_SUPER handling at the supernode in case of user/password authentication scheme. One has just to keep the `community.list` (or whatever the `-c` provided file is called in the specific scenario) files of all supernodes in sync – and do not forget to `reload_communities`.
> 
> So, final bug-hunting for this PR! Let me know what you find. Meanwhile, I will have a deeper look into that `WARNING: process_udp dropped packet encrypted with static key where dynamic key expected.` message.

OK, I will find some time to test it, btw, how to issue 'reload_communities' what is its full command?

Logan007:
> how to issue 'reload_communities' what is its full command?

Just write `reload_communities` to the supernode's management port. Please find some hints over at [that other pull request](https://github.com/ntop/n2n/pull/740#issuecomment-886443065)... :wink:

galaxyskyknight:
> > how to issue 'reload_communities' what is its full command?
> 
> Just write `reload_communities` to the supernode's management port. Please find some hints over at [that other pull request](https://github.com/ntop/n2n/pull/740#issuecomment-886443065)... 😉

I believe there are some problem probabaly on this version, I observed the edge is failover between two supernode frequently and much more times that the last version, I don't know what happend ,say I have 8 edges, the ususal case is 4 on 1 supernode and left 4 on another and in most time it is stable, but in this version, it is always seen all edges are flap into one supernode but no left on another then it has changed to re-balance the load.  In on word, it is not "stable' to keep the edge attach, but trigger the edge flap over between master node and slave.

galaxyskyknight:
> > > how to issue 'reload_communities' what is its full command?
> > 
> > 
> > Just write `reload_communities` to the supernode's management port. Please find some hints over at [that other pull request](https://github.com/ntop/n2n/pull/740#issuecomment-886443065)... 😉
> 
> I believe there are some problem probabaly on this version, I observed the edge is failover between two supernode frequently and much more times that the last version, I don't know what happend ,say I have 8 edges, the ususal case is 4 on 1 supernode and left 4 on another and in most time it is stable, but in this version, it is always seen all edges are flap into one supernode but no left on another then it has changed to re-balance the load. In on word, it is not "stable' to keep the edge attach, but trigger the edge flap over between master node and slave.

sometimes all edges register in one node and looks like load balance not working 

image
image

Logan007:
The regular load balance should choose the one with the lowest load at that moment, even taking some sort of "stickyness" into account. Usually, it balances and equalizes quite well with more than three edges – at least over time.

The load balancing portion has been remaining untouched for a while now. But,  the RE_REGISTER for dynamic key update of user/pw auth happens to all edges at the very same time (after a new supernode has joined). Normally, this should not be a problem because they first try to re-register to the current supernode, the same as before.

Interesting observation... Please give it some time, a few minutes after the last supernode joined it should be alright. How long after last supernode join or last `re_register` did you see this? _I am considering a light randomization of re-registration interval._

The fail-over itself is not a critical thing to see, a TCP connection between two n2n edges, be it supernode-forwarded or not, should not even be interrupted by that.

By the way, have you tried the [RTT-based selection](https://github.com/ntop/n2n/blob/dev/doc/Building.md#federation--supernode-selection-by-round-trip-time)?

Logan007:
If this persists, I would be very interested to know what supernode load values you see at the edges' management port. They might differ per edge as they include the "stickyness" factor and some local information if I remember correctly.

galaxyskyknight:
I have one shot pressing etner time observed a very weired load  value from edge perspective but next enter it has changed to normal, does it impact anything?
image

galaxyskyknight:
I have one shot time by pressing enter key that observed a very weired load value from edge perspective but  it changed to normal on next shot, does it impact anything?
image

galaxyskyknight:
> If this persists, I would be very interested to know what supernode load values you see at the edges' management port. They might differ per edge as they include the "stickyness" factor and some local information if I remember correctly.

currently its load is 13/17 and I have nine edges registered on 4:5 ratio on 2 supnerodes, will observe if above I reported could be reproduced or gone, I haven't check the edge managment port for the load param in last testing.

Logan007:
`13` / `17` load ratio at `4` / `5` edge ratio sounds OK. The 13 looks that small because it artificially is diminished locally for stickyness so that the edge does not immediately hop to the next supernode at the slightest draft.

The `21474836` is not supposed to show up, it is integer representation of a negative number (used for internal purposes when determining _bad_ supernodes or supernodes with _unknown status_. This might happen due to  re-registration as well and is normal. What is **not** normal is that it shows up in the list, the field should be empty instead. I will check the output function.

Thanks  for reporting and thus helping us to build a better n2n!! :orange_heart:

galaxyskyknight:
> > If this persists, I would be very interested to know what supernode load values you see at the edges' management port. They might differ per edge as they include the "stickyness" factor and some local information if I remember correctly.
> 
> currently its load is 13/17 and I have nine edges registered on 4:5 ratio on 2 supnerodes, will observe if above I reported could be reproduced or gone, I haven't check the edge managment port for the load param in last testing.

I didn't see any more for above problem, so I believe that is not an issue for now, I guess probably when I deploy the supernode there is  a temporary high CPU on one node caused the hop and it is not persist there.

galaxyskyknight:
> `13` / `17` load ratio at `4` / `5` edge ratio sounds OK. The 13 looks that small because it artificially is diminished locally for stickyness so that the edge does not immediately hop to the next supernode at the slightest draft.
> 
> The `21474836` is not supposed to show up, it is integer representation of a negative number (used for internal purposes when determining _bad_ supernodes or supernodes with _unknown status_. This might happen due to re-registration as well and is normal. What is **not** normal is that it shows up in the list, the field should be empty instead. I will check the output function.
> 
> Thanks for reporting and thus helping us to build a better n2n!! 🧡

welcome, I believe N2N is a very helpful software, hope we could get it  better:)

Logan007:
a749b3a contains some minor fixes.

Logan007:
Works extremely stable over here (limited test setup). Will merge within 24 hours if no complaints.

With this milestone passed, 3.0 is getting very close... (_doc_ and _log_ left as open action items then).

PR #744, opened 2021-08-05
reviewed log messages

Logan007:
This pull request tries to unify log message format to at least some extent. Also, it adjust verbosity level of some messages aiming at making the edge a bit less noisy in standard mode, i.e. if no `-v` (or more) provided.

As far as I can see, this is one of the last [milestones to pass on our way to version 3.0](https://github.com/ntop/n2n/issues/562).

This is work in progress and thus marked as draft. Just pushing intermediate results for everybody to review if required.


Logan007:
a probably never-ending task :wink:

PR #755, opened 2021-08-29
correct wrong switch logic

fengdaolong:


Logan007:
Thank you!!

Logan007:
I have reverted this pull request as Windows builld [does not digest the changes](https://github.com/ntop/n2n/runs/3453650957?check_suite_focus=true).

Issue #758, opened 2021-08-30
Segmentation fault


skyformat99:
The latest version, after compiling, ran example_edge_embed_quick_edge_init and the following error occurred:

[root@demo build]# ./example_edge_embed_quick_edge_init
30/Aug/2021 09:56:48 [n2n.c:295] supernode2sock successfully resolves supernode IPv4 address for 7.8.9.10
30/Aug/2021 09:56:48 [edge_utils.c:3616] adding supernode = 7.8.9.10:1234
30/Aug/2021 09:56:48 [tuntap_linux.c:204] Waiting for TAP interface to be up and running...
30/Aug/2021 09:56:48 [tuntap_linux.c:225] Interface is up and running
30/Aug/2021 09:56:48 [edge_utils.c:331] number of supernodes in the list: 1
30/Aug/2021 09:56:48 [edge_utils.c:333] supernode 0 => 7.8.9.10:1234
30/Aug/2021 09:56:48 [transform_aes.c:210] setup_aes_key 128-bit key setup completed
30/Aug/2021 09:56:48 [edge_utils.c:1049] successfully joined multicast group 224.0.0.68:1968
30/Aug/2021 09:56:48 [edge_utils.c:1493] registering with supernode [7.8.9.10:1234][number of supernodes 1][attempts left 2]
30/Aug/2021 09:56:48 [edge_utils.c:1174] send REGISTER_SUPER to [7.8.9.10:1234]
30/Aug/2021 09:56:48 [edge_utils.c:526] registering with multicast group 224.0.0.68:1968
30/Aug/2021 09:56:48 [edge_utils.c:1318] send REGISTER to [224.0.0.68:1968]
30/Aug/2021 09:56:50 [n2n.c:614] Purging old registrations
30/Aug/2021 09:56:50 [n2n.c:619] Remove 0 registrations
30/Aug/2021 09:56:50 [edge_utils.c:1101] send PING to supernodes
Segmentation fault

After the previous version was compiled, the run was OK.

skyformat99:
I think we know why,
This is caused by modifying the CMakefile.txt file.
The problem was resolved after only restoring the  4/June   CMakefile.txt  file.

Logan007:
Thank you for reporting and fixing!

I am absolutely not confident with the build systems which still cause some confusion at my side. Excuses upfront!

[...]

PR #759, opened 2021-08-30
Update CMakeLists.txt

skyformat99:
fix Segmentation fault #758


Logan007:
Thank you!!

fengdaolong:
Why not let other operating systems other than windows use libpthread???????????????????????????????????

Logan007:
Please see my comment over at #758. I think the underlying issue still needs some care.

fengdaolong:
Please revert this PR and set libpthread to be disabled by default `OPTION (N2N_OPTION_USE_PTHREAD "USE PTHREAD Library" OFF)`, wait until the problem is completely resolved, and then enable libpthread by default as appropriate.

Logan007:
> Please revert this PR and set libpthread to be disabled by default `OPTION (N2N_OPTION_USE_PTHREAD "USE PTHREAD Library" OFF)`

I will as interim. Let's discuss a sustainable solution to this over at #758.

PR #762, opened 2021-08-30
Segmentation fault #758

skyformat99:
fix Segmentation fault #758

There are cases where the pointer is null.


Logan007:
Thanks for diving into that. :+1:

I will have a look.

Logan007:
Thank you!!

I have wrapped the if around the rest of the code.

fengdaolong:
"If" statements are nested too much, it is recommended to write:
```
if(param == NULL) return ret;
```

Logan007:
D'accord.

Readibility upgrade :wink:

PR #770, opened 2021-09-07
fixed optional SKIP_MULTICAST_PEERS_DISCOVERY macro

Logan007:
This pull request allows proper use of `SKIP_MULTICAST_PEERS_DISCOVERY` macro for configuration, e.g.

`-DSKIP_MULTICAST_PEERS_DISCOVERY`

to disable local multicast peer detection (listening as well as querying).

### DIFF
--- a/doc/Building.md
+++ b/doc/Building.md
@@ -168,3 +168,18 @@ Your specific ARM mileage may vary, so it can be enabled by configuring the defi
 `./configure CFLAGS="-DSPECK_ARM_NEON"`
 
 Just make sure that the correct architecture is set, too. `-march=native` usually works quite well.
+
+## Disable Nulticast Local Peer Detection
+
+For better local peer detection, the edges try to detect local peers by sending REGISTER
+packets to a certain multicast address. Also, edges listen to this address to eventually
+fetch such packets.
+
+If these packets disturb network's peace or even get forwarded by (other) edges through the
+n2n network, this behavior can be disabled, just add
+
+`-DSKIP_MULTICAST_PEERS_DISCOVERY`
+
+to your `CFLAGS` when configuring, e.g.
+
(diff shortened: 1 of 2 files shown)


fengdaolong:
Multicast is disabled by default. Multicast can only be used if the `-E` option is set.
I don't quite understand why it is to disable multicast by compiling options.

Logan007:
`-E` is about accepting multicast packets (and handling / forwarding) at the **TAP** or not, allowing the applications using the TAP to work with multicast packets (or not).

The macro `SKIP_MULTICAST_PEERS_DISCOVERY` is about the multicast detection of other peer edges inside the real LAN.

fengdaolong:
If I remember correctly, the `-E` option is for real LAN, not for TAP.

fengdaolong:
if I recall it correctly, the real LAN needs multicast to discover peers each other, and there is no need to multicast between TAP.
The purpose of introducing multicast at first was to realize the point-to-point direct connection of the local area network, although the effect was not very good.

Logan007:
> if I recall it correctly, the real LAN needs multicast to discover peers each other

Yes, this is correct.

> The purpose of introducing multicast at first was to realize the point-to-point direct connection of the local area network, 

In regular local networks, it should find all peer edges of that same broadcast/multicast domain.

> although the effect was not very good.

It might be supplemented by the new `-e` option in case multicast is not available or broadcast/multicast domain is limited

> and there is no need to multicast between TAP

Well, this depends on your needs and the applications you want to run over the n2n network. Some online games might require it as well as DLNA and others. So, for some users `-E` enabling multicast over TAP is essential.

fengdaolong:
>Some online games might require it as well as DLNA and others. So, for some users -E enabling multicast over TAP is essential.

You are right, one is for local multicast and the other is for tunnel multicast. I misunderstood it, you are great! 😏️

PR #776, opened 2021-09-11
allowed edge to optionally auto-detetct local IP address ('-e auto') for advertising as preferred

Logan007:
This pull request adds auto-detection of local IP address for use with `-e`:

`-e auto`

Inspired by @meyerd 's fork.

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -203,8 +203,10 @@ void reset_sup_attempts (n2n_edge_t *eee) {
 int supernode_connect(n2n_edge_t *eee) {
 
     int sockopt;
-    struct sockaddr_in sock;
-    int sock_len = sizeof(sock);
+    struct sockaddr_in sn_sock;
+    struct sockaddr_in local_sock;
+    int sock_len = sizeof(local_sock);
+    SOCKET probe_sock;
 
     if((eee->conf.connect_tcp) && (eee->sock >= 0)) {
         closesocket(eee->sock);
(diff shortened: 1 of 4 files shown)


src/edge_utils.c
```diff
@@ -275,6 +266,28 @@ int supernode_connect(n2n_edge_t *eee) {
...
+        && (sock_len == sizeof(local_sock))) {
+            // ... and write to local preferred socket -- no matter if used or not
+            eee->conf.preferred_sock.port = ntohs(local_sock.sin_port);
+            // probe for & overwrite local address only if auto-detection mode
+            if(eee->conf.preferred_sock_auto) {
+                probe_sock = socket(PF_INET, SOCK_DGRAM, 0);
+                // connecting the UDP socket makes getsockname read the local address it uses to connect (to the sn in this case)
+                // we cannot do it with the real (eee->sock) socket because socket does not accept any conenction from elsewhere then,
+                // e.g. from another edge instead of the supernode; hence, we use a temporary socket
+                connect(probe_sock, (struct sockaddr *)&sn_sock, sizeof(sn_sock));
```
fengdaolong, line 281: Please take the acquisition of the local IP address as an independent public function, which is also needed in other places, using `getsockname` to get the local address is very concise. I think it is possible to delete the `get_best_interface_ip` in `src/edge_utils_win32.c` and replace it.

Logan007: Can do. Not sure about the windows part though.

Logan007: #781

PR #781, opened 2021-09-13
separated 'detect_local_ip_address()' function

Logan007:


src/n2n.c
```diff
@@ -61,6 +61,54 @@ SOCKET open_socket (int local_port, in_addr_t address, int type /* 0 = UDP, TCP
...
+            if((getsockname(probe_sock, (struct sockaddr *)&local_sock, &sock_len) == 0)
+            && (local_sock.sin_family == AF_INET)
+            && (sock_len == sizeof(local_sock))) {
+                memcpy(&(out_sock->addr.v4), &(local_sock.sin_addr.s_addr), IPV4_SIZE);
+            } else
+                return -4;
+        } else
+            return -3;
+        close(probe_sock);
+    } else
```
fengdaolong, line 102: The code here needs to be optimized, and it needs to be `close` before each return.

Something like this:
```
int ret = 0;
probe_sock = socket(PF_INET, SOCK_DGRAM, 0);
if(){
    ...;
    if(){
        ...;
    } 
    else {
        ret=-1;
        goto end;
    }
} 
else {
    ret=-2;
    goto end;
}

end:
close(probe_sock);
return ret;
```

src/n2n.c
```diff
@@ -61,6 +61,54 @@ SOCKET open_socket (int local_port, in_addr_t address, int type /* 0 = UDP, TCP
 }
 
 
+int detect_local_ip_address (n2n_sock_t* out_sock, const n2n_edge_t* eee) {
```
fengdaolong, line 64: `n2n.c` is a file shared by edge and supernode, so there should be no `n2n_edge_t* eee`.

src/n2n.c
Logan007: You are right. Will fix.

src/n2n.c
Logan007: You are right, will move it to `edge_utils.c`.

PR #790, opened 2021-09-15
Turn on -Wall and address most messages

hamishcoleman:
Most of what is still generating messages is the many miss-matches between the string functions wanting signed char and the n2n callers often using unsigned char - which has a style component to the choice of how to address.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -22,6 +22,7 @@ ifeq ($(OPENSSL_CFLAGS), 0)
   CFLAGS+=$(shell pkg-config --cflags-only-I openssl)
 endif
 
+WARN=-Wall
 CFLAGS+=$(DEBUG) $(OPTIMIZATION) $(WARN) $(OPTIONS) $(PLATOPTS)
 
 INSTALL=install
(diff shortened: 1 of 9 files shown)


skyformat99:
-vvvvv  option can display all messages.

Logan007:
Thank you very much!!

Logan007:
@skyformat99 I think there is a misunderstanding here. This pull request addresses compile-time warnings.

hamishcoleman:
@Logan007 - what is your opinion on the string functions?  We could go through and change anything param/struct that is treated as a zero-terminated-string into a (char *) or could typecast any (uint8_t *) when we call the string lib functions.

I started going through and changing things-used-as-strings into (char *), but stopped as I thought there might be opinions on this.

Logan007:
> We could go through and change anything param/struct that is treated as a zero-terminated-string into a (char *) 

I fully agree!

> as I thought there might be opinions on this.

However, for internal crypto, I actually do prefer the `uint8_t` because it directly compares to the other `uintXX_t` types. I assume that we probably will not get it all too distinct and always see one or two overlaps requiring type cast at some point.

> I started going through and changing things-used-as-strings into (char *), but stopped ...

This is amazing and please, don't stop! :rocket:

fengdaolong:
Well done, there are still a lot of warnings when compiling, there are types that do not match, and there are shadowed declaration, which need to be eliminated one by one.

hamishcoleman:
@fengdaolong you are absolutely correct - I fixed the low hanging fruit first.  I'm looking at the types that do not match next

PR #793, opened 2021-09-16
Address all but one of the remaining compile warnings

hamishcoleman:


Logan007:
Thank you very much!!

PR #795, opened 2021-09-16
Simplify the makefiles

hamishcoleman:


fengdaolong:
I think there is no need to rename the source file.

hamishcoleman:
The renaming is needed to reduce the number of special cases in the Makefile - if the sourcecode matches the binary then the implicit make rules can do a lot of the heavy lifting for us

fengdaolong:
Then `CMakeLists.txt` also needs to be modified synchronously.

hamishcoleman:
I had thought that the cmake system was deprecated, and when I noticed that it still had a CI attached to it, I did update the CMakeLists :-S

fengdaolong:
One day in the future, we will only use CMakeLists.txt instead of Makefile to reduce the maintenance burden.

hamishcoleman:
In the past, n2n did only use Cmake - the Makefiles and autoconf are the new make system ..

hamishcoleman:
(besides, this is one of those vi vs emacs issues - as I use Makefiles, I can maintain them; and conversely, I found it really difficult to follow the Cmake logic when I needed to update that in the past)

fengdaolong:
`Cmake` does have a great advantage in cross-platform compilation. If you can't choose one of the two, you can only coexist the two. The problem is that both of them need to be maintained synchronously. If one party's details are not done well, the compilation expectations may also be different.

Logan007:
>  I found it really difficult to follow the Cmake logic when I needed to update that in the past

I can feel your pain!

> One day in the future, we will only use CMakeLists.txt instead of Makefile to reduce the maintenance burden.

While I can work with both, not really maintain neither of them, I have a slight preference towards the Makefiles...

> as I use Makefiles, I can maintain them

... so, it would be really great if you could do that!

Do you think this pull request is ready to be merged?

hamishcoleman:
I can certainly keep looking at the Makefiles.  At the moment, I've just been shoving stuff around when it hurts - so I will keep doing that and see where we can get to.

I think this pull request is ready to be merged, yes.

Logan007:
Thank you very much for going into the messy corners!!

PR #797, opened 2021-09-16
Simple test framework and code coverage reporting

hamishcoleman:
While thinking about the interoperability issues that have been reported when building in different environments and trying to communicate with nodes from other compile environments, it was clear that a set of simple tests and known good outputs from these tests could help.  (In addition to helping with regression testing and all the usual arguments for tests)

I am proposing a simple framework in this PR - one that does not require any external libraries or a steep learning curve and thus should fit with the aim to be small and portable for the rest of the project.  This PR is initially for discussion and not merging.

Now that I have a framework, I have some more tests that I will push into this branch.


Logan007:
:exclamation: 

Impressive!

hamishcoleman:
now watch as I fight the CI robot

hamishcoleman:
Note that the github action in this PR will force upload a coverage report to the gh-pages branch on any push to any repository.  Which is only an issue if there is something else already using the gh-pages branch (not in this n2n repo)

Logan007:
This is absolutely amazing!

> Note that the github action in this PR will force upload a coverage report to the gh-pages branch on any push to any repository. Which is only an issue if there is something else already using the gh-pages branch (not in this n2n repo)

Do you know any way to change this behavior? Test seemingly even kick in after commenting an issue... EDIT: maybe that was just overlapping with your new commit... :wink:

hamishcoleman:
I think this now demonstrates the basic ideas.  There are functional tests for the same things that the n2n-benchmark does speed tests on and one test giving an example for testing the wire packets.

The CI should only run when code is pushed into the branch - so yes, it was probably overlapping.

I've setup the coverage report to only upload data when it the action is a push - which should make sure that it only runs when you have perms to do the push.  However, if you have more than one branch with this CI module running in it any other branch pushed will overwrite any previous coverage report - this _should_ not be a problem, since the most use for the report is for instant checks during dev (but can be revisited)

An example report can currently be seen at https://hamishcoleman.github.io/n2n/

hamishcoleman:
This now uploads the code coverage report as a "artifact" - which makes the report significantly harder to read (requires you to download and unzip the file before you can open it in a web browser), but also allows every test run to keep its own report

hamishcoleman:
@Logan007 I think this branch is ready to merge now.

lucaderi:
@ aouinizied You are an expert of GitHub actions: what do you think about this PR?

aouinizied:
@hamishcoleman thanks for this great contribution.
As this PR is intended for discussing, I have some questions/suggestions:
* Instead of uploading the cov report as an artefact, why do not use codecov action? See https://about.codecov.io/blog/how-to-set-up-codecov-with-c-and-github-actions/ for a simple example.
* For push/branch setting you can add something like the following where you define a clear/strict strategy:
``` yaml
on:
  push:
    branches:
      - dev
  pull_request:
    branches:
      - dev
    types: [opened, synchronize, reopened]
  release:
    types: [created]
``` 
* As you mentioned that these tests are intended to fix problems from different platforms, why do not add a matrix with ubuntu-latest, ubuntu-18.04, macOS-11, windows-latest?

``` yaml
jobs:
  tests:
    name: Test on ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: true
      matrix:
        os: ["ubuntu-latest", "ubuntu-18.04", "macOS-latest", "macos-11", "windows-latest"]

    steps:
      - uses: actions/checkout@v2

      - name: Install essential
        run: |
          sudo apt-get update
          sudo apt-get install build-essential
``` 

Zied

aouinizied:
If you go with codecov, do not forget to add a nice badge on the readme showing your coverage. BTW, the current build badge is still referring to Travis.

Zied

hamishcoleman:
Thanks for the review @aouinizied, I originally didn't want to add a dependency on another external service (like codecov), but as long as there is a way to create a coverage report locally that is not a problem (Also, while the codecov details are much more extensive, IMNSHO I actually think that the gcovr report is more useful)

I dont want to limit the test runs - I want all my personal branches to run them and any PR's.  (However, if the codecov uploader doesnt cope with that, that job step could be conditional)

I always planned to either add a matrix or add specific builders for each OS - I dont expect these instructions to work unchanged on non Linux, so I was making it simple to start with.  I also see the need to add each of the 'supported' compile-time options and optional libraries to the matrix.

hamishcoleman:
(Oh, and I wanted to add a badge, but none of the options scale across forks and branches - all hardcoding the upstream.. so until this is all merged, there is no sane testing of any badges)

(Yes, the travis-ci badge should go)

Logan007:
So, do you think this can be merged?

hamishcoleman:
@Logan007 , yes, I believe that this is ready to merge

Logan007:
OK... let's see what happens :wink:

Logan007:
Thank you very much!!

PR #798, opened 2021-09-17
Fix build error when libcap-dev is installed

hamishcoleman:


hamishcoleman:
See also 829ba29a2bcb0184a0fd8adeb11536634b9c2760, however this PR is a cleaner fix

lucaderi:
Thanks for the contribution.

PR #829, opened 2021-09-29
Test platform mingw - For discussion

hamishcoleman:
This updates the automated tests to run on windows/mingw.

Which has required a number of changes
- autoconf needed to be replaced for this build
- none of the configure options are available
- the Makefiles have been expanded and overhauled to add support for win32

Most importantly, while the build completes, the created library doesnt create the same transform output as a version built on ubuntu.  Which makes the build fail the tests.

I have also documented a simple set of steps to make the same build locally and this branch along with those instructions can be used to assist in debugging the issue with the transform.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,6 +29,7 @@ jobs:
           - ubuntu-latest
           - ubuntu-18.04
           - macos-latest
+          - windows-latest
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -50,7 +50,7 @@ jobs:
         name: Hack up a fake autoconf
         run: |
           echo true >autogen.sh
-          cp scripts/hack_fakeautoconf configure.sh
+          cp scripts/hack_fakeautoconf configure
 
       - name: generate a makefile and use it to install more packages
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -76,13 +76,19 @@ jobs:
           name: tests-out-${{matrix.os}}
           path: tests/*.out
 
-      - name: Generate coverage reports
+      - name: Generate coverage data
         run: |
           make cover
+        shell: bash
+
+      - if: runner.os != 'Windows'
+        name: Generate gcovr report
+        run: |
           make gcov
         shell: bash
 
-      - name: Upload gcovr report artifact
+      - if: runner.os != 'Windows'
+        name: Upload gcovr report artifact
         uses: actions/upload-artifact@v2
         with:
           name: coverage-${{matrix.os}}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -78,13 +78,13 @@ jobs:
 
       - name: Generate coverage data
         run: |
-          make cover
+          make gcov
         shell: bash
 
       - if: runner.os != 'Windows'
         name: Generate gcovr report
         run: |
-          make gcov
+          make cover
         shell: bash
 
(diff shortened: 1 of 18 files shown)


hamishcoleman:
@Logan007 hopefully, this appeases your curiosity

Logan007:
Indeed, it does! And to some extent, I feel urged to jump right into MinGW debugging...

hamishcoleman:
jump in - thats what I was hoping to inspire!

(and I'm not so good on debugging crypto miscompiles)

Logan007:
Still fighting with mingw compilation...

I had to extend the `-I` to `./win32` and `.` in Makefile.

But then, I see a lot of `undefined reference to` ... network related stuff such as __imp_ntohs and others...

hamishcoleman:
Strange, those are all things that I fixed already.  Can you share your build log?

hamishcoleman:
It should not have made any difference for your testing, but the branch name in my instructions was an old branch name.

Logan007:
Please see the [log](https://github.com/ntop/n2n/files/7253032/log.txt).

hamishcoleman:
Well, I'm running all my steps from a bash shell.  Which probably means that any of the shell commands that the Makefile run dont work - which I think has meant that the Makefile didnt realise that it was running on windows/mingw

Logan007:
I ran the script on bash (Linux) as well, removed the ^M and copied back to Windows...

Do you use bash on Windows?

hamishcoleman:
I used bash on Windows, yes.  The Makefile needs to run with a bash shell as well as the hack_fakeautoconf script.

From the desktop, I could start/run/bash and it started, or from a windows ssh session, I could run "C:\Program Files\Git\usr\bin\bash"

Logan007:
Meanwhile... I have been able to get it compiled. Thanks for setting me up! Will be gone for a while...debugging :wink:

hamishcoleman:
Good Luck!

Let me know if you come up against anything else in the buildsystem or docs that might trap people.

Logan007:
Interesting (?) observation: I did not use choco but `sudo apt-get install gcc-mingw-w64-i686` (Ubuntu subsytem installed).

The tests do not complain...

hamishcoleman:
The Makefile system is hardcoded to use regular gcc, so cross-compiling from a linux container will just use the containers system compiler.

Logan007:
I think I found the bug! (#830) Hopefully...

Without your help, I wouldn't even have gotten a working MinGW compile. Again, thank you!

I guess this PR needs some changes before merging?

hamishcoleman:
It is not ideal as it is, no (though it was surprisingly easy to rip out the autotools) - it is probably better to add easier cross compiling support to the Makefiles, and then using gcc-mingw-w64-i686

hamishcoleman:
Yey, it can pass the tests automatically!

Logan007:
So, shall we merge?

hamishcoleman:
I want to look closer at the diff and see if I can make it less invasive.

I've also been trying to work out how to make mingw on ubuntu compile the same code, so I'm looking at making the compile time target environment detection a little more generic and reusable - which might help justify some of the other changes in here.

Logan007:
Okay, just let me know.

By the way, I think the new compile instructions somehow got placed in the middle of some other section...

hamishcoleman:
ha, I think I threw them anywhere in a mild hurry.  They need looking at anyway - now you have added your patch, they need to stop saying "do not use" :-)

hamishcoleman:
@Logan007 I am happier with this branch now.

There is one change that might trip some people up - with these patches applied, you can no longer "cd tools; make" but must run the make from the top dir.

Logan007:
Thank you very much for this great contribution!

PR #830, opened 2021-09-29
fixed MinGW compile

### LINKED ISSUE #642: Cannot communicate between win and linux  (V2.9.0 r875)
edge v.2.9.0.r875.8589471
supernode v.2.9.0.r875.8589471

Cannot communicate between win and linux.
Win and win or linux an linux can communicate.
It seems that win cannot get the mac address of linux edge

### LINKED ISSUE #785: about pinging between Windows and Linux
1) use last code to compile binary file on Windows and Linux;
2) On linux, Run the server and client separately:
     ./supernode -p 8888 -f
     ./edge -c xxx  -l 127.0.0.1:8888 -d zz -f -a 192.168.188.10 -k xxx
3) on Windows, run the client:
     edge -c xxx  -l   xx.xx.xx.xx:8888 -a 192.168.188.11 -k xxx
4)  from windows ping linux:
     ping 192.168.188.10
Pinging 192.168.188.10 with 32 bytes of data:
Reply from 192.168.188.11: Destination host unreachable.
Request timed out.
Request timed out.

The two edges cannot communicate with each other.

but,  if remove **-k xxx** ,   
They can communicate with each other.
It seems that the encryption method is different under different OS.

Logan007:
For yet unknown reason, `_byteswap_uint64()`  does not work as expected if compiled with MinGW. Even the replacement taken from  a few lines above (line 49) requires adjustment in terms of nulling the upper bits. As if _arithmetic_ shifting kicks in at some place (due to different typing?).

### DIFF
--- a/include/portable_endian.h
+++ b/include/portable_endian.h
@@ -135,8 +135,8 @@
 #       define be32toh(x) _byteswap_ulong(x)
 #       define le32toh(x) (x)
 
-#       define htobe64(x) _byteswap_uint64(x)
-#       define be64toh(x) _byteswap_uint64(x)
+#       define htobe64(x) (((uint64_t)htonl(((uint32_t)(((uint64_t)(x)) >> 32))) & 0x00000000FFFFFFFFULL) | (((uint64_t)htonl(((uint32_t)(x)))) << 32))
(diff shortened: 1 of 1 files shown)


Logan007:
credits to @hamishcoleman

PR #849, opened 2021-10-11
Add automated binary artifacts

hamishcoleman:
This adds steps to the tests workflow to generate installable packages from a branch or PR.

Currently:
- dpkg built on "ubuntu-latest" with amd64 binaries
- rpm built via rpmbuild on ubuntu with amd64 binaries
- zip with windows amd64 binaries built on "windows-latest"
- zip with macos amd64 binaries built on "macos-latest"

Also possible is an openwrt builder, however that build takes over an hour to complete for each architecture, so should probably be kept separate (See https://github.com/hamishcoleman/n2n/actions/runs/1282129789 for a worked example of that)

All generated packages are uploaded as artifacts and are available to download from the summary page for this workflow once the entire workflow completes.


Logan007:
Hamish Coleman, this is great work and you do a lot of good to the project!!

I am extremely tempted to merge, but I need just one question answered, please. I see some [warning annotations](https://github.com/ntop/n2n/actions/runs/1328626230). What do those mean?

hamishcoleman:
One of the warnings is simply github telling us that the version of "macos-latest" will change soon.  However, the other ones are more strange - I never saw these annotations in my (many) test runs.

I've looked closer and it appears that a part of the github action infrastructure had issues downloading some of its requirements (You can see this by drilling down into the "Set up job" notes for the ubuntu-latest and windows-latest builds)

So, it looks like the github API had a momentary reachability blip.  I'll push a change and see if the warnings still show in the next build :-S

hamishcoleman:
@Logan007 a quick re-run of the build and this time there are no warning annotations ...  so, a github hiccup?

Logan007:
Looks good, seems indeed to have been a temporary issue.

If you do not want to revert the _Smoke Test Builds_, which I personally like, you could just rename the "testing" to "Testing".

hamishcoleman:
yeah, I had been thinking of renaming that to Smoke test anyway - so that can stay.  But capitalising the workflow name makes sense too.

Logan007:
Thank you, your contribution is very much appreciated! :orange_heart:

PR #850, opened 2021-10-13
Add some crosscompiled binary outputs to the autobuild

hamishcoleman:
Hopefully this will also demonstrate how to do a cross compile with the makefiles.  

The cross compile process is summarised as:
```
HOST_TRIPLET=arm-linux-gnueabi
sudo apt-get install binutils-$HOST_TRIPLET gcc-$HOST_TRIPLET
./autogen.sh
export CC=$HOST_TRIPLET-gcc
export AR=$HOST_TRIPLET-ar
./configure --host $HOST_TRIPLET
make
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Upload dpkg
         uses: actions/upload-artifact@v2
         with:
-          name: package-dpkg-x86_64-pc-linux-gnu
+          name: package-dpkg-amd64
           path: packages/debian/*.deb
 
   package_rpm:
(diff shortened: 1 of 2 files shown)


Logan007:
This looks extremely good to me.

Do you want to a small section to `doc/Building.md` about cross-compilation? Basically your comment from above with one example system / run?

hamishcoleman:
Documentation is always a good idea.  I've added that.

What did you mean by "example system / run"?  I think I have covered that with the arm-linux-gnueabi example I added, but wanted to be sure I understood you correctly.

Logan007:
Thank you very much!!

PR #851, opened 2021-10-14
2.8 stable

radaig:


.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+      matrix:
+        include:
+          - os: ubuntu-latest
+            TARGET: x86_64-unknown-linux-musl
+            COMPILER: gcc
+            LINKER: gcc
+          - os: ubuntu-latest
+            TARGET: armv7-unknown-linux-musleabihf
+            COMPILER: arm-linux-gnueabihf-gcc-5
+            LINKER: gcc-5-arm-linux-gnueabihf
```
hamishcoleman, line 23: Note that this doesnt actually build ARM binaries, and doesnt build with the musl libc either - so it would be worth avoiding creating release binary files that claim those attributes.

(And, yes, the cmake-linux.yml file in the dev branch should also be updated to remove these unusable parts)

Logan007:
As I am not too confident with build systems, could you please explain what exactly this pull request intends to do? Thank you!

And why is it for 2.8?

Logan007:
Hello @radaig 

I still am not getting exactly what your changes intend to do. Could you please explain?

Concerning the replacement of `pearson.c` and `pearson.h`, I have a notion that you want to fix a bug discussed over at [#785](https://github.com/ntop/n2n/issues/785#issuecomment-918892426). Please be advised that this bug has been fixed by #830 – it was not the `pearson.c/h` files that had a bug.

Logan007:
If required, please re-open with additional explanation. Thank you!

PR #852, opened 2021-10-14
Test on all available runner environments and add autogenerated crossbuilt dpkg packages

hamishcoleman:
I also have attempted to refactor the workflow to make it faster to complete - if github allows us to spin up enough VMs it /should/ be able to complete in about 6 minutes.

However, there is now quite a lot of builds happening, which makes it a little unwieldy and more likely to hit a VM shortage and stall for a while.  I ended up deciding to leave all these builds enabled for now, but it is something to keep considering.


Logan007:
Thank you!!

PR #853, opened 2021-10-15
Reduce the number of artifacts to a reasonable number

hamishcoleman:
Combine all the similar artefacts into the same named buckets to create less total artefacts.

This will also help in the future if we start using tagged releases to automatically publish artefacts as release assets


Logan007:
Thank you so much!!

This is becoming a Porsche build system :wink:

hamishcoleman:
ha!
Or at maybe a Hummer - Large, Power Hungry, Complex, but does everything ...

PR #854, opened 2021-10-15
Add some JSON interfaces to the edge management port

hamishcoleman:
This allows a couple of things:
- allows the removal of presentation logic from inside the edge binary
- machine readable queries would help to allow automated tests (eg: once there is a stable protocol, we could confirm that the built binaries can establish a connection to the public supernode within a specified timeframe)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -17,6 +17,7 @@ jobs:
           ./autogen.sh
           ./configure
           make test
+          make lint.python
 
   test_linux:
     needs: smoketest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,6 +12,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
+      - name: Install essential
+        run: |
+          sudo apt-get install flake8
       - name: Run minimal test set
         run: |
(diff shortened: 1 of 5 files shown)


Logan007:
What a wonderful idea!

As you brought up json, I would like to reveal an idea we were pondering for a while. We have actually been thinking of implementing a tiny http-server for management port output maybe also allowing to make some settings, such as a `reload_communities` button (instead of command to management port), or `verb+` or `verb-`...

Do you know if there is some simple java script delivered along with the json data in an html-doc and which can make the json data be displayed in a nice way?

Oh, and let me know when I can merge this PR.

hamishcoleman:
I imagine that having a small webserver would help a lot of people who cannot or do not want to use netcat :-)  - it does run the risk of adding feature bloat to the core.

Javascript is not my usual playground, but the simple python logic I have used in my str_table() should be easily portable to javascript (and there would be a zillion heavier-weight options floating around the net)

You can merge this PR

hamishcoleman:
@Logan007 As an alternative to embedding a http server in the edge process, I've added a simple http gateway server to this commit.

Logan007:
WOW! :open_mouth:

Thank you so much!

PR #855, opened 2021-10-15
Add a rendered html page to the example httpd server

hamishcoleman:
The httpd server now renders the data available from the json mgmt interfaces - making it a fully working demonstration

### DIFF
--- a/scripts/n2nhttpd
+++ b/scripts/n2nhttpd
@@ -3,10 +3,12 @@
 #
 # Simple http server to allow user control of n2n edge nodes
 #
-# Currently only for demonstration - needs javascript written to render the
-# results properly.
+# Currently only for demonstration
+# - needs nicer looking html written
+# - needs more json interfaces in edge
 #
 # Try it out with
+#   http://localhost:8080/
 #   http://localhost:8080/edge/peer
 #   http://localhost:8080/edge/super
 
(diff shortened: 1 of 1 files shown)


Logan007:
Thank you very much!

By the way, do you mind if I rename the python files to `.py` and the shell script to `.sh`?

hamishcoleman:
I envisioned the n2nctl script being a utility installed on the path - at which point an extension becomes annoying.  However, I would be interested to know what benefits you think come from renaming them.

Oh, I should tidy up the code lint branch a little and submit a PR - it has a little readme I wrote for the scripts dir.

Logan007:
> I would be interested to know what benefits you think come from renaming them.

Just as indicator for the sake clarity.

> I envisioned the n2nctl script being a utility installed on the path - at which point an extension becomes annoying. 

I agree. You convinced me not to rename.

> a little readme I wrote for the scripts dir

`doc/Scripts.md`?

hamishcoleman:
Clarity makes sense too - the `hack_fakeautoconf` script will never be an installed command (and the same for the `tools/test_harness` and the `indent` script from my lint branch)

You are right about the docs too - I'll write some more words and start a rename/cleanup/docs patch

Logan007:
Porsche :wink:

hamishcoleman:
With the (unplanned by me, so it didnt occur to me!) extension into semi-RESTful APIs via the management interface, I have been thinking about how to keep good REST semantics and now realise that to have a cleanly implemented and clear separation of GET (non-state-changing queries) and POST (eg a JSON equiv of +verb or stop) needs a protocol change :-(

So, I'll start adding infrastructure for that..

PR #856, opened 2021-10-17
Cleanup and Documentation for JSON management API

hamishcoleman:
Of note, there is the bare bones of an authentication system for the JSON management API in here - see src/edge_utils.c handleMgmtJson_auth() - which will need improving to be actually useful.  I'm thinking that this could be as simple as a new command line option to set the management password (instead of the current hardcoded "CHANGEME" password)

Since this ended up being a larger patch, I did not include any script renames in it.

### DIFF
--- a/doc/ManagementAPI.md
+++ b/doc/ManagementAPI.md
@@ -1,0 +1,175 @@
+# Management API
+
+This document is focused on the machine readable API interfaces. 
+
+Both the edge and the supernode provide a management interface UDP port.
+These interfaces have some documentation on their non machine readable
+commands in the respective daemon man pages.
+
+Default Ports:
+- UDP/5644 - edge
+- UDP/5645 - supernode
+
+## JSON Query interface
+
+As part of the management interface, A machine readable API exists for the
+edge daemon.  It takes a simple text request and replies with JSON formatted
+data.
+
+The request is in simple text so that the daemon does not need to include any
+complex parser.
+
+The replies are all in JSON so that the data is fully machine readable and
+the schema can be updated as needed - the burden is entirely on the client
+to handle different replies with different fields.  It is expected that
+any client software will be written in higher level programming languages
+where this flexibility is easy to provide.
+
+Since the API is over UDP, the replies are structured so that each part of
+the reply is clearly tagged as belonging to one request and there is a
+clear begin and end marker for the reply.  This is expected to support the
+future possibilities of pipelined and overlapping transactions as well as
+pub/sub asynchronous event channels.
+
+The replies will also handle some small amount of re-ordering of the
+packets, but that is not an specific goal of the protocol.
+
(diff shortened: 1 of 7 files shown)


Logan007:
!!!

Logan007:
I think we can leave the default password in there for now because the management port still is accessible in the very same way, bound to local host only. By the way, and this is just a thought, in the long-term, reading could require password-protection as well if we want to output the community name (which serves as key to header encryption).

However, I was thinking of maybe using the user-provided key as password (maybe not so good) or indeed, we can add yet another command line option, or, maybe enhance the `-t` option to optionally accept a password using some (not so) nifty syntax such as `-t [port][#<password>]`.

Let me know what you think. I could try to implement the cli password parsing somewhat later this week.

hamishcoleman:
Yeah, I certainly wrote it with the assumption that we could leave it hardcoded - it is no worse than the existing API.  But I didnt actually think of it until part way through, so I wanted to be sure to highlight it for discussion.

I've been trying to think of a clean and simple way to password protect /some/ read calls but not all .. and decided to punt that to the future too (the simplest one I thought of was changing the two main request types from 'read'/'write' access to 'read-only-unauth' and 'not-always-write-authed', but that feels hacky to me today)

I wouldnt worry about adding a cli parser yet - I would wait until it is clear that all the same functionality is available in both human readable and JSON-reply APIs and we are happy to deprecate the human readable functions.

Logan007:
Thank you very much – your contribution is deeply appreciated!

Logan007:
By the way, what do you think of moving the python scripts to the tools directory – and naming them `n2n-*` which is touchy for the `ctl`, I know...? But generally, I am under the impression that they have outlived their script status.

hamishcoleman:
I was actually thinking more the other direction!

I kind of had the idea that compiled code could go into one set of directories and interpreted code could go into another (tools and scripts, in this case) with the install step bringing the relevant useful scripts into a directory on the path.

Renaming them like that seems reasonable.  There is a bunch of shared code between the n2nctl and n2nhttpd, but extracting that out into a library just brings another code distribution issue.  I was personally thinking that the n2nhttpd was more experimental and less useful for installing, which might fit it into your 'script status' category.

Logan007:
Sounds like a very acceptable compromise :wink:

hamishcoleman:
I'll put together that renaming patch and see what you think.

Oh, and after reading your close notes on the 3.0 discussion - I dont see this API work as release critical.  Until and unless we start removing the old management commands, that is..

Logan007:
> I dont see this API work as release critical.

That's true on the one hand. But on the other hand, I think a release – especially a major one – should feel somewhat complete in terms of the available features. But take your time and let us know when you would call it a stable status.

> Until and unless we start removing the old management commands, that is..

Indeed. But this probably is more like a fast-to-make digital decision (`1`/`0`) which could happen anytime.

PR #859, opened 2021-10-18
Basic C Code lint checker and shell checker

hamishcoleman:
This also contains some of the script file renames that have been discussed (there are some others that would follow this, but are not as related to the lint steps in the rest of this patch)

The C Code lint changes include a small number of changes in places that I felt were unambiguously correct and files that 100% pass the linter have been added to a list of files to check on each run of the tests workflow.

This change might be too large to fit inside the definitions of no new features for the lead up to 3.0, but is theoretically all just whitespace and code correctness.  If we decide it is too large, the renames and docs changes could be cherry picked from here.


Logan007:
No objections at all, it does not add to core functions, just let me check... it might take a while as I am somewhat busy these days.

So, the linter did all the changes to the code? At first look, I saw some seemingly inconsistent changes, such as `auth.h` having the white space at function forward declaration and `include/edge_utils_win32.h` not showing it. I might be mistaken though.

The (de-)linter will just flag the files checked? Or will it correct these files?

hamishcoleman:
The scripts/indent.sh script has the option to overwrite the file with its suggested changes, however the default mode is to make no changes, simply print the diff and exit with an errorcode - basically flagging them, yes.

If there is inconsistency, I may have misunderstood the style rules - but looking at those two headers, it is clear that there is something up.  The linter is force removing the space from /some/ of the forward definitions in auth.h but force adding it to some other ones!

hamishcoleman:
I've (manually) looked through the lint changes here and I believe I have dropped the style-inconsistent updates.

Logan007:
Thank you!!

PR #861, opened 2021-10-19
JSON Reply Management API - feature parity with old management interfaces

hamishcoleman:
I believe that this is now implementing all the features of the current human readable interface and that the functionality and design are proven to work.


Logan007:
Thank you very much for this outstanding contribution!

Logan007:
Let me add as a post script that the `html` page is awesome – including the buttons and the auth!!

PR #869, opened 2021-10-24
updated JSON API  password handling

Logan007:
This PR intends to change JSON API password handling in several increments.


Logan007:
Next step would be to allow the password being set through CLI. Any suggestions for a CLI parameter name? Maybe `--management-password`?

Please note that the default password has been changed to `n2n`. The documentation will be adapted accordingly in the next step, too.

PR #870, opened 2021-10-24
Fix support for python versions prior to 3.6

hamishcoleman:
which do not support a context manager in the socketserver


Logan007:
Thank you!

PR #876, opened 2021-10-25
Fix python JSON clients not to hang if the daemon does not respond

hamishcoleman:
Basically, add timeouts


Logan007:
Thank you!

PR #877, opened 2021-10-27
UPnP port redirection is supported

fengdaolong:
For the core function of n2n: `p2p`, upnp may be the most useful improvement in recent years.
Through our attempts, we found that it significantly improves the success rate of p2p.


fengdaolong:
@hamishcoleman The automatic construction and automatic checking of the code seems to be a poor support for `git submodules`, please help!

hamishcoleman:
The default `actions/checkout` does not get submodules, so adding a:
```
 with:
      submodules: true
```
to every checkout is needed.

However, the smoketest error also shows that you are compiling some of this new code even when `HAVE_MINIUPNP` is no defined.

fengdaolong:
> However, the smoketest error also shows that you are compiling some of this new code even when HAVE_MINIUPNP is no defined.

Will the make script automatically compile all source files in the `src` directory into targets?

hamishcoleman:
Yes, unfortunately that is how the current makefile is setup

fengdaolong:
The make script compiles all the source code in src into targets, which I think is not particularly reasonable.

hamishcoleman:
Yes, it is not reasonable.  However until it gets rewritten, that is how it works

fengdaolong:
If we need to change this unreasonableness, can you help?

hamishcoleman:
Fixing up all the makefiles is an ongoing task - one that I hope to be able to continue doing.  However, I am avoiding making any major changes while we are in a feature freeze.

Logan007:
@fengdaolong, this is great!! I can hear people applaud.

On that build related discussion, a more general question comes into my mind: Shouldn't we just integrate the code 'physically' (as files) into our code base just as we did with minilzo? Just to be consistent and also take pre-cautionary measures against miniupnp changing / deleting / .... their code.

hamishcoleman:
Using a submodule means that you point at a specific version of the upstream code - so as long as their repo does not get deleted, we will always see the version we expect.  That being said, we would probably be served by pointing at a stable release commit - and we should consider just using the packaged library when compiling on Ubuntu/Debian

fengdaolong:
>Shouldn't we just integrate the code 'physically' (as files) into our code base just as we did with minilzo?

Choosing to link an existing binary library or choose to compile with source code is a multiple-choice question, and different people have different answers. Just like when using the n2n tool, some people like to use the compiled binary file, while some people like to compile the source code into a binary by themselves.

>miniupnp changing / deleting / .... their code.

No, git submodule has recorded the commit ID of miniupnp, which can be checked with the `git submodule status --recursive` or `git ls-tree -r --long HEAD | awk -F " " '$2 ~ /commit/ {printf"%s\n", $0}'`. After the upstream of miniupnp is changed, the local can still switch to the expected commit. 

>Using a submodule means that you point at a specific version of the upstream code - so as long as their repo does not get deleted, we will always see the version we expect. 

What you said is right, git submodule can guarantee the exact version of the library it depends on.

>we should consider just using the packaged library when compiling on Ubuntu/Debian

However, miniupnp has different binary version numbers in different Linux distribution repositories, and different versions have different APIs, which will cause compatibility problems.

fengdaolong (commented):
I personally think that adding a sentence like "#ifdef HAVE_MINIUPNP ..." to each function makes the code not elegant enough and looks messy. If you want to add other sub-modules in the future, you must do it all in this inelegant way, it will become more and more inconvenient to expand in the future.
It should be reasonable to modify the make script to the correct compilation method, don't let make compile the source code in the entire directory without thinking.

Logan007:
I see two ways doing it, there certainly are more than that and I am just not aware... :wink:

I just tried to be consistent with the N2N_HAVE_PTHREAD implementation and minimize the number of `ifdefs` in the _main_ code.

Any other way would be fine as well. It was just the most natural way appearing to me at that moment.

Whatever way you chose, and please change it accordingly, I will be fine with it as long as it compiles with the `make` build system. With the current changes, it seems to properly build without miniupnp. Next aim before merging should be to also get it built with miniupnp using the `make` build system. Any ideas?

CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ if(N2N_OPTION_USE_ZSTD)
 endif(N2N_OPTION_USE_ZSTD)
 
 if(N2N_OPTION_USE_MINIUPNP)
-  ADD_DEFINITIONS("-DHAVE_MINIUPNP")
+  ADD_DEFINITIONS("-DN2N_HAVE_MINIUPNP")
```
hamishcoleman: The syntax of `HAVE_xyz` is pretty common for this kind of definition.
Prefixing it with a namespace is not required if we are careful in how it is used in the public include files that will be used when embedding.

Logan007: D'accord. We will have to change this for the other feature-related defines as well sometime soon though, just to get the code more consistent in this regard then.

hamishcoleman: I like consistency

Logan007: For now, I propose to leave it as it as. I will address it in another pull request sometime later for all those defines once this one will have been merged.

fengdaolong:
>Next aim before merging should be to also get it built with miniupnp using the make build system. Any ideas?

Yes, it's just that I am not very familiar with make scripts, so I need help from others.
I think the current task should first modify the make script so that it compiles without errors.

hamishcoleman:
Adding a substantial external module like this has significant costs. (Even just using their existing makefiles is not straightforward because they have different makefiles for Windows compilation targets)

There is also a question here about why there are two build systems - and which one should be the one required to be supported

fengdaolong:
>Adding a substantial external module like this has significant costs. (Even just using their existing makefiles is not straightforward because they have different makefiles for Windows compilation targets)

After my test, in my linux environment, using cmake script to compile windows version under x86_64-w64-mingw32-gcc environment is no problem at all.
I have been using cmake to compile versions for various platforms.


>There is also a question here about why there are two build systems

This... is a problem.😉
I always think it's best to keep only one way, and maintaining two build scripts at the same time will result in too much maintenance cost. I prefer cmake not because of how good it is, but because it is more convenient for cross-platform compilation. But the two construction methods have their own advantages, and everyone has their own preferences.

>and which one should be the one required to be supported

No one can make a decision, it depends on the turnout.

hamishcoleman:
@fengdaolong have you seen that the makefiles can do some of that cross compilation now?

fengdaolong:
>have you seen that the makefiles can do some of that cross compilation now?

I know, essentially, a written cmake script will be translated into a make script, maybe I think cmake is easy to write, so I got used to it.
If only one compilation script is kept, and more people vote for make than cmake, I can still choose to learn make.

hamishcoleman:
When I talk about the costs of importing an external module, I am referring to several aspects:
- adding this code pretty much doubles the size of the codebase
- Not using OS packaged versions of this library means that we need to be responsible for security and bugfix updates
- Using a submodule means that we still need to be clearly aware of the upstream release cycle and any fixes they apply
- Including the code directly is essentially a fork and runs the risk that we will diverge from upstream, as well as us needing to be just as aware of as when using the submodule

I can think of several other costs, but these are just the import and maintenance concerns

Logan007:
We still are somewhat far from 3.2 so we have some time left to think about _how_ to integrate **third party code**. On the one hand, I see the cost. On the other hand, I see the benefit of the alternatives. We will need to find a way once someone wants to integrate it as always-available run-time option at the latest.

Concerning the **build system**, I see that CMake might make things easier but it is not as capable as plain Makefiles, especially for cross-compilation which a lot of users do need for several platforms. So, for now, I cannot help but to see it as an unavoidable **requirement** to support both build systems. Note that this only will be relevant for big changes such as this one including third party code or alike.

fengdaolong:
I think that the existing components (pcap, zstd, openssl) can be integrated and tracked source code using git submodules, but the prerequisite is that the scripts of the compilation system need to be well-written.

Logan007:
> existing components (pcap, zstd, openssl) can be integrated and tracked source code using git submodules

I fear that compilation will grow to an unbearable extent then. Also, the binaries might partially duplicate what is intended to be found in libraries. Probably a less preferred option for smaller systems such as routers suffering from more limits (storage, RAM, ...).

Maybe, the way to go is making `miniupnp` an optional library just like the others (`zstd`, `openssl`, ...)? We could rely on the system providing a stable and recent version then.

fengdaolong:
>I fear that compilation will grow to an unbearable extent then. 

You can set options to choose whether to link the existing libraries in the system or compile by yourself when compiling.

>We could rely on the system providing a stable and recent version then.

As mentioned earlier, different operating systems provide different versions, different versions, and different interfaces. Therefore, the compatibility of various versions must be considered.

fengdaolong:
@Logan007  Friendly reminder: the original `upnp` in the code has been changed to `port mapping` because it is not limited to the upnp protocol.

Logan007:
I had to re-adjust the `ifdef`s to the previous version for the thread-related code compiles well as it calls the `set`/`del` functions.

fengdaolong:
`n2n_upnpc.c` no longer exists, it has been renamed to `n2n_port_mapping.c`, please do not push this file again.

Logan007:
So, the way forward would be to merge this pull request sometime soon even though it still causes a two second delay stalling the edge at each supernode change, i.e. due to load balancing. This pre-mature merge would happen for better access to testing. A follow-up step would include to fill the still visible cavities for proper threadification (communication-related, through management port).

**But before we do all that, I would like to get some feedback from all of you on how this implementation is helpful to p2p. So, please, check how [this current code](https://github.com/fengdaolong/n2n/archive/upnp.zip) works for you and share some feedback here.**

fengdaolong (commented):
How to revoke a git commit operation error?

fengdaolong:
> How to revoke a git commit operation error?

I checked and found that there was no error in the accidental merge operation, and it does not need to be undone. It makes sense to merge the dev branch into the upnp branch.

hamishcoleman:
You can easily revert an accidental commit.
- First, step your local repo backwards: `git reset HEAD^`
- Optionally, ensure the files in your checkout match the commit: `git reset --hard`
- Push your branch: `git push --force` (the `--force` option is because the history of your local and the remote branch has now diverged)
-

lucktu:
I recommend copying “upnp(or others)" into a special directory to keep n2n stable and easy to use.

I chose to keep "CMakeLists.txt" and remove "Makefile.in" for cross-compilation and reduce errors.

Logan007:
> I recommend copying “upnp(or others)" into a special directory to keep n2n stable and easy to use.

We are still working on it. As far as I can see, it already can open a router's port through UPnP quite well. From my point of view, there are only two things left that need to be addressed before merging for now.

- The current implementation blocks the whole edge for 2 seconds at the beginning, at every supernode change, and at the end. It can be avoided by multi-threading which is still "under construction"
- ~~Not everyone is happy with using this feature. So, I want an easy way to disable it, perhaps at CLI, or to simply compile completely without it.~~ _taken care of by 6bb625f_

> I chose to keep "CMakeLists.txt" and remove "Makefile.in" for cross-compilation and reduce errors.

The makefile has been changed to use dynamic libraries `libupnpc`and `natpmp` if present.

By the way, I observed that the makey way results in way smaller executables compared to CMake, ~ 240K vs. ~700K on Linux.

Logan007:
So, is this ready to be merged? Please test and let me know.

src/edge_utils.c
```diff
@@ -347,12 +347,6 @@ void supernode_disconnect (n2n_edge_t *eee) {
     if(eee->sock >= 0) {
         closesocket(eee->sock);
         eee->sock = -1;
-
```
fengdaolong: Please note that when exiting the edge, don't forget to clean up the port mapping.

Logan007:
I had to manually intervene to resolve some conflicts, so #905 merges it.

Logan007:
@fengdaolong thank you very much!

PR #878, opened 2021-10-27
Fix build system errors

hamishcoleman:
The github runners have an inconsistent package database, so this pull adds the needed `apt-get update` lines.
It also includes some macos cross compile builder additions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -371,7 +371,9 @@ jobs:
           export SDKROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
           ./autogen.sh
           export CC=clang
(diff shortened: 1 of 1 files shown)


Logan007:
Thank you!!

hamishcoleman:
We had unfortunate timing with the release - the runners started breaking some time in the last 6 hours.

hamishcoleman:
Also, the automatic release on tag automation will not re-run unless a new tag is made :-(

Logan007:
That's life ;)

As soon as we have started adding a new features again, UPnP is a hot candidate, we can try the pre-releases again.

hamishcoleman:
An alternative is to move this tag to the current stable commit (yes, reusing tags is bad, but in this case, not enough time has past for many people to have cloned the old tag)

PR #879, opened 2021-10-27
Fix build system errors

hamishcoleman:
The github runners have an inconsistent package database, so this pull adds the needed apt-get update lines.
It also includes some macos cross compile builder additions.


Logan007:
Thank you!!

PR #888, opened 2021-11-01
assured correct supernode MAC address properties

Logan007:


hamishcoleman:
Good idea.

Wont this just be surprising to those who configure their own addresses, and keep getting a different answer to the one they configured?

Logan007:
`-m` at supernode has been enabled today so I do not expect too many users already having had used it before – especially as we did not advertise the related compile-time switch. Also, we have this filter at the edges already, more "hidden" as part of the tun_tap configuration.

Logan007:
> Good idea.

Yeah, actually it is [yours](https://github.com/ntop/n2n/commit/89571292f4651d35e110089f7ee8c802efc96e8f#commitcomment-59108743)... Credits!!

hamishcoleman:
Haha, nah - I was thinking of documenting or erroring, it never occured to me to simply mask the bits in/out

PR #889, opened 2021-11-01
Revamp version number handling

hamishcoleman:
This takes the existance of the annotated tags and feeds that into the standard `git describe` program to generate version numbers.  There is also a fallback version number for builds done outside of a git repo - which is now all unified into the VERSION file (for both CMake and Makefiles)

This has allowed a bunch of autoconf cleanups as well.  The place that the version numbers are calculated is slowly moving closer to the compile-time (with the intent that eventually you will not need to re-run configure to get a new version number in the build)


fengdaolong:
`scripts/version.sh` is not applicable to windows, how to get `git describe` for windows?

hamishcoleman:
I think you mean that `scripts/version.sh` is not applicable to CMake with no MSYS or Cygwin available.

We could just fall back to using the short version if the bash script cannot run.

fengdaolong:
> We could just fall back to using the short version if the bash script cannot run.

Or, write another batch "bat" file for windows?

hamishcoleman:
> Or, write another batch "bat" file for windows?

That is possible, but both complexifies the calling process and requires two scripts to be kept in sync, so I would want to avoid it

hamishcoleman:
@fengdaolong I have added an alternate path for the CMake builds

PR #899, opened 2021-11-16
Update Openwrt Makefile

dwj0:


hamishcoleman:
Thanks!

I am actually in the process of preparing a more comprehensive overhaul of the included openwrt recipe, are you able to test the process and configuration in https://github.com/hamishcoleman/n2n/tree/openwrt ?

PR #906, opened 2021-12-23
made github cmake runner fetch submodules

Logan007:


hamishcoleman:
When we want to re-add this, It is probably better to use the actions/checkout do this with `submodules: true`  (Trying to use as much 'standard' tooling as works)

Logan007:
Manually added it. I guess it will be working starting from next PR?

hamishcoleman:
It ran as soon as you committed - I can see a couple of extra lines in the checkout log.  The cmake build is still broken for unrelated reasons (and the linter disagrees with your yaml indentation :-P)

Issue #908, opened 2021-12-24
compile error on linux


skyformat99:
1) git clone latest code
2) cmake  &  make ,   The following compilation error occurs:

In file included from /var/zz/n2ngit2/n2n/include/n2n.h:162:0,
                 from /var/zz/n2ngit2/n2n/src/n2n.c:19:
/var/zz/n2ngit2/n2n/include/n2n_port_mapping.h:7:33: fatal error: miniupnpc/miniupnpc.h: No such file or directory
 #include <miniupnpc/miniupnpc.h>
                                 ^
compilation terminated.

PR #909, opened 2021-12-24
Update n2n_port_mapping.h

skyformat99:
fix  compile error on linux #908


Logan007:
@skyformat99 Thank you very much for pointing this out! Based on your idea, #912 uses a slightly more general approach which also keeps open the regular makey way of building.

PR #911, opened 2021-12-24
Update README.md

skyformat99:
update readme.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,17 +52,6 @@ Now the supernode service should be up and running on port 1234. On your edge no
 
 
 ## Manual Compilation
-First of all, you may clone the source codes from github:
-```bash
-git clone https://github.com/ntop/n2n.git
-cd n2n
-```
-
-The libnatpmp & miniupnp have been moved to separated repositories,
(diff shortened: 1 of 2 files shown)


hamishcoleman:
There is already some documentation in the doc/Building.md file - rather than making the main README.md bigger, perhaps we should remove the duplicate build details from the README file and continue to point people at the Building.md (possibly improving the Building file if needed)

Logan007:
@hamishcoleman I agree. Also, I still prefer the makey way as default. We could add `CMake`to the enumeration of alternatives:

For Windows, MacOS, **CMake** and general building options, please check out [Building documentation](doc/Building.md) for compilation and running.

skyformat99:
> @hamishcoleman I agree. Also, I still prefer the makey way as default. We could add `CMake`to the enumeration of alternatives:
> 
> For Windows, MacOS, **CMake** and general building options, please check out [Building documentation](doc/Building.md) for compilation and running.

ok, I agree.
close it.

Logan007:
No need to close, we can adapt it.

PR #921, opened 2022-01-02
3.0 stable - unnecessary patch for OpenWRT

siskai:


Logan007:
I am not sure why this would point to another repository. Maybe some residual?

@hamishcoleman what do you think?

hamishcoleman:
Thanks for reminding me - I made a similar fix in dev, but forgot that it was also affecting the stable branch.

We definitely want to remove the patch to let it actually build, and fix the version number.  

It might also be a good moment to look at the other changes in the packages/openwrt/Makefile from the dev branch to see if any are applicable for a stable branch.

@siskai, For the simple option, you can split this into two PRs - the first with the patch file remove - then once that is merged you can reference a PKG_SOURCE_VERSION that is known to exist in the stable branch in a second PR (with the right PKG_SOURCE_URL).  However, if you wanted,  you could also look at the openwrt build changes in dev and back-port a stable-suitable version of that.

siskai:
My fault, url should refer to the original repository.

I'll fix it.

Issue #964, opened 2022-03-05
Federation bug report


ghcruise:
Background : 2 supernodes formed federation , 2 edges connected
When one of supernodes offline, edges can't work with "[edge_utils.c:1541] WARNING: supernode not responding, now trying ["The offline supernode IP adress"]"

Logan007:
Thank you for reporting. It sounds a bit like #923.

Can you describe your scenario, your command line parameters at edge and supernode, how you setup Federation and maybe share parts of the management port output?

jradxl:
I am finding the same...
I am not convinced that all the Edges recover to the active Supernodes, nor recover when the inactive Supernode returns.
(without restarting the Edge: If I was going to use right now as my only connectivity I'd put each Edge on a 24h Cron!)
I note your documentation says it might take some time, and I have waited!!!

I am trying with 3 supernodes. The config file is similar, except that the -l option does not address itself (pity as that makes the config file different on each supernode). [If I do self define the supernode with -l, then system jams up!]

One of the Supernode.Conf:-
```
-F=jsrsoftcouk
#-a=192.168.195.0-192.168.195.0/24
-p=6432
-c=/etc/n2n/communities.list
-l=XXX.89.148.78:6432
#-l=YYY.ZZZ.63.119:6432
-l=YYY.ZZZ.78.139:6432
```

Communities.List:-
```
jsrsoft 192.168.195.0/24
* z390ma <secret key>
* vps67254500 <secret key>
* xcase1 <secret key>
* vps-may <secret key>
* ubuntues2 <secret key>
* vps-june <secret key>
```
There are 6 edge nodes....
Each Supernode server has an Edge
One Edge is in Spain, behind triple NAT
Two Edges are in UK, behind quadruple NAT
So the far-edges can only make outgoing calls.
So it's as if the Supernode Servers are Jump/Bastian hosts
Thus this is a site-to-site configuration.

A typical Edge.Conf is:-
```
# cat edge.conf 
-d=n2n0
-c=jsrsoft
-k=<secret>
-m=62:DA:23:E0:C4:A2
-e=auto
-a=192.168.195.2/24
-p=5432
-H
-P=<secret>
-A5
-I=vps-may
-J=<secret>
-l=XXX.89.148.78:6432
-l=YYY.ZZZ.63.119:6432
-l=YYY.ZZZ.78.139:6432
```
```
# echo | nc -w1 -u 127.0.0.1 5645
[...]

PR #970, opened 2022-03-16
made supernode ignore its own registration attempts …

Logan007:
…to allow unified configuration of federated supernodes (command line-provided supernodes can now include self) as discussed [here](https://github.com/ntop/n2n/issues/964#issuecomment-1069592317)

### DIFF
--- a/src/sn_utils.c
+++ b/src/sn_utils.c
@@ -1938,6 +1938,11 @@ static int process_udp (n2n_sn_t * sss,
                }
             }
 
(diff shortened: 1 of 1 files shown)


hamishcoleman:
Would this make sense to back port into the 3.0 branch?  It is small and self-contained and could help simplify the new user experience.

Logan007:
Oh yes, I will add it to the `3.0-stable` branch.

Issue #991, opened 2022-05-07
coredump for the latest version on openwrt


galaxyskyknight:
it happened after a mwan3 re-apply and firewall reload. somehow the edge exit and core dumped. should not be like this way.

```
Sun May  8 01:58:49 2022 kern.info kernel: [ 1229.539617] traps: edge[327] trap stack segment ip:41a686 sp:7ffe09873030 error:0 in edge[403000+80000]
Sun May  8 01:58:49 2022 daemon.info avahi-daemon[6009]: Interface n2n.IPv6 no longer relevant for mDNS.
Sun May  8 01:58:49 2022 daemon.info avahi-daemon[6009]: Leaving mDNS multicast group on interface n2n.IPv6 with address fe80::290:10ff:fe00:2.
Sun May  8 01:58:49 2022 daemon.info : 12[KNL] interface n2n deactivated
Sun May  8 01:58:49 2022 daemon.info : 14[KNL] fe80::290:10ff:fe00:2 disappeared from n2n
Sun May  8 01:58:49 2022 daemon.info : 10[KNL] 10.0.0.2 disappeared from n2n
Sun May  8 01:58:49 2022 daemon.info : 12[KNL] interface n2n deleted
```

galaxyskyknight:
Also I see the following log:
`Sun May  8 02:07:13 2022 daemon.info n2n[546]: WARNING: sendto(x.x.x.x:7681) failed (1) Operation not permitted`

seems like the firewall try to deny the n2n accessing the peer. but anyway, the n2n code should not running into core dump. there must be some exception in socket or I/O handling I guess, in this special case.

there are 3 addtional core dump traps from kernel:

```
[ 1229.539617] traps: edge[327] trap stack segment ip:41a686 sp:7ffe09873030 error:0 in edge[403000+80000]
[ 1733.745667] traps: edge[546] trap stack segment ip:41a686 sp:7ffdffec47b0 error:0 in edge[403000+80000]
[ 1884.042965] traps: edge[13350] trap stack segment ip:41a686 sp:7ffc20df6d10 error:0 in edge[403000+80000]
```

hamishcoleman:
You can see from the trap errors that the same IP is causing the trap each time.  Without a symbol table from your compiled binary, we cannot tell where that IP is in the code.  Are you able to run the edge with gdb and capture a backtrace?  Is your build process automated and documented anywhere?  Do you have the core dump file that was generated?

[...]

PR #997, opened 2022-05-17
Fixed coredump

small-5:
Fixed https://github.com/ntop/n2n/issues/991

### DIFF
--- a/src/edge_utils.c
+++ b/src/edge_utils.c
@@ -1061,13 +1061,14 @@ static ssize_t sendto_fd (n2n_edge_t *eee, const void *buf,
             traceEvent(level, "WSAGetLastError(): %u", WSAGetLastError());
 #endif
         }
+    } else {
+        /* we reach here, either because !check_sock_ready() or (errno) */
+        supernode_disconnect(eee);
+        eee->sn_wait = 1;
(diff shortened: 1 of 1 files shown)


hamishcoleman:
Thanks!

Can you share how you found the fault?  I would like to have a reproducible test case to try.

Logan007:
Hey @small-5 , good to see you still around!

Your PR would prevent the call to `supernode_disconnect()`  in case of error while sending (which probably was the case before the core dump). So, I am not sure if I understand the idea behind it (but I would really like to). Could you provide some more details, please?

small-5:
Hello,I have tried that if  the wan  is down, and then start edge, "failed to bind to local multicast group" will be displayed, and coredump will occur. If the wan  is  up  and have ip, start edge and "successfully joined multicast group" will be displayed. At this time, coredump will not occur.
@Logan007 nice to see you again.
By comparing https://github.com/ntop/n2n/commit/009311d016bf27f40259e6bb992ce4a78af24424, I found that there is a return sent; at the end of this part(before https://github.com/ntop/n2n/commit/009311d016bf27f40259e6bb992ce4a78af24424,this part return sent or return -1.), so after adding it, it will return to normal.
However, I found that there is a new version(a274818854c01008a4106a44fd5ecd33d14091a4). I'll test it to see if it returns to normal.
Thanks!

PR #999, opened 2022-05-18
Fixed coredump

small-5:
Fix https://github.com/ntop/n2n/issues/991


hamishcoleman:
Are you able to explain how this fixes the coredumps?  None of the code you have touched is dereferencing any pointers that could lead to a coredump.  Also, a zero return value is a valid result for the sendto() function, so that seems wrong.

Are you actually able to replicate this coredump?  can you run diagnostics to work out where the coredump is occuring?  It is somewhere upstream of these functions and would be really good to know where so we can fix the cause.

PR #1001, opened 2022-05-19
Sendto refactor

hamishcoleman:
More searching in the darkness for ghostly bugs


small-5:
This PR solves the problem. When wan is down, there is no coredump.Thanks

PR #1012, opened 2022-06-11
Modify the scan_address function

dwj0:
Modify the scan_address function, when the IP address no slash present, it is possible to make the bitlen error.


Logan007:
Thank you!

Issue #1015, opened 2022-06-15
Supernode how to lisen/bind special ip address like localhost


Eloco:
```
Welcome to n2n v.3.0.0.r1038.d2683f2 for Debian bullseye/sid

❯ sudo supernode -h | grep host
 options for under-        [-l <supernode host:port>]

❯ sudo supernode -l localhost:7777
❯ sudo lsof -i -n | egrep 'COMMAND|LISTEN|UDP|TCP' | grep super ;sudo pkill -f super
supernode 80291             n2n    4u  IPv4 356559      0t0  UDP *:7654
supernode 80291             n2n    5u  IPv4 356560      0t0  TCP *:7654 (LISTEN)
supernode 80291             n2n    6u  IPv4 356561      0t0  UDP 127.0.0.1:5645
```

i want to bind the address to localhost or other typical ip, but `-l` param not work

Logan007:
Hello,

While egde supports an enhanced `-p [<ip address>:]<port>` syntax with optional IP address of some local interface to specify interface **and** port to bind to, the supernode only supports choice of port, i.e. `-p <port>` so far. But I think it would make a nice and useful enhancement porting this from edge over to supernode as well. I will add it to the list, thanks for this idea!

By the way, `-l` is not for **l**isten, it is for **l**ink to other supernodes and provides addresses of other supernodes to connect to, see `doc/Federation.md`.

Logan007:
Added with #1016.

PR #1016, opened 2022-06-16
added optional choice of bind address (-p) to supernode

Logan007:
This PR ports optional bind address (default 'any') from edge to supernode, e.g. `-p 127.0.0.1:7777` where the IP address is optional such that `-p 7777` still works as well.

Solves #1015


happyntec:
Good work;

This is a very useful enhancement for who run supernode in docker container or docker swarm or cloud network with serveral ethernet interfaces; 

I've made serveral supernodes runned for a  federation by docker swarm network, and each supernode just didn't get external ip, it's a big problem; Now i can bind special external ip address for each internal supernode;

Thank you. I'd glad to do testing for this enhancement;

PR #1023, opened 2022-06-24
added Windows support to n2n-route tool

### LINKED ISSUE #1017: why abandom -r parameters but not support n2n-router tools on windows??what should I do to add router on windows? Raw command??
Hello

I have upgraded to latest code and compiled it on windows and now I find the n2n binary has abandomed the '-r' parameter to allow me adding the static routes on windows any more, and it prompts I can use 'n2n-router' tools to do that, however, what makes me very disappointed is when I run this n2n-router tools, it says  only support on Linux?  I am very puzzled how it could be a feature so easily removed and make a very important change to break some basic function like adding routes for a peer 2 peer vpn software but non plan B for user's choise? **This is totally destroyed my senario of usage on windows platform: run the n2n binary and specificy all static routes via '-r' option to all edge peers‘ subnets run on different sites via a '.bat' batch script in one shot on a windows cmd admin terminal.**

In one word, **I am very suprised the point of your consideration and decesion made on legacy features looks like taking none  priority of its continuity and os platform compatibility/continuity which to me is hard to be understandable....**  I understand to reduce the application size and ligth weigh it and get more focus on the P2P core function itself  is proper way and good, but it should not stand at cost of breaking any current working function or OS compatibility, right? Or you should migrate it  after there is an alternative solution but not deploy directly by breaking or deprecate other functions/OS compatiblity etc w/o any fully covered solution or you totally count on the common user to find out the way in source code... After all, not everyone has the skillset to dig it out from the source code.

image

image

Logan007:
This PR intends to add Windows support to the `n2n-route` tool, development as we go along.

Fixes #1017

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,7 +206,7 @@ add_library(n2n STATIC
 if(DEFINED WIN32)
   add_library(edge_utils_win32 src/edge_utils_win32.c)
   add_subdirectory(win32)
-  target_link_libraries(n2n edge_utils_win32 n2n_win32 iphlpapi)
+  target_link_libraries(n2n edge_utils_win32 n2n_win32 iphlpapi netapi32)
 endif(DEFINED WIN32)
 
 add_executable(edge src/edge.c)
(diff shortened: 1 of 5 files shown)


Logan007:
Any volunteers to test this on Windows?

GreatMichaelLee:
> Any volunteers to test this on Windows?

will try it on next Monday, thanks

Logan007:
Merging in anticipation of more feedback.

PR #1036, opened 2022-07-07
build(CMakeLists): dynamic link to libcap by default

1715173329:
This is a only exception while all other libraries are dynamically linked.

The "TODO" does not make any sense as libcap provides both dynamic and static libraries.
There is not reason to explain why it needs to be statically linked 'especially' as dynamic linking works very well.
Then just make it dynamically-linked by default, which also matches the behaviour on autotools.

Compiled and tested on aarch64, no errors found.


hamishcoleman:
I probably could have been more verbose when I wrote that TODO, but I was copying from an earlier file which specifically linked this against the static version and intended it to be a reminder to revisit, which you have done!

Thanks!

PR #1037, opened 2022-07-07
CI(openwrt): refactor build and test process

1715173329:
This refactored the whole build & test process.

Cleaned up the Makefile, and switched to cmake build.

For CI, switched to OpenWrt's official build process, and build ipk for popular architectures.
Also implemented runtime test for armv7 / aarch64 and i386 / amd64.

### DIFF
--- a/.github/openwrt/Dockerfile
+++ b/.github/openwrt/Dockerfile
@@ -1,0 +1,6 @@
+ARG ARCH=x86-64
+FROM openwrt/rootfs:$ARCH
+
+ADD entrypoint.sh /entrypoint.sh
+
+CMD ["/entrypoint.sh"]
--- a/.github/openwrt/entrypoint.sh
+++ b/.github/openwrt/entrypoint.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+. /etc/openwrt_release
+
+mkdir -p /var/lock/
+
+opkg update
+
+for PKG in "/ci/ipk-$DISTRIB_ARCH"/*.ipk; do
+	tar -xzOf "$PKG" ./control.tar.gz | tar xzf - ./control
+	# package name including variant
+	PKG_NAME=$(sed -ne 's#^Package: \(.*\)$#\1#p' ./control)
+	# package version without release
+	PKG_VERSION=$(sed -ne 's#^Version: \(.*\)-[0-9]*$#\1#p' ./control)
+	# package source contianing test.sh script
+	PKG_SOURCE=$(sed -ne 's#^Source: .*/\(.*\)$#\1#p' ./control)
+
+	echo "Testing package $PKG_NAME in version $PKG_VERSION from $PKG_SOURCE"
+
+	opkg install "$PKG"
+
+	export PKG_NAME PKG_VERSION
+
+	TEST_SCRIPT="/ci/.github/openwrt/test.sh"
+
+	if [ -f "$TEST_SCRIPT" ]; then
+		echo "Use package specific test.sh"
+		if sh "$TEST_SCRIPT" "$PKG_NAME" "$PKG_VERSION"; then
+			echo "Test successful"
+		else
+			echo "Test failed"
+			exit 1
+		fi
+	else
+		echo "No test.sh script available"
+	fi
+
+	opkg remove "$PKG_NAME" --force-removal-of-dependent-packages --force-remove
+done
--- a/.github/openwrt/test.sh
+++ b/.github/openwrt/test.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+
+case "$1" in
+	"n2n-edge"|"n2n-supernode")
+		eval "${1#n2n-}" | grep "$2"
+		;;
+	"n2n-tests")
+		for test in $(grep -v '#' '/usr/bin/n2n-tests_units.list'); do
+			eval "n2n-$test"
+		done
+		;;
+esac
--- a/.github/workflows/openwrt.yml
+++ b/.github/workflows/openwrt.yml
@@ -17,33 +17,43 @@ jobs:
         include:
           - arch: arc_archs
             target: archs38-generic
+            runtime_test: false
 
           - arch: arm_cortex-a9_vfpv3-d16
             target: mvebu-cortexa9
+            runtime_test: false
 
           - arch: mips_24kc
             target: ath79-generic
+            runtime_test: false
 
           - arch: mipsel_24kc
             target: mt7621
+            runtime_test: false
 
           - arch: powerpc_464fp
             target: apm821xx-nand
+            runtime_test: false
 
           - arch: powerpc_8540
             target: mpc85xx-p1010
+            runtime_test: false
 
           - arch: aarch64_cortex-a53
             target: mvebu-cortexa53
+            runtime_test: true
 
           - arch: arm_cortex-a15_neon-vfpv4
             target: armvirt-32
+            runtime_test: true
 
           - arch: i386_pentium-mmx
             target: x86-geode
+            runtime_test: true
 
           - arch: x86_64
             target: x86-64
+            runtime_test: true
 
     steps:
       - uses: actions/checkout@v3
@@ -79,3 +89,19 @@ jobs:
 
       - name: Remove logs
         run: sudo rm -rf logs-${{ matrix.arch }}/ || true
+
+      - name: Register QEMU
+        if: ${{ matrix.runtime_test }}
+        run: sudo docker run --rm --privileged aptman/qus -s -- -p
+
+      - name: Build Docker container
+        if: ${{ matrix.runtime_test }}
+        run: |
+          git checkout HEAD .github/openwrt
+          docker build -t test-container --build-arg ARCH .github/openwrt/
+        env:
+          ARCH: ${{ matrix.arch }}
+
+      - name: Test via Docker container
+        if: ${{ matrix.runtime_test }}
+        run: docker run --rm -v $GITHUB_WORKSPACE:/ci test-container
--- a/.github/workflows/openwrt.yml
+++ b/.github/workflows/openwrt.yml
@@ -1,112 +1,81 @@
 ---
-name: Openwrt Build
+name: OpenWrt Build
 
 # yamllint disable-line rule:truthy
 on:
-  release:
-    types:
-      - published
-      - created
-      - edited
   push:
-    branches:
-      - openwrt
-
+  pull_request:
   workflow_dispatch:
 
 jobs:
   build:
-    name: Build ipkg
+    name: Test ${{ matrix.arch }}
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: arc_archs
+            target: archs38-generic
 
-    defaults:
-      run:
-        working-directory: openwrt
+          - arch: arm_cortex-a9_vfpv3-d16
+            target: mvebu-cortexa9
+
+          - arch: mips_24kc
+            target: ath79-generic
+
+          - arch: mipsel_24kc
+            target: mt7621
+
+          - arch: powerpc_464fp
+            target: apm821xx-nand
+
+          - arch: powerpc_8540
+            target: mpc85xx-p1010
+
+          - arch: aarch64_cortex-a53
+            target: mvebu-cortexa53
+
+          - arch: arm_cortex-a15_neon-vfpv4
+            target: armvirt-32
+
+          - arch: i386_pentium-mmx
+            target: x86-geode
+
+          - arch: x86_64
+            target: x86-64
 
     steps:
-      - name: Checkout openwrt
-        uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
-          path: openwrt
-          repository: openwrt/openwrt
-
-      - name: Set openwrt ref
-        id: openwrt_ref
-        run: |
-          echo ::set-output name=REF::$(git rev-parse --short HEAD)
-          git rev-parse --short HEAD
-
-      - name: Checkout n2n
-        uses: actions/checkout@v2
-        with:
-          path: n2n
           fetch-depth: 0
 
-      - name: Fix Checkout
-        run: |
-          git fetch --force --tags
-        working-directory: n2n
-
-      - name: Set n2n ref
-        id: n2n_ref
-        run: |
-          echo ::set-output name=REF::$(./scripts/version.sh)
-          ./scripts/version.sh
-        working-directory: n2n
-
-      - name: Copy n2n package definition into openwrt
-        run: |
-          cp -r n2n/packages/openwrt openwrt/package/n2n
-        working-directory: ./
-
-      - name: Cache openwrt source downloads
-        uses: actions/cache@v2
-        with:
-          path: openwrt/dl
-          key: openwrt-dl-${{ steps.openwrt_ref.outputs.REF }}
-
-      - name: Setup openwrt config and environment
-        run: |
-          echo "CONFIG_TARGET_x86=y" >.config
-          echo "CONFIG_TARGET_x86_64=y" >>.config
-
-      - name: Add n2n package to openwrt config
-        run: |
-          echo "CONFIG_PACKAGE_n2n-edge=m" >>.config
-          echo "CONFIG_PACKAGE_n2n-supernode=m" >>.config
-
-      - name: Build a full config from our stub file
-        run: |
-          make defconfig
-
-      - name: Download openwrt sources
-        run: |
-          make download
-
-      - name: Build openwrt build environment
-        run: |
-          make -j `nproc` tools/install toolchain/install
-
-      - name: Build n2n dependancies
-        run: |
-          make -j `nproc` package/libs/libpcap/compile
-
-      - name: Build n2n openwrt packages
+      - name: Build
+        uses: openwrt/gh-action-sdk@v5
         env:
-          N2N_PKG_VERSION: ${{ steps.n2n_ref.outputs.REF }}
+          ARCH: ${{ matrix.arch }}
+          FEEDNAME: n2n_ci
+          PACKAGES: openwrt
+
+      - name: Move created packages to project dir
         run: |
-          echo "Build for $N2N_PKG_VERSION"
-          export N2N_PKG_VERSION
-          make package/n2n/clean V=s
-          make package/n2n/prepare USE_SOURCE_DIR=$GITHUB_WORKSPACE/n2n V=s
-          make package/n2n/compile V=s
+          mkdir -p ipk-${{ matrix.arch }}
+          cp bin/packages/${{ matrix.arch }}/n2n_ci/*.ipk ipk-${{ matrix.arch }} || true
 
-# FIXME: add a way to run the test suite!
-#      - name: Run embedded tests
-#        run: make test
-
-      - name: Upload built artifacts
+      - name: Store packages
         uses: actions/upload-artifact@v2
         with:
-          name: built-ipkgs
-          path: openwrt/bin/packages/*/base/*.ipk
+          name: packages
+          path: ipk-${{ matrix.arch }}/
+
+      - name: Rename logs
+        run: sudo mv -f logs logs-${{ matrix.arch }}
+
+      - name: Store logs
+        uses: actions/upload-artifact@v2
+        with:
+          name: logs
+          path: logs-${{ matrix.arch }}/
+
+      - name: Remove logs
+        run: sudo rm -rf logs-${{ matrix.arch }}/ || true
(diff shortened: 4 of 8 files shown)


hamishcoleman:
I think that there are some good things here, but I am concerned about accepting all of the change.  It would be easier if it were broken down into smaller - functionally separate - changes that we could review one at a time.

For a start, you have switched the build to use the deprecated cmake system - the Makefiles are the preferred build process, so we dont want to increase the usage of the cmake.

1715173329:
> I think that there are some good things here, but I am concerned about accepting all of the change. It would be easier if it were broken down into smaller - functionally separate - changes that we could review one at a time.

Okay I will do it today.

> For a start, you have switched the build to use the deprecated cmake system - the Makefiles  #are the preferred build process, so we dont want to increase the usage of the cmake.

Cmake is the modern build system, I have no idea why deprecated it. Cmake is simpler than legacy autotools, more human readable, and more friendly to crossing-compile. It also compiles faster.
But sure, I can switch to autotools back ;)

1715173329:
Hi @hamishcoleman, is there any way to build test tools without actually executing them? It's impossible to execute them on the host as we are doing crossing-compile.

```
make[4]: Leaving directory '/home/buildbot/immortalwrt/build_dir/target-aarch64_cortex-a53_musl/n2n-ba8855fa71ca864e180876e0a10088a73b3ce1fa/tools'
scripts/test_harness.sh tests/tests_units.list
./tools/tests-auth >tests/tests-auth.out
/lib/ld-musl-aarch64.so.1: No such file or directory
Makefile:211: recipe for target 'test.units' failed
make[3]: *** [test.units] Error 255
```

hamishcoleman:
> Hi @hamishcoleman, is there any way to build test tools without actually executing them?

The default make target will build all the binaries - including the test tools

hamishcoleman:
@1715173329 Can you point me to the documentation that you refer to for the openwrt official build process?

1715173329:
> @1715173329 Can you point me to the documentation that you refer to for the openwrt official build process?

https://github.com/openwrt/gh-action-sdk
idk if it can be called as a "documentation ", but no more details available. It's pretty simple I think.

And this is how upstream uses it:
https://github.com/openwrt/packages/tree/master/.github/workflows

hamishcoleman:
Thank you for that pointer.

How are you going with splitting this PR into smaller ones so that each functional change can be separately addressed?

1715173329:
> The default make target will build all the binaries - including the test tools

Thanks. I found them located in tools, which is different from CMake that puts binaries into root.

> How are you going with splitting this PR into smaller ones so that each functional change can be separately addressed?

All of these changes are around with OpenWrt, I have split them into different commits already.
The latter commit needs the former one to work, I have no idea how to split this PR into even "smaller" ones.

hamishcoleman:
> > The default make target will build all the binaries - including the test tools
> 
> Thanks. I found them located in tools, which is different from CMake that puts binaries into root.

Yes, the CMake files are not well maintained and have not always placed things in the right place.

> > How are you going with splitting this PR into smaller ones so that each functional change can be separately addressed?
> 
> All of these changes are around with OpenWrt, I have split them into different commits already. The latter commit needs the former one to work, I have no idea how to split this PR into even "smaller" ones.

While these are all related to OpenWRT, you are making several different changes all at once.  Some of which we would want to discuss.  If you cannot split this PR then we need to consider it all or nothing, which may not be what you were trying to achieve.  In this case, let me add some review notes.

.github/openwrt/Dockerfile
```diff
@@ -0,0 +1,6 @@
+ARG ARCH=x86-64
+FROM openwrt/rootfs:$ARCH
+
+ADD entrypoint.sh /entrypoint.sh
+
+CMD ["/entrypoint.sh"]
```
hamishcoleman, line 6: We have not previously used Docker in any of the build processes.  Can you explain exactly what this is going to be used for?  Can we avoid building a Docker image?  Is there somewhere else upstream that has this image available?

.github/openwrt/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/sh
+
+mkdir -p /var/lock/
+
+opkg update
+
+for PKG in /ci/*.ipk; do
+	tar -xzOf "$PKG" ./control.tar.gz | tar xzf - ./control
```
hamishcoleman, line 10: This looks like it is hand unpacking a package - is there not a tool that is provided to handle ipk files?

.github/openwrt/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+	# package name including variant
+	PKG_NAME=$(sed -ne 's#^Package: \(.*\)$#\1#p' ./control)
+	# package version without release
+	PKG_VERSION=$(sed -ne 's#^Version: \(.*\)-[0-9]*$#\1#p' ./control)
```
hamishcoleman, line 14: Can you explain what a "variant" and what this "release" is?

.github/openwrt/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+
+	echo "Testing package $PKG_NAME in version $PKG_VERSION from $PKG_SOURCE"
+
+	opkg install "$PKG"
+
+	export PKG_NAME PKG_VERSION
+
+	TEST_SCRIPT="/ci/.github/openwrt/test.sh"
+
+	if [ -f "$TEST_SCRIPT" ]; then
```
hamishcoleman, line 26: Since this commit has included a test.sh, it makes me think that this entrypoint.sh has been copied from a different project - if so, can you document what that project is.  Also, is it possible to share the code with the other project, instead of copying it into this one?

.github/openwrt/test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+case "$1" in
+	"n2n-edge"|"n2n-supernode")
+		eval "$1" | grep "$2"
```
hamishcoleman: This doesnt look like a very useful test.  It would probably make more sense to check the exit code of running the binary.

Note that this also relies on binary names that are non standard within this project (see more comment on this below)

.github/openwrt/test.sh
```diff
@@ -0,0 +1,12 @@
...
+		;;
+	"n2n-tests")
+		for test in $(grep -v '#' '/usr/bin/n2n-tests_units.list'); do
+			eval "n2n-$test"
+		done
+		;;
+esac
```
hamishcoleman, line 12: The functionality contained in this file looks like it might be useful outside the context of a CI build.  Have you considered putting it in the scripts/ subdirectory - or improving one of the scripts already existing for running the tests so it can be used?

.github/workflows/openwrt.yml
```diff
@@ -1,112 +1,103 @@
...
 
-      - name: Upload built artifacts
-        uses: actions/upload-artifact@v2
-        with:
-          name: built-ipkgs
-          path: openwrt/bin/packages/*/base/*.ipk
+      - name: Test via Docker container
+        if: ${{ matrix.runtime_test }}
+        run: |
+          docker run --rm -v $GITHUB_WORKSPACE:/ci test-container
```
hamishcoleman: Because this file is basically completely rewritten, it is hard to provide a valid review for it.  If the file was incrementally improved to provide the feature set that you are trying to achieve, it would be easier to review.  Failing that, it might be simpler if there were two different files - the file you are rewriting is not broken could exist side-by-side with your replacement.  This would allow a comparison in actual use to be done.

Additionally, it may well be that the existing file can still provide a good demonstration on how to do development of this code in an opwnwrt environment with quick cycle-times and access to the development codebase (which would be another reason to keep it).  I can see that this is where Docker is used, and adding that to a build often has lead to making a codebase more difficult to quickly develop, debug and fix.

CMakeLists.txt
```diff
@@ -276,7 +276,9 @@ target_link_libraries(tests-transform n2n)
 add_executable(tests-wire tools/tests-wire.c)
 target_link_libraries(tests-wire n2n)
 
-install(TARGETS n2n-benchmark RUNTIME DESTINATION bin)
+install(TARGETS n2n-benchmark n2n-keygen n2n-route n2n-portfwd
+	RUNTIME DESTINATION bin
+	)
 
 # Documentation
 if(DEFINED UNIX)
```
hamishcoleman: The cmake build system for this project is deprecated and should not be used

1715173329:
> Yes, the CMake files are not well maintained and have not always placed things in the right place.

Where CMake puts things it depends on CMakeLists, but I'm also not familiar with CMake system. These parts can be dropped.

packages/openwrt/Makefile
```diff
@@ -51,50 +30,88 @@ define Package/n2n-edge
...
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-decode $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-keygen $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-portfwd $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-route $(1)/usr/bin/
+endef
+
 $(eval $(call BuildPackage,n2n-edge))
 $(eval $(call BuildPackage,n2n-supernode))
+$(eval $(call BuildPackage,n2n-tests))
+$(eval $(call BuildPackage,n2n-utils))
```
hamishcoleman, line 120: The almost complete rewrite of this file makes it hard to review.

Additionally, it looks to me like you have copied the new file contents from another repository - a separately maintained fork of a OpenWRT build appears to have a similar file.  Can you document the upstream for this file comes from?  Can you explain any differences from the upstream?  Is there any way we can simply use the upstream file without copying it into this project?

packages/openwrt/README.md
```diff
@@ -1,62 +1,48 @@
...
-Its configuration file is `/etc/n2n/edge.conf`.
+   Copy them to the target device, and install with `opkg install`.
 
-The supernode can be started with `/etc/init.d/supernode start`.
-Its configuration file is `/etc/n2n/supernode.conf`.
+7. Configure
+
+   The edge node can be started with `/etc/init.d/edge start`. Its configuration file is `/etc/n2n/edge.conf`.
+
+   The supernode can be started with `/etc/init.d/supernode start`. Its configuration file is `/etc/n2n/supernode.conf`.
```
hamishcoleman, line 48: This is another file that appears to be basically a rewrite.  Some of the changes here do not appear to improve the documentation, they look like they have deleted some explanatory text.  Is there some reasoning behind these changes that you can expand on?

packages/openwrt/etc/init.d/edge
```diff
@@ -3,7 +3,7 @@ START=90
 STOP=10
 
 USE_PROCD=1
-PROG=/usr/bin/edge
+PROG=/usr/bin/n2n-edge
```
hamishcoleman: Changing the name of the binary breaks existing user's expectations.  It also breaks compatibility between this and the non openwrt versions

packages/openwrt/etc/init.d/edge
```diff
@@ -14,9 +14,10 @@ start_service() {
 	procd_close_instance
 }
 
-stop_service()
+reload_service()
 {
-	service_stop $PROG
+	stop
+	start
```
hamishcoleman, line 20: You appear to have removed the stop_service() completely, is that expected?

packages/openwrt/etc/init.d/supernode
```diff
@@ -3,7 +3,7 @@ START=90
 STOP=10
 
 USE_PROCD=1
-PROG=/usr/bin/supernode
+PROG=/usr/bin/n2n-supernode
```
hamishcoleman: See above, with reference to changing the binary name

packages/openwrt/etc/init.d/supernode
```diff
@@ -14,9 +14,10 @@ start_service() {
 	procd_close_instance
 }
 
-stop_service()
+reload_service()
```
hamishcoleman, line 17: See above, same question about removing stop_service

packages/openwrt/Makefile
```diff
@@ -51,50 +30,88 @@ define Package/n2n-edge
...
+	done )
+endef
+
+define Package/n2n-utils/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-benchmark $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-decode $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-keygen $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-portfwd $(1)/usr/bin/
+	$(INSTALL_BIN) $(PKG_INSTALL_DIR)/usr/bin/n2n-route $(1)/usr/bin/
```
hamishcoleman: If you are splitting this into small packages for embedded use, It seems that some of these utils can be used by the daemon in regular operations and some are only used for testing or configuration.  So, this may not be the best set to install in each package

hamishcoleman (changes requested):

hamishcoleman:
Have you noticed that there are multiple build failures in the CI?  Is there a known cause for that?

.github/workflows/openwrt.yml
```diff
@@ -1,112 +1,103 @@
...
-      - name: Setup openwrt config and environment
-        run: |
-          echo "CONFIG_TARGET_x86=y" >.config
-          echo "CONFIG_TARGET_x86_64=y" >>.config
-
-      - name: Add n2n package to openwrt config
-        run: |
-          echo "CONFIG_PACKAGE_n2n-edge=m" >>.config
-          echo "CONFIG_PACKAGE_n2n-supernode=m" >>.config
+          name: ${{ matrix.arch}}-${{ github.sha}}-packages
```
hamishcoleman: Uploading each different build to a different artifact container makes it hard to see the artifacts.  Have you considered creating one artifact for "packages-ipkg", just like we are doing with the RPM and dpkg builds?

packages/openwrt/etc/init.d/edge
1715173329: Yes, procd will automatically stop the program by sending singal `SIGTERM`.
See https://forum.openwrt.org/t/procd-shutdown-issues-questions/33759/2

1715173329:
> Have you noticed that there are multiple build failures in the CI? Is there a known cause for that?

Yeah. It's caused by a feature `AUTOREMOVE` from OpenWrt, I'm looking for a way to fix it.

packages/openwrt/etc/init.d/edge
1715173329: Okay I'll revert this change.

.github/openwrt/Dockerfile
1715173329: https://github.com/ntop/n2n/blob/06c489fd8ad42d6c025beaea0fd62d7d4d948c31/.github/workflows/openwrt.yml#L104-L106

This fixes the runtime tests. The docker image is only used for development purpose so upstream doesn't provide prebuilt one. You can see this is copied from openwrt/packages as-is.

.github/openwrt/entrypoint.sh
1715173329: Not yet.

.github/openwrt/entrypoint.sh
1715173329: Mentioned above, from github.com/openwrt/packages. `svn` can be used to download these code if you like.

.github/openwrt/entrypoint.sh
1715173329: > Can you explain what a "variant" and what this "release" is?

"variant", is just the name defined via Package/`foo`.

"release" is a part of the OpenWrt's version protocol system, see https://openwrt.org/docs/guide-developer/packages

- PKG_VERSION - `The upstream version number that we're downloading`
- PKG_RELEASE - `The version of this package Makefile`

.github/openwrt/test.sh
1715173329: It's just a rewritten of `make tests` I think. We cannot run that command in openwrt env.

.github/openwrt/test.sh
1715173329: > This doesnt look like a very useful test. It would probably make more sense to check the exit code of running the binary.

I have no idea. Do a actual runtime test can be very complex. So just do a version check here instead. If the binrary is broken it will also fail.

.github/workflows/openwrt.yml
1715173329: > Failing that, it might be simpler if there were two different files - the file you are rewriting is not broken could exist side-by-side with your replacement. This would allow a comparison in actual use to be done.

I'm not sure what you want. You may simply change `Diff view` to `Split`.

1715173329: > Additionally, it may well be that the existing file can still provide a good demonstration on how to do development of this code in an opwnwrt environment with quick cycle-times and access to the development codebase (which would be another reason to keep it).

If you want to do some developments, the proper approach is to download and use the SDK provided by upstream. In this case building toolchain from source is time-wasting IMHO.

> I can see that this is where Docker is used, and adding that to a build often has lead to making a codebase more difficult to quickly develop, debug and fix.

Mentioned above, just used for runtime tests. This is crossing-compile, we cannot do that on host.

CMakeLists.txt
1715173329: Okay, I will drop this change.

.github/workflows/openwrt.yml
1715173329: I'm not familiar with `actions/upload-artifact`. Can it be called multiple times with the same name?

packages/openwrt/Makefile
1715173329: It's from https://github.com/immortalwrt/packages/blob/master/net/n2n/Makefile, which is copied from this repo, and rewritten by me recently, then I submit these changes back.

In ImmortalWrt, we used more user-friendly uci config to generate the final configurations. Also merged `n2n-edge` and `n2n-supernode` into a single `n2n` package. Of course, one is need to download the remote source code and another is built from local source.

My $0.02 is not to rely on other projects as much as we can.

packages/openwrt/README.md
1715173329: Yeah, I just updated the build instructions and some outdated lines were removed. 
"explanatory text" is at OpenWrt docs where I pointed out.

packages/openwrt/Makefile
1715173329: I'm not familiar with them. It would be best if you can give me some advice.

hamishcoleman:
Each time you force push, it not only makes it annoying to see your change, but much more importantly, there is no comment or context for /why/ you make whatever change you made.  If you simply add new commits, each one can have a command explaining the commit - before merging, we could easily squash the extra commits that this might cause (or, perhaps better, leave them so there is a track and trail for others who are attempting to understand why the code was written in the way it is)

hamishcoleman:
I suspect that you are using some kind of automated yaml formatting tool - unless the tool can avoid making random whitespace changes between each commit, it makes it harder to see the changes that have happened.  This is especially clear with the dancing between two different types of yaml string blocks depending on the length of the string - which makes the differences touch more lines than needed for the change.

1715173329:
Uhhhh, different community has different rules. For me I just followed rules from OpenWrt (by default).

1715173329:
> I suspect that you are using some kind of automated yaml formatting tool - unless the tool can avoid making random whitespace changes between each commit, it makes it harder to see the changes that have happened. This is especially clear with the dancing between two different types of yaml string blocks depending on the length of the string - which makes the differences touch more lines than needed for the change.

Aha, are u sure? All changes were made by hand.

hamishcoleman:
> > [...] This is especially clear with the dancing between two different types of yaml string blocks depending on the length of the string - which makes the differences touch more lines than needed for the change.
> 
> Aha, are u sure? All changes were made by hand.

No, it was a guess - based on other times I have seen this.

Ideally, try to minimise the number of lines your patch touches - so if a string block is already multi-line, just leaving it
as a multi-line allows the patch to touch less lines.

(Obviously if a single-line string needs to be split for line-length reasons, this still should be done)

1715173329:
> Ideally, try to minimise the number of lines your patch touches - so if a string block is already multi-line, just leaving it
as a multi-line allows the patch to touch less lines.

My $0.02 is to keep the final file as tiny as I can.

> (Obviously if a single-line string needs to be split for line-length reasons, this still should be done)

Yeah, but this line isn't that so long and wrapping a new line doesn't really help (due to indentation.)

1715173329:
Based on the comments above, the maintainers might not use openwrt for development, so I decide to close this PR, as maintaining something unfamiliar is not a good idea.
feel free ping me if there's anyone really interested in this.

PR #1042, opened 2022-07-16
Remove the non-maintained CMake build system.

hamishcoleman:
It is difficult to maintain one build system, let alone two - especially
if one build system is known to be better maintained and the other build
system continues to be used - with bug and feature requests raised
against it.

The maintainers are aware that there are people using the CMake system,
but believe that the extra maintainability of only having one build
system, combined with the knowledge that the current CI system
demonstrates that all supported architectures are working with the
Makefiles make the use of these Makefiles a more sustainable direction.


fengdaolong:
Nowadays, more and more programs use the cmake build system, its cross-platform aspect has advantages over make, and removing cmake in n2n is not a good choice.

Logan007:
I honestly have no idea how we can maintain the CMake system and how to always keep it in sync with the automake tools at each and every change.

As far as I can see, the main concern might be with Windows platform (for which we have/offer a solution)?

llfj:
I always use CMAKE to cross compile N2N, which is much more convenient than MAKE. I do not wish to remove CMAKE.

hamishcoleman:
On Wed, Aug 10, 2022 at 02:30:22AM -0700, llfj wrote:
>    I always use CMAKE to cross compile N2N, which is much more convenient
>    than MAKE. I do not wish to remove CMAKE.

What do you need to improve the convenience of make.  You see, I always use
MAKE to cross compile N2N as it is much more convenient.

Logan007:
Also, who will maintain CMake here?

PR #1061, opened 2022-11-19
Bump config.guess

billchenchina:
config.guess now has support for:
* musl
* aarch64eb
* riscv32, riscv64
* loongarch32, loongarch64
And many other new stuff. Bump config.guess for that.


lucaderi:
Thank you

PR #1073, opened 2023-01-26
fix build issue

fnMrRice:
fix 'cannot find library -lzstd' when build as a subproject with cmake and vcpkg


hamishcoleman:
Thanks!

(Please also be aware that we consider the Cmake build in n2n as unmaintained and intend to remove it)

PR #1076, opened 2023-01-31
fix 'superndode' typo in edge and supernode help text

lemon-sh:


hamishcoleman:
Thanks!

(Though. I feel like I should create a 'ndode' service, so this can be for that.. :-S )

PR #1089, opened 2023-04-18
Fix some bugs on Windows

Legend-Master:
- Add Visual Studio files and pycache to gitignore
- Make edge prompt UAC (run as admin)
- Fix a crash on httpd since SIGPIPE doesn't exist on windows

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Install libraries needed for all options turned on
         run: |
-          sudo apt-get install -y libminiupnpc-dev libnatpmp-dev libpcap-dev
+          sudo apt-get install -y libminiupnpc-dev libnatpmp-dev libpcap-dev libcap-dev
 
       - name: Run minimal test set
         run: |


Legend-Master:
By the way, I'm confused about why python files don't have .py extension, and containing `-` in the name (can't use import)

hamishcoleman:
The python files are intended to be commandline tools, which on everything other than Windows means that they dont have an extension.

The dash in their name is something that probably should change - as you say it makes things hard for python - but was done at the request of one of the other maintainers, so we should talk to them about it.

CMakeLists.txt
```diff
@@ -211,6 +211,10 @@ endif(DEFINED WIN32)
 
 add_executable(edge src/edge.c)
 target_link_libraries(edge n2n)
+if(DEFINED WIN32)
+  # Prompt UAC
+  set_target_properties(edge PROPERTIES LINK_FLAGS "/MANIFESTUAC:\"level='requireAdministrator' uiAccess='false'\"")
+endif(DEFINED WIN32)
```
hamishcoleman: The cmake build system is deprecated and will be removed soon.
Can you make an equivalent change to the Makefiles?

scripts/n2n-httpd
```diff
@@ -456,13 +456,15 @@ def main():
...
-    signal.signal(signal.SIGPIPE, signal.SIG_DFL)
+    if hasattr(signal, 'SIGPIPE'):
+        signal.signal(signal.SIGPIPE, signal.SIG_DFL)
 
     socketserver.TCPServer.allow_reuse_address = True
     handler = functools.partial(SimpleHandler, rpc, snrpc)
 
     httpd = socketserver.TCPServer(("", args.port), handler)
     try:
+        print(f"Serving HTTP at port {args.port} (http://localhost:{args.port}/) ...")
```
hamishcoleman: Do you really think this is necessary? The existing code is following the standard unix pattern of silence-as-success.

hamishcoleman (changes requested):
Thanks!

scripts/n2n-httpd
Legend-Master: Not necessary, but I do think it's better to let the user know what happened

Feel free to delete this if you don't like it

CMakeLists.txt
Legend-Master: I see #1042, I haven't tried mingw to build, visual studio complied the cmake version automatically for me

Might try it later

Legend-Master:
> The python files are intended to be commandline tools, which on everything other than Windows means that they dont have an extension.

I would argue that It's pretty common for commandline tools to have an extension (like other .sh files), it's just easier to understand what format the file is

Legend-Master:
Have to say I hate the current work flow on Windows, the cmake version is much easier to work with

hamishcoleman:
Can you expand on what your workflow is?

(Remember that most of the people developing this software are not using visual studio)

Legend-Master:
I use vscode most of the time, but visual studio simplified a lot of the compiling and debugging configurations

I can just run debug without any configuration in visual studio before with cmake, and changing between debug and release version is just a click of button (and they're organized in their own folder)

Now I'll need to first run `hack_fakeautoconf.sh` (and the version became FIXME), and then configure vscode task and launch (and I need to define debug and optimization flags myself)

And it won't compile because of checking git tag for version if I don't use git (like directly download zip from github)

Release build from gcc is much bigger than msvc (edge.exe 338KB vs 203KB)

Legend-Master:
To be honest, it's not a big problem, but for someone who's new to the project, it's gonna make things much harder especially if they don't have much experience on GNU ecosystem like myself

hamishcoleman:
It is still good to know how others are using things.  While there is only a limited amount of resources - so we cant make everything simple - if we know the pain points, some of them might come up in passing.

Thanks for your explanation.


On 21 April 2023 1:25:39 pm GMT+01:00, Tony ***@***.***> wrote:
>To be honest, it's not a big problem, but for someone who's new to the project, it's gonna make things much harder especially if they don't have much experience on GNU ecosystem like myself
>

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Legend-Master:
I'm not sure if it's a good idea, but I think if we can have some instructions or IDE configurations to start with, it's going to make it easier to get started

Just being curious, how you guys write/build/debug it?

My vscode config, I just change the code, and press f5 to build and debug (having some problems though, vscode gdb doesn't work with ctrl-c and printing strings are messed up (none of them happens using the cmake version)):

`task.json`

```json
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "Debug build all",
            "type": "shell",
            "command": "make",
            "args": [
                "DEBUG=-g3",
            ],
            "problemMatcher": [
                "$gcc"
            ],
            "group": {
                "kind": "build",
                "isDefault": true
            },
        },
    ],
}
```

`launch.json`

```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "(gdb) Launch edge",
            "type": "cppdbg",
            "request": "launch",
            "program": "${workspaceRoot}/src/edge",
            "MIMode": "gdb",
            "cwd": "${workspaceRoot}",
            "preLaunchTask": "Debug build all"
        },
    ]
}
```

Legend-Master:
By the way, making debug build and release build in the same folder makes caching pretty messy and hard to organize, but that's something I guess can be improved in the Makefile system

hamishcoleman:
I've made some updates, you should confirm that everything still does what you expect.

hamishcoleman:
Hrm, github appears to have closed this for some reason - that was not my intent

Legend-Master:
I'm not sure what should I expect, but all my changes are gone (and the version is still FIXME)

hamishcoleman:
I figured out what I did - I pushed the wrong branch, sorry - have a look at #1103

Legend-Master:
I should have named my branch something else to be honest, was not thinking about much when I did the pull request

Just recovered from the local git repo, pushed a new branch to https://github.com/Legend-Master/n2n/tree/windows-fixes

PR #1103, opened 2023-05-08
Fix some bugs on Windows

hamishcoleman:
Replacement for #1089 after I accidentally closed it

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,8 @@ packages/etc/systemd/system/edge@.service
 packages/etc/systemd/system/supernode.service
 *dSYM*
 
+__pycache__
+
 # Binaries built to run tests
 tools/tests-auth
 tools/tests-compress
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ tools/n2n-route
 build
 .idea
 .vscode
+.vs
 packages/debian/debian/changelog
 packages/debian/debian/control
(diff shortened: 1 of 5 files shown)


Legend-Master:
I'm not sure how should I test this branch, would you mind I open another pull request from my fork's branch?

hamishcoleman:
You can checkout this branch for testing pretty easily:

```
git checkout -b hamishcoleman-Legend-Master-dev dev
git pull git@github.com:hamishcoleman/n2n.git Legend-Master-dev
```

This branch contains the changes that I am suggesting

Legend-Master:
Yes, I think it still works like before

hamishcoleman:
Great!, thanks for testing

hamishcoleman:
An unexpected side effect of the UAC prompt is that it /always/ prompts, and thus will /always/ run in a separate subshell - which makes debugging difficult and feels like a regression

Legend-Master:
> An unexpected side effect of the UAC prompt is that it /always/ prompts, and thus will /always/ run in a separate subshell - which makes debugging difficult and feels like a regression

I'm not sure how you debug it, for me, I run vscode/visual studio in admin mode even before the change to debug it, since it won't work without admin anyway

But yeah, it's kinda bad to have it flash a console window when running in a non admin terminal and quit immediately if all you want is just showing the help message

Issue #1087, opened 2023-04-17
Slow shutdown on Windows


Legend-Master:
Shutting down the edge will take a while most of the times

After doing some debugging, it seemed that `select` function is blocking us from shutting down (up to 10s when not waiting for server response)

https://github.com/ntop/n2n/blob/709590d66e4aed880b6fc2338d516d20437c17ad/src/edge_utils.c#L2876

Legend-Master:
@hamishcoleman May I ask why do we have such a long wait time that blocks exiting?

And I don't really think this is necessary, since we'll exit after this and stop reading anyway

https://github.com/ntop/n2n/blob/6ecc201017729e3065905fe7ef085d2d1f49f06c/src/edge_utils.c#L3026-L3028

Legend-Master:
> https://github.com/ntop/n2n/pull/1111#issuecomment-1601315352

@hamishcoleman I added some logging and this is what I got, note that the time waiting for tun read thread is going to be much longer if you keep edge open for a while

And it did not go through `Sleep(INFINITE)` at all

```
22/Jun/2023 10:12:03 [edge_utils.c:2597] [OK] edge <<< ================ >>> supernode
22/Jun/2023 10:12:03 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:03 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:03 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:13 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:13 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:13 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:13 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:23 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:23 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:23 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:23 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:23 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:23 [edge_utils.c:2939] start waiting for select
22/Jun/2023 10:12:33 [edge_utils.c:2941] finish waiting for select
22/Jun/2023 10:12:33 [edge_utils.c:2939] start waiting for select
[...]

PR #1111, opened 2023-05-11
Add close window support for Windows

Legend-Master:
Add cleanup support for closing the terminal/shell window on Windows

### DIFF
--- a/src/edge.c
+++ b/src/edge.c
@@ -946,6 +946,11 @@ BOOL WINAPI term_handler(DWORD sig)
 
     keep_on_running = 0;
 #ifdef WIN32
+    if (sig == CTRL_CLOSE_EVENT) {
+        // Will terminate when we return
+        // Wait for cleanup or timeout (5000ms)
+        Sleep(INFINITE);
+    }
     return(TRUE);
 #endif
 }
--- a/src/edge.c
+++ b/src/edge.c
@@ -946,10 +946,12 @@ BOOL WINAPI term_handler(DWORD sig)
 
     keep_on_running = 0;
 #ifdef WIN32
-    if (sig == CTRL_CLOSE_EVENT) {
-        // Will terminate when we return
-        // Wait for cleanup or timeout (5000ms)
-        Sleep(INFINITE);
+    switch (sig) {
+        case CTRL_CLOSE_EVENT:
+        case CTRL_LOGOFF_EVENT:
+        case CTRL_SHUTDOWN_EVENT:
+            // Will terminate us after we return, blocking it to cleanup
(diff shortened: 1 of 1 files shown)


hamishcoleman:
I've been trying to understand the logic here.

- What is the problem that is occurring for you without this patch?
- Can we avoid putting a `sleep(INFINITE)` in there? (can we not just wait for the other threads to exit?)

I'm guessing, but I think what is happening is - when the signal handler returns, windows immediately kills the process and all its threads.  So, if we signal the main loop to exit and never return from the signal handler, that gives the main loop time to do any cleanup needed.  So, what kills the sleep() thread in the signal handler?

Legend-Master:
Without it, when you close the terminal/shell window, it won't have the time to clean up which will cause authentication error if you try to reconnect

Blocking this infinitely is fine since when our main thread returns, it will shut down anyway

> I'm guessing, but I think what is happening is - when the signal handler returns, windows immediately kills the process and all its threads.

Yes, exactly

I found Go is doing this when I was searching, and there's a comment explained how it works (it also mentioned a library called libuv which is used by Node.js): https://github.com/golang/go/issues/41884#issuecomment-706695923

hamishcoleman:
Thanks for expanding on the explanation

hamishcoleman:
Is it expected that it takes 5 minutes between Ctrl-C and the process actually exiting?

Legend-Master:
> Is it expected that it takes 5 minutes between Ctrl-C and the process actually exiting?

No, I created an issue about this at #1087 2 months ago

And if you mean 5 seconds on hitting the x from the Windows console window, then that's the time when Windows decides to kill the program no matter if it's finished or not

hamishcoleman:
I wondered if it was related to the sleep(infinite).

I dont run long running tasks in console windows, so I would never close those windows from hitting the x.

These select timeouts make sense on everything other than windows.  Lets keep that other ticket open and I will see if I can look at it some time.

Legend-Master:
> I wondered if it was related to the sleep(infinite).

I don't think it runs on ctrl c

> These select timeouts make sense on everything other than windows.

I will take a look on select, if I remembered correctly the shutdown is all good when I ran the supernode on linux, but not on windows

hamishcoleman:
The select() timeout should be 10 seconds, but I am seeing anything from 3 minutes to 9 minutes!

The messages clearly show that it is going through the term_handler() codepath, but I dont have a debugger installed on my windows test VMs, so I have not traced the exact place it is hanging - so it could be anywhere.

I don't think I have seen these kind of delays on a linux systems - I would remember.

PR #1117, opened 2023-06-16
Fix purgeable is always false on supernode peer list

Legend-Master:
- And change `sn_purge` and `keep_running` to bool

Fix a regression caused by #982

Purgeable was set to 0 through `calloc` and the change from `SN_PURGEABLE = 0` to `PURGEABLE = 1` broke this

### DIFF
--- a/include/n2n.h
+++ b/include/n2n.h
@@ -50,7 +50,7 @@
 #endif
 
 
-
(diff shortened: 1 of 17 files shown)


hamishcoleman:
Thanks!

PR #1138, opened 2023-07-20
Add --persistent flag to Linux edge

sweihub:
Hi n2n

Thanks to the stable n2n-3.0, it's super stable, fast and awsome!

However, I need to make the Linux TUNTAP device persistent for routing purpose, so I added a `--persistent` flag to Linux edge. Would you review and accept this PR?

By default, or re-run without persistent flag, the TUNTAP device will be deleted when edge exits.

Thanks!

Alex Wei

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 
 # n2n
 
+NOTE: edge with `--persistent` flag will make the Linux TUNTAP device persistent, this is useful for routing.
+
 n2n is a light VPN software which makes it easy to create virtual networks bypassing intermediate firewalls.
 
 In order to start using n2n, two elements are required:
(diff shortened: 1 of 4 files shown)


hamishcoleman:
Its probably better to use the existing tools to create a persistant tun device than to patch n2n - that way you can also integrate with your system-wide network configuration and use ownership for delegation.

If you are doing this by hand:
```
ip tap add mode tap user $RUNASUSER name $IFNAME
```

sweihub:
Oh, thanks, I didn't realize that N2N can utilize the existing TAP device.
However, it's still good to have the `--persistent` flag to have a N2N persistent TAP device in one go.

Logan007:
I do no support this one either as it looks way too specific to me and can easily be solved by other tools. Also, how would this one translate into Windows version?

Issue #1139, opened 2023-07-24
Auto IP function duplicates IP address


shaxxx:
I have my supernode nice and running on Windows, with about 100 communities in community list.
Each community has subnet defined in community list, ie.
```
example1 10.0.10.0/24 #EXAMPLE 1
example2 10.0.10.0/24 #EXAMPLE 2
```
Each community has one server available 24/7, on predefined IP address set by edge client (ie. 10.0.10.1)
Other clients are dynamic, and don't have IP specified - they get it from supernode in range specified in community.list
Each edge sends it's machine name with -I option.
It's all was working perfectly, until recent problem when one of edge clients, let's call it XXX, in one specific community would get IP from supernode (ie. 10.0.10.15) but connection would not work and `ipconfig /all` would show problematic IP address marked as DUPLICATE.
Looking at output from supernode managment port, there is no clients with that IP.
If I keep disconnecting and connecting, edge client would eventually get some other IP and everything would work fine.
Currently I worked around it by randomizing information sent with -I option.

Now, this is where things get really interesting.
On this same subnet/community I have another client, lets call it YYY that usually gets IP allocated really close to the one from problematic client (IE. 10.0.10.16 or 10.0.10.17).  And I believe this is the origin of the problem.
Whenever I check managment port it shows YYY connected, but NOT with actual IP address used in community subnet.
For example, not it shows 10.0.10.17, while actual address on TAP adapter of YYY is 10.0.10.16.
I can't ping 10.0.10.17 from any edge client, but everything is fine with 10.0.10.16
Keep in mind, YYY is connected every morning, has no issues at all, can work with server on 10.0.10.1 just fine all day.
And I can see it on managment port connected all the time but with invalid IP address (usually with offset +/- of one).

Conclusion, there is list of potential problems

- auto dhcp fails to check for existing IP address properly
- auto dhcp checks for existing IP properly but existing IP list of edges is not valid for some reason
- managment is providing false data
- I'm an idiot ?

Logan007:
Hey, thank you for describing the issue you discovered. Do you only use one supernode or do you use a federated supernode setup?

We do not use real DHCP but some other simple algorithm. As you pointed to the name string, that's exactly what we hash into an IP address, and go up/downwards in case it is already taken. This does not work in a distributed way yet.

shaxxx:
No, I'm using single supernode, but with multiple communities, and like I said every community defines it's subnet in community.list and this is the first one with the problem (more than 100 communities).

[...]

PR #1142, opened 2023-08-01
Fix auto IP tracking

shaxxx:
When auto IP function assigns new IP to existing edge node, it's current IP address is not updated.
Which leads to [duplicate IP's assigned](https://github.com/ntop/n2n/issues/1139) when Pearson hash returns same value for two edge clients.


hamishcoleman:
When you have finished your testing, mark this as ready for review and we will get it merged

shaxxx:
Will do. So far it works fine, only change from previous version is there are no disconnected clients listed on the management port. 
Before, if clients did not make clean shutdown it's info would be still visible with last seen time somewhere in the past.
Now, ali visible clients last seen time is within one minute. For me this is fantastic.
Let's wait until Monday.

shaxxx:
All good, ready for merge.

Issue #1146, opened 2023-09-02
Edge not purging disconnected peers.


sduensin:
My test setup:

- Machine "winchesterhouse" running edge.
- Machine "remote2" running supernode and edge.

Starting and stopping the edge on "remote2" several times results in this from the "winchesterhouse" edge management console:

![image](https://github.com/ntop/n2n/assets/98645/db10fd98-5e13-41ee-8769-f47511fd5631)

Every time I restart edge on "remote2" I end up with another entry.

What am I doing wrong?

hamishcoleman:
The old entries should be purged.  However, I see that the oldest entry on your list is 80 seconds already - did you happen to check again after this? (I cannot remember what the defined expiry in the code is, but it should happen within 2 minutes)

sduensin:
That's just a quick/short example to show what's going on.  There's also this one:

![image](https://github.com/ntop/n2n/assets/98645/50f40a78-31e6-48a9-a597-626004cbd7df)

(And, no, I'm not an egotistical Linux God.  The machine is named "scott-linux-godot".)

hamishcoleman:
Yeah, that's definitely unexpected!

From your original screenshot, it looks like you are running v3.1.1, but that version number is a little unexpected as the dev branch has moved on from that point.  Can you confirm what version you are running and how you compiled it?

Can you share the config you are using for both the edge and supernode?

With all that in hand, we might be able to reproduce the issue

sduensin:
I'm trying to do some experiments now to see exactly what I've compiled and where it went sideways.  I *think* I compiled the head of 3.1.1 from a week ago, but dumb me didn't keep the git metadata around so I don't know which commit it was.  I'm going to re-pull and try a few different versions to see what is really going on.

The only thing different that I did that probably isn't recommended is my build only includes pthreads - no other external libraries.

hamishcoleman:
Strange, simply compiling a git checkout of head should have resulted in more details showing in the version number shown (eg "3.1.1-235-gcd7b3b")

pthreads is not "not recommended" - just less of the autobuilds use it.  I also dont believe adding threads should touch the code that does the entry purging either..

sduensin:
I just did this:
```
mkdir n2nwork
cd n2nwork
git clone https://github.com/ntop/n2n.git
cd n2n
./autogen.sh
./configure --enable-pthread
make
```

The resulting binaries are still behaving as they were.  Neither of these edge instances are currently running:

[...]

PR #1147, opened 2023-09-06
Purgable fixes

hamishcoleman:
Hopefully addressing #1146


JimLee1996:
Hope this fix merged ASAP, as the reconnection stategy may fail due to uncorrect count of peers.
https://github.com/ntop/n2n/blob/dev/src/edge_utils.c#L1622

sduensin:
I've been using this since it was posted.  It takes a bit longer than I'd like to purge a disconnected peer, but it works!

PR #1174, opened 2024-04-30
Add SRV record support for win32

nifanfa:
just for hidding the port.
like i can connect to n2n server by using this
```
edge -c cf -l _n2n_udp.example.com --no-port-forwarding -r -D -E
```
instead of
```
edge -c cf -l example.com:7777 --no-port-forwarding -r -D -E
```
but i don't know how to use a cross-platfrom way to get SRV records. so only win32 is implemented.

Please sign (check) the below before submitting the Pull Request:

- [x] I have signed the ntop Contributor License Agreement at https://github.com/ntop/legal/blob/main/individual-contributor-licence-agreement.md
- [x] I have updated the documentation (in doc/) to reflect the changes made (if applicable)

Link to the related [issue](https://github.com/ntop/n2n/issues):

Describe changes:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,6 +185,7 @@ src/edge: $(N2N_LIB)
 src/supernode: $(N2N_LIB)
 
 ifneq (,$(findstring mingw,$(CONFIG_HOST_OS)))
+LDLIBS+=-ldnsapi
 N2N_OBJS+=src/win32/edge_utils_win32.o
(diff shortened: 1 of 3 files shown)


hamishcoleman:
The project you have raised a PR against is no longer very active.  If you would like to raise a feature request ticket in the recently released fork [n3n](https://github.com/n42n/n3n/),  we could discuss things further as part of that project.

PR #1230, opened 2025-09-08
Fixing missing dependencies in Makefile

Meiye-lj:
This PR fixes an issue in the Makefile. Specifically, previously, any modifications of files like include/uthash.h would not trigger a rebuild of src/wire.o. The PR fixes this by including them as additional dependencies.


lucaderi:
Thank you

PR #1239, opened 2026-03-03
Harden CI workflows and update few Security Findings

mchsk:
## Summary

Security hardening for GitHub Actions workflows and a Windows API target update. Addresses supply chain, injection, least-privilege, and API surface concerns.

### Supply chain: Pin actions to SHA hashes
All third-party GitHub Actions are now pinned to immutable commit SHAs instead of mutable version tags. This prevents a compromised upstream action repository from injecting malicious code via tag reassignment.

| Action | Before | After |
|--------|--------|-------|
| `actions/checkout` | `@v4` | `@34e114876b...` |
| `actions/upload-artifact` | `@v4` | `@ea165f8d65...` |
| `actions/download-artifact` | `@v4` | `@d3f86a106a...` |
| `codecov/codecov-action` | `@v4` | `@b9fd7d16f6...` |
| `softprops/action-gh-release` | `@v2` | `@a06a81a03e...` |

### Injection prevention: Replace expression interpolation in shell
All `${{ matrix.* }}` expressions in `run:` shell blocks are replaced with `env:` variable references (`$MATRIX_OS`, `$MATRIX_ARCH`). This prevents potential script injection if matrix values were ever sourced from external input (PR titles, labels, workflow_dispatch inputs).

### Least privilege: Explicit permissions
- Added top-level `permissions: contents: read`
- Scoped `permissions: contents: write` only to the `upload_release` job that needs it

### Codecov token
Added explicit `token: ${{ secrets.CODECOV_TOKEN }}` to all `codecov-action` invocations. Tokenless uploads for public repos have been a [historical attack vector](https://about.codecov.io/security-update/).

### Shell safety
- Quoted `$GITHUB_REF` in the release job to prevent word splitting
- Added comments documenting the rationale for `|| true` and `continue-on-error: true` in Windows coverage steps

### Windows API target
Updated `_WIN32_WINNT` from `0x501` (Windows XP SP1) to `0x0601` (Windows 7) in `win32/n2n_win32.h`. Windows XP has been EOL since 2014; targeting it restricts access to modern security APIs and compiler mitigations.

## Files changed (3)
- `.github/workflows/tests.yml`
- `.github/workflows/cmake-linux.yml`
- `win32/n2n_win32.h`

## Test plan
- All CI jobs should pass (the workflow changes are backwards-compatible)
- Note: `CODECOV_TOKEN` secret must be configured in the repository settings for codecov uploads to succeed with token auth. Without it, codecov steps will fail but won't block the pipeline.

### DIFF
--- a/.github/workflows/cmake-linux.yml
+++ b/.github/workflows/cmake-linux.yml
@@ -4,6 +4,9 @@ name: CMake
 # yamllint disable-line rule:truthy
 on: [push]
 
+permissions:
+  contents: read
+
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
   BUILD_TYPE: Release
@@ -24,7 +27,7 @@ jobs:
           - os: macos-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
 
       - name: Create Build Environment
         # Some projects don't allow in-source building, so create a separate
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,13 +6,16 @@ on:
   push:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   smoketest:
     name: Smoke test
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
       - name: Run minimal test set
         run: |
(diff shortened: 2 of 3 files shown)


mchsk:
@lucaderi Hi Luca, wanna merge this, or help somehow to create `3.1` version perhaps? Since this is coming back to `3.0`

mchsk:
I would suggest to say, that 3.1 would be mainly about support for MacOS - using the native `utun` interface, and under the hood there obviously is all the updating to the latest, greatest, etc. I think I covered all of it. @lucaderi

lucaderi:
Do we really need to pin versions?

